PR #24, opened 2024-11-06
feat(bulk-import): upgrading to mui5

its-mitesh-kumar:
Resolves : https://issues.redhat.com/browse/RHIDP-4308


its-mitesh-kumar:
cc :  @debsmita1

04kash:
@its-mitesh-kumar To fix CI issues, you'd have to run `yarn dedupe` from the bulk-import workspace

workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositories.tsx
```diff
@@ -34,23 +36,11 @@ import { PreviewFileSidebar } from '../PreviewFile/PreviewFileSidebar';
...
-    flexDirection: 'row',
-    justifyContent: 'left',
-    alignItems: 'center',
-    paddingTop: '24px',
-    paddingBottom: '24px',
-    paddingLeft: '16px',
-    backgroundColor: theme.palette.background.paper,
-    borderBottomStyle: 'groove',
-    border: theme.palette.divider,
-  },
```
debsmita1, line 53: Can you comment this for now instead of removing. We would need this once the `ServiceNow` approval tool is incorporated in the plugin

debsmita1:
Tested the changes locally, and the changes look fine
@its-mitesh-kumar Please fix the Resize observer issue with the side panel

its-mitesh-kumar:
@divyanshiGupta 
Resize observer issue @debsmita1  has mentioned a warning that will be shown only in dev mode . 
Screenshot 2024-11-13 at 3 41 39 PM
Cc : @christoph-jerolimov

divyanshiGupta:
If we aren't using ResizeObserver in the code we can maybe ignore this issue for this PR as it will not appear in prod but it will be still better to fix it in the dev mode in this PR if possible and if not then we can merge this PR and raise another for the dev mode once we have a solution.

its-mitesh-kumar:
@divyanshiGupta ! With the below commit resize issue has gone .
https://github.com/redhat-developer/rhdh-plugins/pull/24/commits/69036a1f8d9dc64df2f3bf9cadf315f2cf839290

workspaces/bulk-import/packages/app/src/components/search/SearchPage.tsx
```diff
@@ -42,13 +42,10 @@ import { makeStyles } from '@mui/styles';
...
-    padding: theme.spacing(1, 0),
-  },
-  filters: {
-    padding: theme.spacing(2),
-    marginTop: theme.spacing(2),
-  },
+  // filters: {
+  //   padding: theme.spacing(2),
+  //   marginTop: theme.spacing(2),
+  // },
```
divyanshiGupta, line 48: Remove or uncomment

workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositories.tsx
divyanshiGupta: @its-mitesh-kumar please add comment describing why this code is commented out.

divyanshiGupta (changes requested):

christoph-jerolimov:
I still have the resize issue; it depends on how your OS renders the scrollbar.

The issue is logged here in MUI: https://github.com/mui/material-ui/issues/43718 (there are also multiple duplicates)

For me, this issue appears only when multiple TextFields with the `multiline` prop (an auto-resizing text field) are used at the same time. But that's exactly what we want here.

I think we can go ahead for now.

divyanshiGupta (approved):
/lgtm

christoph-jerolimov:
I noticed one difference, can we check if we can fix this?

The background color in dark mode changed:

Before:

before

With this PR:

after

its-mitesh-kumar:
@christoph-jerolimov ! For me its looking like below . 
Screenshot 2024-11-14 at 4 11 51 PM

christoph-jerolimov:
Don't understand how this is possible, but we can start merging and then we will take a look again.

christoph-jerolimov (approved):

PR #55, opened 2024-11-19
chore: regenerate knip reports

04kash:

- Regenerate knip reports for all workspaces
- Make knip a devDependency instead of a dependency

Fixes: https://issues.redhat.com/browse/RHIDP-4959


workspaces/dynamic-home-page/.yarn/install-state.gz
christoph-jerolimov, line 1: Can you remove this file  from your PR, its a left over after I renamed the `dynamic-home-page` workspace to `homepage`.

workspaces/bulk-import/packages/app/knip-report.md
```diff
@@ -0,0 +1,20 @@
+# Knip report
+
+## Unused dependencies (5)
+
+| Name                                                | Location     | Severity |
+| :-------------------------------------------------- | :----------- | :------- |
+| @red-hat-developer-hub/backstage-plugin-bulk-import | package.json | error    |
```
christoph-jerolimov, line 7: Do you understand this false-positive?

The import is used in `workspaces/bulk-import/packages/app/src/App.tsx`

workspaces/bulk-import/packages/backend/knip-report.md
```diff
@@ -0,0 +1,25 @@
+# Knip report
+
+## Unused dependencies (11)
+
+| Name                                                        | Location     | Severity |
+| :---------------------------------------------------------- | :----------- | :------- |
+| @red-hat-developer-hub/backstage-plugin-bulk-import-backend | package.json | error    |
```
christoph-jerolimov, line 7: Same here, it's used... Hmm :thinking:

christoph-jerolimov (changes requested):
Hi, please remove at least the yarn blob (you can remove th dynamic-home-page workspace from your machine).

Also I see some false positives for "unused dependencies". I don't know how useful this is when the report is incorrect.

We can merge anyway but I think we should also check that we check that the file is accurate  (generate and ensure there is no diff)  on each PR.

workspaces/dynamic-home-page/.yarn/install-state.gz
04kash: Totally missed that, thanks @christoph-jerolimov !

christoph-jerolimov (approved):
Feel free to merge it when you're fine with it.

With two things we should follow-up:

1. Check/remove the false positives
2. Add this as a PR check (the idea is similar to the api-report.md, it makes changes like unused dependencies easier to notice)

04kash:
> Add this as a PR check (the idea is similar to the api-report.md, it makes changes like unused dependencies easier to notice)
I was actually thinking of setting up a workflow to run this script as a cron job on a regular basis: https://github.com/redhat-developer/rhdh-plugins/blob/main/scripts/ci/generate-knip-report.js

But your suggestion to include it in the PR checks also sounds good

04kash:
Merging this as is, will later open a follow-up PR to remove false positives.

PR #73, opened 2024-11-25
fix(orchestrator): execute should allow no inputs

batzionb:
The bug is that execute fails also if field inputData is passed but is empty.
This causes workflows without required inputs to fail.
Changing inputData field to be optional

This resolves the bug in a cleaner way than https://github.com/janus-idp/backstage-plugins/pull/2550

Resolves [FLPATH-1880](https://issues.redhat.com/browse/FLPATH-1880)


workspaces/orchestrator/plugins/orchestrator-backend/src/service/api/v2.ts
```diff
@@ -163,12 +163,11 @@ export class V2 {
     workflowId: string,
     businessKey: string | undefined,
   ): Promise<ExecuteWorkflowResponseDTO> {
-    if (Object.keys(executeWorkflowRequestDTO?.inputData).length === 0) {
+    if (!executeWorkflowRequestDTO?.inputData) {
```
gciavarrini, line 166: inputData is a required field in the spec. You need to fix the spec too

batzionb: if it's required why are we checking this?

batzionb: Revisisted it
I think the inputData should be optional, so I changed the spec for it to be optional
Actually useful, a workflow without parameters can be run without need to pass a body:
curl -X POST http://localhost:7007/api/orchestrator/v2/workflows/hello_world/execute

workspaces/orchestrator/app-config.yaml
```diff
@@ -35,6 +35,9 @@ backend:
     connection: ':memory:'
   cache:
     store: memory
+  # useful for debugging APIs via curl in dev env without requiring a auth:
```
mareklibra, line 38: This can be achieved by https://backstage.io/docs/auth/service-to-service-auth#static-tokens
Very easy to set up and use.

mareklibra (approved):

PR #84, opened 2024-11-26
fix(orchestrator): fix fetchWorkflowOverviewBySource to fetch less instances

ydayagi:
FLPATH-1893
https://issues.redhat.com/browse/FLPATH-1893

PR #94, opened 2024-11-28
fix(orchestrator): remove average duration from workflow table and overview page

batzionb:
Due to performance limitations, backend API cannot return the average duration.
This is a follow up on https://github.com/redhat-developer/rhdh-plugins/pull/84/files


batzionb:
@mareklibra PTAL

batzionb:
@mareklibra PTAL

ydayagi (approved):

PR #102, opened 2024-12-02
fix(orchestrator): improve alignment of key-value pairs on the Review step

mareklibra:
Fixes: FLPATH-1315


batzionb:
I see this problem exists upstream as well
Maybe worth creating a PR there as well

image

batzionb (approved):

PR #122, opened 2024-12-04
fix(theme): fixed that all disabled elements used a disable background color instead of just disabled tabs

christoph-jerolimov:

TODO

### DIFF
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -436,7 +436,7 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         root: {
           textTransform: 'none',
           minWidth: 'initial !important',
-          '&.Mui-disabled': {
+          '&:disabled': {
             backgroundColor: general.disabledBackground,
           },
         },
--- a/workspaces/theme/.changeset/chilled-countries-divide.md
+++ b/workspaces/theme/.changeset/chilled-countries-divide.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+fixed that all disabled elements used a disable background color instead of just disabled tabs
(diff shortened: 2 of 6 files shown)


workspaces/theme/plugins/theme/src/utils/createComponents.ts
```diff
@@ -436,7 +436,7 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         root: {
           textTransform: 'none',
           minWidth: 'initial !important',
-          '&.Mui-disabled': {
+          '&:disabled': {
             backgroundColor: general.disabledBackground,
```
ciiay, line 440:
```suggestion
            backgroundColor: 'unset',
```
⬆️ fixes the following issue with tab button background color being gray.
image

ciiay (changes requested):
Disabled buttons are fixed, just left a comment about the tab button being gray issue.

workspaces/theme/plugins/theme/src/utils/createComponents.ts
christoph-jerolimov: Then I can just remove this disabled code. I thought setting a gray background to the tab was exactly what we want to make it more clear that the tab is disabled.

With just the different text Color is really hard to see if a tab is disabled. I will check with @ShiranHi and the PF 5/6 documentation.

christoph-jerolimov: @ciiay yeah this "idea" comes from PF. See: https://v5-archive.patternfly.org/components/tabs

image

ciiay (approved):
/lgtm

Thank you 🤝

workspaces/theme/plugins/theme/src/utils/createComponents.ts
ShiranHi: Right, let's follow [PF guidelines](https://www.patternfly.org/components/tabs/design-guidelines#disabled-tabs). Do they use the same gray color for both light and dark mode?

christoph-jerolimov: @ShiranHi, I like that you asking the uncomfortable, on-point questions :rofl: 

Here we go: #129

PR #128, opened 2024-12-05
chore(orchestrator): add export dynamic scripts and fix backstage version

batzionb:
- added export-dynamic scsripts
- fixed backstage version in backstage.json that didn't match used packages


workspaces/orchestrator/plugins/orchestrator/package.json
```diff
@@ -84,6 +85,7 @@
     "@backstage/dev-utils": "1.1.2",
     "@backstage/test-utils": "1.7.0",
     "@backstage/types": "1.1.1",
+    "@janus-idp/cli": "1.17.0",
```
gciavarrini, line 88: is it necessary to add also this?
```json
"files": [
    "dist",
    "dist-scalprum" 
]
```

batzionb: seems like it's already there?
https://github.com/redhat-developer/rhdh-plugins/pull/128/files/6a794197ff98dc9c28953e15c5841f2dfe0eb351#diff-80998956ee31ced75eaa0eaf1d1b60a7a7a84fc2a6176b4b55a8a2fb3471469eL37

gciavarrini: oh ok! it's above! thanks for checking

gciavarrini (approved):
/lgtm

PR #130, opened 2024-12-05
chore(orchestrator): fix release action

batzionb:
fix failing release action - https://github.com/redhat-developer/rhdh-plugins/actions/runs/12120055908/job/33787933673

failed since changeset in https://github.com/redhat-developer/rhdh-plugins/pull/95 included a private package @red-hat-developer-hub/backstage-plugin-orchestrator-swf-editor-envelope 
since this PR also included fixing the build, we can make the package public


gciavarrini (approved):
/lgtm

PR #160, opened 2024-12-11
fix(orchestrator): fix scalprum config

batzionb:
fix scalprum name following move to new repository, to match name in https://github.com/redhat-developer/rhdh-plugins/blob/fixscalprum/workspaces/orchestrator/plugins/orchestrator/app-config.yaml#L3


gciavarrini (approved):
/lgtm

mareklibra (approved):

PR #166, opened 2024-12-12
Orchestrator: Workflow runs

LiorSoffer:
resolve [FLPATH-1042](https://issues.redhat.com/browse/FLPATH-1042)
image

https://github.com/user-attachments/assets/9e722f85-03bf-40e3-9bae-87311813d769


workspaces/orchestrator/plugins/orchestrator/src/routes.ts
```diff
@@ -15,30 +15,40 @@
...
-  parent: orchestratorRootRouteRef,
-  path: '/workflows/:format/:workflowId',
+  id: 'orchestrator', // display WorkflowsTabContent
 });
 
 export const workflowInstancesRouteRef = createSubRouteRef({
   id: 'orchestrator/instances',
   parent: orchestratorRootRouteRef,
-  path: '/instances',
+  path: '/instances', // display WorkflowRunsTabContent
```
mareklibra, line 26: I got the motivation behind these comments but they can be confusing if not properly maintained in the long term.

I don't think we need to give this hint at all. If you think otherwise, I would suggest to mention the `router.ts` as the reference of mapping. But I think it's obvious in this context.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContentFiltered.tsx
```diff
@@ -0,0 +1,165 @@
...
+  const { loading, error, value } = usePolling(fetchInstances);
+
+  const columns = React.useMemo(
+    (): TableColumn<WorkflowRunDetail>[] => [
+      {
+        title: 'ID',
+        field: 'id',
+        render: data => (
+          <Link to={workflowInstanceLink({ instanceId: data.id })}>
+            {ellipsis(data.id)}
```
mareklibra, line 90: The UUID length is constant, we do not need to shorten it in the majority of cases.
For very narrow displays, we can play with the CSS `overflow`.

---
Recently we are wasting the space:

Screenshot From 2024-12-13 09-06-35

mareklibra (changes requested):
Works nice, just a few comments. Thank you :-)

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContentFiltered.tsx
batzionb: @mareklibra 
the instance ids are rather long and there's no need to see the whole thing
This is also unrelated to this code change
image

mareklibra: > the instance ids are rather long and there's no need to see the whole thing

In that case we do not need to show it at all.

LiorSoffer:
Hey, thank you for the detailed notes ,working on fixing them all :)

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContentFiltered.tsx
batzionb: what should we show instead? What should the user do if he wants to search for a specific instance?

batzionb: This is out of the scope of this PR, needs to be discussed with UX

workspaces/orchestrator/plugins/orchestrator/src/hooks/usePolling.ts
```diff
@@ -74,7 +74,7 @@ const usePolling = <T>(
 
   return {
     value: data,
-    error,
```
batzionb, line 77: why did you make this change?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContentFiltered.tsx
mareklibra: I have noticed this issue since the code is being touched here. But agree, we can move it to another issue.

IMO, we should show the user either correct and full info or nothing. What is a fragment of the UUID for?

If the user is not interested in the UUID (which is the case IMO), we don't need to show it. And link the instance detail through the Workflow name or Started column. The UUID can be shown on the details page.

I think we can move the Started column to the first position and make a link out of it.

I will open a Jira issue.

batzionb:
@LiorSoffer 
please add screenshots to PR description

batzionb:
@LiorSoffer 
The second line in the tabs toolbar is shorter than the first?

ShiranHi:
@LiorSoffer could you please provide a screen recording of the flow? A screenshot isn’t very helpful in this case. Thanks :)

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowDefinitionViewerPage/WorkflowDefinitionViewerPage.tsx
```diff
@@ -14,113 +14,50 @@
...
-import {
-  executeWorkflowRouteRef,
-  workflowDefinitionsRouteRef,
-} from '../../routes';
-import { BaseOrchestratorPage } from '../BaseOrchestratorPage';
+import { workflowRouteRef } from '../../routes';
 import { EditorViewKind, WorkflowEditor } from '../WorkflowEditor';
 import WorkflowDefinitionDetailsCard from './WorkflowDefinitionDetailsCard';
 
-export const WorkflowDefinitionViewerPage = () => {
```
batzionb, line 45: I suggest to rename the component and file to WorkflowDetailsTabContent, since Page indicates that it's a whole page

LiorSoffer: should I change the name of the matching file as well?
what about the card name?
image

batzionb: yes, also the the matching file
The card name is fine

ShiranHi:
@LiorSoffer looks good. We need to remove the Category column, and increase the ID column width

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -131,9 +131,12 @@ export const WorkflowInstancePage = ({
     return res.data;
   }, [instanceId, orchestratorApi, queryInstanceId]);
 
-  const { loading, error, value, restart } = usePolling<
-    AssessedProcessInstanceDTO | undefined
-  >(
+  const {
+    loading,
+    error: error,
```
batzionb, line 136:
```suggestion
error,
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -68,10 +74,12 @@ export const WorkflowRunsTabContent = () => {
...
       instances.data.items?.map(mapProcessInstanceToDetails) || [];
-    return clonedData;
-  }, [orchestratorApi]);
+    return workflowId
+      ? clonedData.filter(item => item.workflowId === workflowId)
+      : clonedData;
+  }, [orchestratorApi, workflowId]);
 
-  const { loading, error, value } = usePolling(fetchInstances);
+  const { loading: loading, error: error, value } = usePolling(fetchInstances);
```
batzionb, line 82:
```suggestion
  const { loading, error, value } = usePolling(fetchInstances);
```

workspaces/orchestrator/plugins/orchestrator/src/hooks/usePolling.ts
```diff
@@ -74,7 +74,7 @@ const usePolling = <T>(
...
+    error: error,
```
batzionb, line 77:
```suggestion
    error,
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowPageTabContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+                color="primary"
+                onClick={handleExecute}
+                disabled={!canRun}
+              >
+                Run
+              </Button>
+            </Tooltip>
+          )}
+        </Grid>
+      </Grid>
```
batzionb, line 76:
```suggestion
      </Grid>
      <Grid item>
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowPageTabContent.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {children}
```
batzionb, line 77:
```suggestion
         {children}
      </Grid>
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowPage.tsx
```diff
@@ -0,0 +1,91 @@
...
+            canRun={canRun}
+          >
+            <WorkflowDetailsTabContent
+              loading={loading}
+              workflowOverviewDTO={workflowOverviewDTO}
+            />
+          </WorkflowPageTabContent>
+        </TabbedLayout.Route>
+        <TabbedLayout.Route
+          path={workflowRunsRouteRef.path.split('/').pop() || 'runs'}
```
batzionb, line 76: hmmm...
maybe better put 'runs' as a constant in routes.ts
and use it both here and in definition of workflowRunsRouteRef
so you won't need this string processing and the undefined handling

batzionb:
@LiorSoffer Well done!
I added some comments, and as @ShiranHi mentioned the category column needs to be removed when in workflow page
And since there's plenty of room, ID doesn't need to be minimized

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { workflowRouteRef, workflowRunsRouteRef } from '../routes';
+import { BaseOrchestratorPage } from './BaseOrchestratorPage';
+import { WorkflowDefinitionViewerPage } from './WorkflowDefinitionViewerPage';
+import { WorkflowRunsTabContentFiltered } from './WorkflowRunsTabContentFiltered';
+
+export const WorkflowPage = () => {
+  const { workflowId } = useRouteRefParams(workflowRouteRef);
+  return (
+    <>
+      <BaseOrchestratorPage title={workflowId} noPadding>
```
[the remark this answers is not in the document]
batzionb: was fixed to show correct header

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -57,6 +61,8 @@ const makeSelectItemsFromProcessInstanceValues = () =>
   );
 
 export const WorkflowRunsTabContent = () => {
+  const { workflowId } = useRouteRefParams(workflowRouteRef);
+  const statuses = makeSelectItemsFromProcessInstanceValues();
```
batzionb, line 65: with this change, the statuses are recalculated for every render...
what @mareklibra meant was to move this to be a constant outside the component, right under the function definition of makeSelectItemsFromProcessInstanceValues

LiorSoffer:
This is how the ID looks now @ShiranHi 
image
image

ShiranHi:
@LiorSoffer that's great! One small note, can we bring closer the last two columns to the Status column? It will be nice to keep the same padding between all of the columns

batzionb:
> @LiorSoffer that's great! One small note, can we bring closer the last two columns to the Status column? It will be nice to keep the same padding between all of the columns

We can make status, duration and started take less space, but we can't garantuee the same padding

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowsTabContent.tsx
```diff
@@ -40,9 +40,11 @@ export const WorkflowsTabContent = () => {
     return overviewsResp.data.overviews;
   }, [orchestratorApi]);
 
-  const { loading, error, value } = usePolling<
-    WorkflowOverviewDTO[] | undefined
-  >(fetchWorkflowOverviews);
+  const {
+    loading,
+    error: error,
```
batzionb, line 45:
```suggestion
     error,
```

batzionb:
@mareklibra 
Can you please take one more look to see your change requests are done?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
```diff
@@ -0,0 +1,63 @@
...
+  loading: boolean;
+  workflowOverviewDTO: any;
+}
+
+export const WorkflowDetailsTabContent = ({
+  loading,
+  workflowOverviewDTO,
+}: Props) => {
+  const { workflowId, format } = useRouteRefParams(workflowRouteRef);
+  const workflowFormat = useMemo(
```
mareklibra, line 37: nitpick (no need to fix): since both the `json` and `yaml` are string constants, avoiding the useMemo() will perform faster, it will no effect on re-render at the line 56.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <>
+      <Grid item>
+        <WorkflowDefinitionDetailsCard
+          workflowOverview={workflowOverviewDTO}
+          loading={loading}
+        />
+      </Grid>
+      <Grid item>
+        <InfoCard title="Workflow definition">
+          <div style={{ height: '600px' }}>
```
mareklibra, line 52: we should use `useStyle()` instead

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowPage.tsx
```diff
@@ -0,0 +1,88 @@
...
+      orchestratorWorkflowUsePermission,
+      orchestratorWorkflowUseSpecificPermission(workflowId),
+    ]);
+
+  const {
+    value: workflowOverviewDTO,
+    loading,
+    error,
+  } = useAsync(() => {
+    return orchestratorApi.getWorkflowOverview(workflowId);
```
mareklibra, line 51: nitpick (no need to fix):
```suggestion
    useAsync(() => orchestratorApi.getWorkflowOverview(workflowId))
```

mareklibra (approved):
Nice.
I left just nitpicks, it's up to your decision whether fixing them.

PR #178, opened 2024-12-13
feat(bulk-import): adding sorting based on column in /imports

its-mitesh-kumar:
Resolves :  https://issues.redhat.com/browse/RHIDP-5295


workspaces/bulk-import/plugins/bulk-import/src/types/types.ts
```diff
@@ -97,7 +97,17 @@ export enum RepositoryStatus {
...
+  repoName = 'repository.name',
+  organizationUrl = 'repository.organization',
+  repoUrl = 'repository.url',
+  catalogInfoYamlLastUpdated = 'lastUpdate',
+  catalogInfoYamlStatus = 'status',
+}
+export enum SortingOrderEnum {
+  ASC = 'asc',
+  DESC = 'desc',
+}
```
christoph-jerolimov, line 110: Please respect the convention in this file

* and add empty lines between each type definition
* and it looks like enum keys here are all UPPER_CASE

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
```diff
@@ -156,8 +165,9 @@ export async function findAllImports(
     });
 
   // sorting the output to make it deterministic and easy to navigate in the UI
-  sortImports(imports);
-
+  if (sortColumn) {
+    sortImports(imports, sortColumn, sortOrder);
+  }
```
christoph-jerolimov, line 170: Are we always fetching ALL elements in the backend?

And guess that we are and should not do that. So this sort function will only sort the current window (fetched page).

Instead of that we should pass the sorting to the backend calls like GitHub and the catalog API and remove the sorting from our backend.

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
```diff
@@ -45,18 +50,22 @@ type FindAllImportsResponse =
...
+    // Handle cases where values are undefined
+    if (value1 === undefined && value2 === undefined) return 0;
+    if (value1 === undefined) return SortingOrderEnum.ASC ? -1 : 1;
+    if (value2 === undefined) return SortingOrderEnum.ASC ? 1 : -1;
+    // Compare values based on sort order
+    return SortingOrderEnum.ASC === sortOrder
+      ? value1.localeCompare(value2)
+      : value2.localeCompare(value1);
   });
 }
```
christoph-jerolimov, line 70: We should eliminate this and add the sorting parameter to the GitHub / Catalog API calls.

christoph-jerolimov (changes requested):
Hi @its-mitesh-kumar, thanks for picking up also the backend work.

I added some comments, that IMHO we should remove the sorting in our backend, and add sorting params to the request calls for GitHub / Catalog API calls.

Let me know if I miss something and its not possible for some reason. And let the backend team and me know if you want continue with this.

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
dzemanov: Yes, it would be the best if we could sort and paginate on the Github side.
I believe we are at the backend unfortunately fetching for `imports` all Catalog locations and then filtering those accessible for user via integrations. @rm3l would be the best to answer if this could be adjusted to be more efficient. Maybe paginating and sorting can be added to [filterLocationsAccessibleFromIntegrations](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/github/githubApiService.ts#L372-#L417).

dzemanov: Yes, it would be the best if we could sort and paginate at the Github side. Unfortunately, for `/imports` endpoint, I believe we are fetching all Catalog locations and then filtering those reachable with GH integrations. @rm3l might have valuable insights into the reasoning behind this solution and whether we can approach this more effectively. Maybe we can adjust [filterLocationsAccessibleFromIntegrations](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts#L116-#L120) [here](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/github/githubApiService.ts#L372-#L430).

When [fetching Catalog locations](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts#L90-#L96), `pageNumber` and `pageSize` are passed as params but not really used there, since we don't know which ones are actually reachable GH integrations. Maybe the best approach would be to switch stuff around and first fetch GH integrations, maybe it would help performance.

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
its-mitesh-kumar: Github APIs not supporting sorting on these params ,

workspaces/bulk-import/plugins/bulk-import-backend/src/helpers/utils.ts
```diff
@@ -0,0 +1,30 @@
...
+        acc && (acc as Record<string, any>)[key]
+          ? (acc as Record<string, any>)[key]
+          : undefined,
+      obj,
+    );
+}
+export enum SortingOrderEnum {
+  ASC = 'asc',
+  DESC = 'desc',
+}
```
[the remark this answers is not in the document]
its-mitesh-kumar: Github APIs not supporting sorting on these params .

its-mitesh-kumar:
@christoph-jerolimov @dzemanov I have resolved the comments , Please re-review it

workspaces/bulk-import/plugins/bulk-import/src/types/types.ts
```diff
@@ -98,6 +98,19 @@ export enum RepositoryStatus {
...
 }
 
+export enum AddedRepositoryColumnNameEnum {
+  repoName = 'repository.name', // because of property name
+  repositoryName = 'repository.name',
+  organizationUrl = 'repository.organization',
+  repoUrl = 'repository.url',
+  catalogInfoYamlLastUpdated = 'lastUpdate',
+  catalogInfoYamlStatus = 'status',
+}
```
dzemanov, line 108: Maybe it would be good to unify this and not repeat values. If `repoName` is needed, you can remove `repositoryName` and instead always use `repoName`.

```suggestion
export enum AddedRepositoryColumnNameEnum {
  repoName = 'repository.name',
  organizationUrl = 'repository.organization',
  repoUrl = 'repository.url',
  catalogInfoYamlLastUpdated = 'lastUpdate',
  catalogInfoYamlStatus = 'status',
}
```

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
```diff
@@ -74,16 +91,16 @@ export async function findAllImports(
...
   const apiVersion = requestHeaders?.apiVersion ?? 'v1';
   const search = queryParams?.search;
   const pageNumber = queryParams?.pageNumber ?? DefaultPageNumber;
   const pageSize = queryParams?.pageSize ?? DefaultPageSize;
-
-  deps.logger.debug(
-    `Getting all bulk import jobs (apiVersion=${apiVersion}, search=${search}, page=${pageNumber}, size=${pageSize})..`,
-  );
+  const sortColumn = queryParams?.sortColumn;
+  const sortOrder = queryParams?.sortOrder;
```
dzemanov, line 103: You have defined in `openapi.yaml` both `sortOrderQueryParam` and `sortColumnQueryParam` to have default values if user does not provide them. But they are not used (they are not set automatically to default if undefined).
 So either remove default values - in that case, we need to handle invalid option when user does not provide both of these parameters. Or I would suggest adding them here and then remove the change with `if (sortColumn)`

workspaces/bulk-import/plugins/bulk-import-backend/src/schema/openapi.yaml
```diff
@@ -360,6 +361,38 @@ components:
...
+      description: |
+        The allowed values for sorting columns:
+        - `repository.name`: Sort by repository name.
+        - `repository.organization`: Sort by organization URL.
+        - `repository.url`: Sort by repository URL.
+        - `lastUpdate`: Sort by the last time the catalog-info.yaml was updated.
+        - `status`: Sort by the status of the catalog-info.yaml.
+      required: false
+      schema:
+        enum:
```
dzemanov, line 376: Please add type also here:
```suggestion
      schema:
        type: string
        enum:
```

dzemanov (changes requested):
Tested with all sort values, on the backend and frontend. Works as expected except for when `sortColumn` is not specified. Please run `prettier` to fix failing tests :).

dzemanov:
Also, please add [changeset](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets).

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
rm3l: Indeed, IIRC, the `/imports` endpoint fetches Catalog locations first and only keeps those reachable from the configured GH integrations. Initially, it was querying GH first, but this was causing major performance issues, as reported in https://issues.redhat.com/browse/RHIDP-3680

Also, note that the imports may be coming from different sources (app-config, repos registered manually in Backstage, auto-discovery, ...), and the user may have configured several GH integrations (like different GH apps installed in multiple orgs).
That's essentially why the backend needed to aggregate such various sources to properly implement not only sorting, but also search and pagination, and return what the user is requesting.
But I guess pagination should help reduce the number of data fetched and iterated over.

That being said, open to better suggestions.

workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.ts
```diff
@@ -45,18 +49,31 @@ type FindAllImportsResponse =
   | Components.Schemas.Import[]
   | Components.Schemas.ImportJobListV2;
 
-function sortImports(imports: Components.Schemas.Import[]) {
+function sortImports(
```
rm3l, line 52: Could you add some tests in the backend for this feature? At least a few unit tests for this function. Or, better, some more complete tests in https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/bulkImports.test.ts and/or https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/bulk-import/plugins/bulk-import-backend/src/service/handlers/import/imports.test.ts

its-mitesh-kumar: added unit tests for it .

its-mitesh-kumar:
@dzemanov I have resolved all the comments , CI job failing please try prettier:fix and push it from your system , its not getting resolve from end .

dzemanov:
@its-mitesh-kumar it is failing on type check in `bulkImports.test.ts` which needs to be updated.

workspaces/bulk-import/plugins/bulk-import-backend/src/schema/openapi.yaml
```diff
@@ -360,6 +361,40 @@ components:
...
+      in: query
+      name: sortOrder
+      description: The order of sorting asc for ascending or desc for descending
+      required: false
+      schema:
+        type: string
+        enum:
+          - asc
+          - desc
+          - ''
```
dzemanov, line 395:
```suggestion
```

workspaces/bulk-import/plugins/bulk-import-backend/src/schema/openapi.yaml
```diff
@@ -360,6 +361,40 @@ components:
...
+        - `status`: Sort by the status of the catalog-info.yaml.
+      required: false
+      schema:
+        enum:
+          - repository.name
+          - repository.organization
+          - repository.url
+          - lastUpdate
+          - status
+          - ''
```
dzemanov, line 382:
```suggestion
```

dzemanov: Does it make sense to sort by empty column name? What does it mean sorting by empty column name?

workspaces/bulk-import/plugins/bulk-import-backend/src/generated/openapi.d.ts
```diff
@@ -328,17 +338,26 @@ declare namespace Paths {
...
       page?: Parameters.Page;
       size?: Parameters.Size;
+      sortOrder?: Parameters.SortOrder;
+      sortColumn?: Parameters.SortColumn;
       search?: Parameters.Search;
     }
     namespace Responses {
-      export type $200 =
-        /* Import Job with source it originates from */
+      export type $200 = /* Import Job with source it originates from */
```
dzemanov, line 360: > @dzemanov Now CI is failing due to prettier issue

I got permission denied for making changes. Just remove this change and prettier should be fixed :).

workspaces/bulk-import/plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
```diff
@@ -37,20 +40,36 @@ export const RepositoriesList = () => {
...
   const { openDrawer, setOpenDrawer, drawerData } = useDrawer();
   const [pageNumber, setPageNumber] = React.useState(0);
   const [rowsPerPage, setRowsPerPage] = React.useState(5);
   const [debouncedSearch, setDebouncedSearch] = React.useState('');
 
+  const orderByColumn = React.useMemo(() => {
+    const column = orderBy?.replace(/\.([a-zA-Z])/g, (_, char) =>
+      char.toUpperCase('en-US'),
+    ) as keyof typeof AddedRepositoryColumnNameEnum;
+    return column === 'repositoryName' ? 'repoName' : column;
```
dzemanov, line 59: Why do we need to check here for repositoryName?

its-mitesh-kumar: don't remember it exactly , it was due to two different keys for name column was coming for name column , so I needed to handle it .

workspaces/bulk-import/plugins/bulk-import/src/types/types.ts
```diff
@@ -99,6 +99,19 @@ export enum RepositoryStatus {
   'CODEOWNERS_FILE_NOT_FOUND_IN_REPO' = 'CODEOWNERS_FILE_NOT_FOUND_IN_REPO',
 }
 
+export enum AddedRepositoryColumnNameEnum {
+  repoName = 'repository.name', // because of property name
+  repositoryName = 'repository.name',
```
dzemanov, line 104: This can be deleted now:

```suggestion
```

workspaces/bulk-import/plugins/bulk-import/src/components/Repositories/RepositoriesList.tsx
dzemanov: Oh, alright, this is happening only when we have initial state here:
```
  const [orderBy, setOrderBy] = React.useState<string>(
    AddedRepositoryColumnNameEnum.repoName,
  );
```
which is giving `repository.name`.

You can use:
```
  const [orderBy, setOrderBy] = React.useState<string>('repoName');
```
and delete this check or leave it as it is.

its-mitesh-kumar: Thanks for your input , it was great help.

workspaces/bulk-import/plugins/bulk-import/src/types/types.ts
dzemanov, line 104:
```suggestion
```

dzemanov (changes requested):
@its-mitesh-kumar Thank you for all the changes! Enum key `repositoryName` can be now safely deleted. Tested again with the changes and works fine except for one bug I found, but not sure it it is related to this PR:

https://github.com/user-attachments/assets/8c871502-1341-47c3-8f6b-96bfb4abe9f3

its-mitesh-kumar:
@dzemanov ! Quality gate is failing due to its considering the objects of **locationUrls** as duplicate of each other .
Screenshot 2025-01-13 at 4 17 06 PM (2)
](url)

dzemanov (approved):

dzemanov:
/lgtm
However, @its-mitesh-kumar you will need to request review from [CODEOWNERS](https://github.com/redhat-developer/rhdh-plugins/blob/main/.github/CODEOWNERS#L8)

its-mitesh-kumar:
@christoph-jerolimov @debsmita1 can we get this merge ?

debsmita1 (changes requested):

debsmita1:
When I click on a column to change the order, it triggers the API with the current column name and order along with the previous column name and order





https://github.com/user-attachments/assets/22cdc946-210e-4116-865d-11ae15e2fb24

its-mitesh-kumar:
@debsmita1 , That's happening due to some odd behavior of useQuery , What I can see if we updating value of any key of queryKey , its recalling queryFunction once with old data and once with new value . We can fix this later , I have raised[ bug](https://issues.redhat.com/browse/RHIDP-5626) for the same .

debsmita1 (approved):
/lgtm

debsmita1:
@its-mitesh-kumar  Could you resolve this https://github.com/redhat-developer/rhdh-plugins/pull/178#discussion_r1886651344

PR #187, opened 2024-12-14
fix(theme): add mui v4 and v5 grid examples

christoph-jerolimov:

This PR adds

1. some new Grid examples to **test plugins** to see and understand the difference between Backstage and RHDH. But there is none; both look except different colors and borders really similar!
2. add workaround to load mui v5 theme correctly when package test plugins

Backstage defines a default spacing for all Grids, see:

https://github.com/backstage/backstage/blob/3d2e95f71c2be9a63530697cf01e482e4f50f98d/packages/theme/src/v5/defaultComponentThemes.ts#L67-L71

The RHDH theme doesn't change the padding or default spacing. But in MUI 5 it's required to add additional padding via spacing to the `Grid container` or to the content (in the `Grid item`). 

The UI will looks unaligned when a padding is added to the `Grid item`.

### MUI v4 and v5 looks similar if no padding is added to the `Grid item`

image

image

### But it's not okay to add padding to `Grid item` in MUI v5:

MUI v4

image

MUI v5

image

### DIFF
--- a/workspaces/theme/plugins/bc-test/package.json
+++ b/workspaces/theme/plugins/bc-test/package.json
@@ -35,7 +35,8 @@
   "dependencies": {
     "@backstage/core-components": "^0.15.1",
     "@backstage/core-plugin-api": "^1.10.0",
-    "@backstage/plugin-user-settings": "^0.8.16"
+    "@backstage/plugin-user-settings": "^0.8.16",
+    "@mui/material": "^5"
   },
   "peerDependencies": {
     "react": "^16.13.1 || ^17.0.0 || ^18.0.0"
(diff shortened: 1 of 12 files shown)


debsmita1 (changes requested):
Could you update the year in the notice to `2025` ? 
otherwise, it looks good to me.

christoph-jerolimov:
> Could you update the year in the notice to `2025` ? otherwise, it looks good to me.

@debsmita1 but the code is from 2024. :) Kashish created also #235 for this, so I like to keep it for now and update it later if needed after the other PR is merged (or this PR will update this headers as well if needed). But my understanding is that the year is when the code is written and shouldn't be updated to the current year.

debsmita1 (approved):
/lgtm

PR #202, opened 2024-12-16
fix(theme): fix font loading after migrating theme

christoph-jerolimov:

TODO


workspaces/theme/plugins/theme/package.json
ciiay, line 1: The current `mui/material` version is incompatible with showcase's mui/material version. If we keep this will it be a problem when we switch to this theme plugin dependency from the `"@redhat-developer/red-hat-developer-hub-theme"` one?
```
  "dependencies": {
    "mui/material": "^5.0.0"
  },
```

ciiay (commented):
Verified with yarn link on local showcase and it works as expected. The `mui/material` version is not compatible to the showcase one. Did you have to match the version or it worked out with no extra step for you?

workspaces/theme/plugins/theme/package.json
christoph-jerolimov: Hi, why is this incompatible? ^5.0.0 means it supports all 5.x.x versions, right? But I already thought I might switch to a peer dependency to ensure we don't load the dependency twice.

christoph-jerolimov: Maybe we can give https://github.com/redhat-developer/rhdh-plugins/pull/205 a chance? :)

ciiay:
/lgtm

Thank you 🤝

debsmita1 (approved):

PR #204, opened 2024-12-18
fix(theme): added a workaround so that defaultProps are also picked up from MUI v5 components (like Button disabledRipple or Grid spacing)

christoph-jerolimov:


ciiay (approved):
/lgtm

I wasn't able to reproduce the issue, but verified this solution doesn't break anything in local showcase 👍 Let's test out when the new version is out.

PR #205, opened 2024-12-18
fix(theme): move the mui v5 package to a peer dependency to ensure that we don't load it twice

christoph-jerolimov:

To be honest, this is more like an experiment. Maybe we can give it a try to check this with a new theme release (which isn't used yet) and rhdh-showcase?


ciiay (approved):
/lgtm

This resolves the issue I mentioned in #202 comment. Thanks for the fix 🚀

PR #212, opened 2024-12-19
chore: unify prettier to 3.4.2

mareklibra:
Unifies prettier to a single 3.4.2 version through-out the whole repo.

Different prettier versions give different results on the same sources which is causing issues/confusion when switching between local dev and CI envs.

The first commit is result of a manual change, the later one is the result of calling the prettier from the root-level and then from every workspace.


batzionb (approved):

workspaces/orchestrator/plugins/orchestrator-backend/static/generated/envelope/combined.bundle.js
christoph-jerolimov, line 1: You want that this files are part of the repo? It looks like it's formatted also if you have added the generated line to `.prettierignore`?

workspaces/orchestrator/plugins/orchestrator-backend/static/generated/envelope/diagram/org.kie.workbench.common.stunner.sw.KogitoSWEditor/css/bootstrap-3.4.1.min.cache.css
christoph-jerolimov, line 1: It's you plugin but I would recommend to exclude this generated files. Wdyt?

workspaces/repo-tools/packages/cli/src/lib/workspaces/templates/workspace/.gitignore.hbs
```diff
@@ -1,54 +1,9 @@
...
-
-# Sensitive credentials
-*-credentials.yaml
-
-# vscode database functionality support files
-*.session.sql
-
-# E2E test reports
-e2e-test-report/
+# macOS .DS_Store # Logs logs *.log npm-debug.log* yarn-debug.log*
```
christoph-jerolimov, line 1: This change (removing new lines) looks invalid to me.

gciavarrini (approved):
lgtm

workspaces/orchestrator/plugins/orchestrator-backend/static/generated/envelope/combined.bundle.js
christoph-jerolimov, line 1: Based on the generated in .prettierignore I'm surprised to see this. I think you don't want format this.

workspaces/repo-tools/packages/cli/src/lib/workspaces/templates/workspace/catalog-info.yaml.hbs
```diff
@@ -1,14 +1,6 @@
...
-  #   github.com/project-slug: backstage/backstage
-  #   backstage.io/techdocs-ref: dir:.
-spec:
-  type: website
-  owner: john@example.com
-  lifecycle: experimental
-  
+apiVersion: backstage.io/v1alpha1 kind: Component metadata: name:
+{{name}}
+description: An example of a Backstage application. # Example for optional
```
christoph-jerolimov, line 3: That looks also broken to me.

Can you ignore *.hbs files?

workspaces/orchestrator/plugins/orchestrator-backend/static/generated/envelope/combined.bundle.js
gciavarrini: Good catch! Also all the files in [orchestrator-common/src/generate](https://github.com/redhat-developer/rhdh-plugins/pull/212/files#diff-1ff6e60734aeec83e00940c38177b01c4d130f0443402955cfd7ae3960d42a76) should be ignored too.

gciavarrini (changes requested):
ignore all the generated files and revert the changes

workspaces/orchestrator/plugins/orchestrator-backend/static/generated/envelope/diagram/org.kie.workbench.common.stunner.sw.KogitoSWEditor/css/bootstrap-3.4.1.min.cache.css
mareklibra: If so then in a separate PR

package.json
```diff
@@ -15,25 +15,50 @@
   "type": "module",
   "packageManager": "yarn@3.8.7",
   "devDependencies": {
+    "@backstage/catalog-model": "^1.7.2",
```
christoph-jerolimov, line 18: I think we should not add this kind of dependencies to the root package.json

christoph-jerolimov, line 18: I think we should not add this kind of dependencies to the root package.json

package.json
```diff
@@ -15,25 +15,50 @@
...
     "@octokit/rest": "^20.1.1",
     "@red-hat-developer-hub/cli": "portal:./workspaces/repo-tools/packages/cli",
     "@spotify/eslint-plugin": "^15.0.0",
     "@spotify/prettier-config": "^15.0.0",
+    "ajv-formats": "^3.0.1",
     "array-to-table": "^1.0.1",
+    "axios": "^1.7.9",
+    "chalk": "^5.4.0",
+    "commander": "^12.1.0",
+    "cross-fetch": "^4.0.0",
```
christoph-jerolimov, line 37: Shouldn't be needed here.

christoph-jerolimov (changes requested):
Hi @mareklibra, this looks much better now. But can you take a look into the root package.json and yarn.lock?

mareklibra:
I don't see a commit adding those changes.
I will rebase due to merge conflicts, I guess those changes will disappear then.

christoph-jerolimov (approved):
Thanks for this PR and cleaning it up again and again :+1: 

Lgtm!

PR #214, opened 2024-12-19
feat(fab): add global floating button

debsmita1:

**Resolves:**
https://issues.redhat.com/browse/RHIDP-5298

**GIF:**

https://github.com/user-attachments/assets/afc6c64e-7928-4cb7-9a07-d7c42a6c09d2

**TODO:**
_Add more position options for the floating button_


workspaces/global-floating-action-button/.changeset/sweet-clocks-move.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-floating-action-button': major
```
christoph-jerolimov, line 2: I would start with a patch release 0.0.1 until we're comfortable. I expect that we will have more then just this PR to have it ready.

workspaces/global-floating-action-button/packages/app/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "@backstage/plugin-techdocs": "^1.11.0",
+    "@backstage/plugin-techdocs-module-addons-contrib": "^1.1.16",
+    "@backstage/plugin-techdocs-react": "^1.2.9",
+    "@backstage/plugin-user-settings": "^0.8.14",
+    "@backstage/theme": "^0.6.0",
+    "@mui/icons-material": "5.15.17",
+    "@mui/lab": "5.0.0-alpha.173",
+    "@mui/material": "5.15.17",
+    "@mui/styles": "5.16.7",
+    "@red-hat-developer-hub/backstage-plugin-global-floating-action-button": "^0.1.0",
```
christoph-jerolimov, line 49: This should be workspace dependency so that it doesn't use the released package from npm:

```suggestion
    "@red-hat-developer-hub/backstage-plugin-global-floating-action-button": "workspace:^",
```

workspaces/global-floating-action-button/packages/app/src/components/Root/Root.tsx
```diff
@@ -0,0 +1,149 @@
...
+    </div>
+  );
+};
+
+export const Root = ({ children }: PropsWithChildren<{}>) => {
+  const { state, toggleModal } = useSearchModal();
+
+  return (
+    <SidebarPage>
+      <GlobalFloatingButton
```
christoph-jerolimov, line 85: :partying_face:

workspaces/global-floating-action-button/plugins/global-floating-action-button/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@red-hat-developer-hub/backstage-plugin-global-floating-action-button",
+  "version": "0.1.0",
```
christoph-jerolimov, line 3: To start with 0.0.1 you must change this to 0.0.0 :)

workspaces/global-floating-action-button/plugins/global-floating-action-button/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "prettier:fix": "prettier --ignore-unknown --write .",
+    "ui-test": "start-server-and-test start localhost:3000 'playwright test'"
+  },
+  "dependencies": {
+    "@backstage/core-components": "^0.15.1",
+    "@backstage/core-plugin-api": "^1.10.0",
+    "@backstage/theme": "^0.6.0",
+    "@material-ui/core": "^4.12.2",
+    "@material-ui/icons": "^4.9.1",
+    "@material-ui/lab": "4.0.0-alpha.61",
```
christoph-jerolimov, line 41: Do we use some of this dependencies? If possible it would be great if we can use MUI v5 only?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+import * as React from 'react';
+import classnames from 'classnames';
+import Fab from '@mui/material/Fab';
+import { makeStyles, Typography } from '@material-ui/core';
+import { useTheme } from '@mui/material/styles';
+import CloseIcon from '@mui/icons-material/Close';
+import Tooltip from '@mui/material/Tooltip';
+import Slide from '@mui/material/Slide';
+
+type FloatingButtonAction = {
```
christoph-jerolimov, line 28: Should this be a `FloatingActionButton`? :)

And I think we should export this and if you agree move it to `src/types.ts`? Wdyt?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+    | 'primary'
+    | 'secondary'
+    | 'success'
+    | 'warning'
+    | string;
+  onClick?: React.MouseEventHandler;
+  url?: string;
+  label?: string | React.ReactNode;
+  toolTip?: string;
+  actions?: FloatingButtonAction[];
```
christoph-jerolimov, line 48: Is this a recursion? Could we handle this only in this package without exporting this.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+const fab = (
+  icon: React.ReactElement,
+  size: 'small' | 'medium' | 'large' | string,
+  color?: any,
+  label?: string | React.ReactNode,
+  onClick?: any,
+  url?: string,
+  tooltip?: string,
+) => {
```
christoph-jerolimov, line 69: Feels like a component, but isn't one. Couldn't yo convert this to a component where a FAB (your FAB type) is a prop?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+  return (
+    <Tooltip title={tooltip} placement="left">
+      <div>
+        <Fab
+          id="add"
```
christoph-jerolimov, line 74: Is id required?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,197 @@
...
+          variant={label ? 'extended' : 'circular'}
+          size={(size as any) || 'medium'}
+          color={(color as any) || 'default'}
+          aria-label="add"
```
christoph-jerolimov, line 78: Maybe we should make the label mandatory for the a11y label and then add another FAB option `showLabel: boolean` which is false by default? Wdyt?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/plugin.ts
```diff
@@ -0,0 +1,54 @@
...
+import {
+  createComponentExtension,
+  createPlugin,
+  createRoutableExtension,
+} from '@backstage/core-plugin-api';
+import { createRouteRef } from '@backstage/core-plugin-api';
+
+const testRouteRef = createRouteRef({
+  id: 'test-global-floating-action-button',
+});
```
christoph-jerolimov, line 25: Similar to header, I guess we can live without any route

christoph-jerolimov (changes requested):
That's a great start!! Thanks.

Of course, this initial PR must not be perfect and I saw it's still WIP.

I added some feedback anyway. Feel free to pick what makes sense to you as well or let us discuss that next year.

You can merge in the meantime of course, but please don't release a 1.0.0 so early.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/plugin.ts
debsmita1: I created this Route for the test environment, the floating button plugin has no route

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,158 @@
...
+              return (
+                <FAB
+                  icon={fb.icon}
+                  size={fb.size || 'medium'}
+                  color={fb.color}
+                  label={fb.label}
+                  onClick={fb.onClick}
+                  url={fb.url}
+                  tooltip={fb.toolTip}
+                  showLabel={fb.showLabel}
```
christoph-jerolimov, line 149: Here as well. I think it will be cleaner if you pass one prop fab or action=fb.

I'm think calling everything FAB is confusing. That's why I bring up action here.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export enum Position {
+  bottomRight = 'bottomRight',
+  bottomCenter = 'bottomCenter',
+}
```
christoph-jerolimov, line 20: I know that's now really opinionated, and I don't know if we have any rule for that, so I checked in backstage code and found less enums then expected. But at least this two:

https://github.com/backstage/backstage/blob/6b92433d5e68bc57174ed9c6546d085639f82dee/plugins/scaffolder-backend-module-gitlab/report.api.md?plain=1#L266-L284

https://github.com/backstage/backstage/blob/6b92433d5e68bc57174ed9c6546d085639f82dee/plugins/catalog-graph/src/components/EntityRelationsGraph/types.ts#L110-L127

So I would suggest at least uppercase KEYS and - or _ as delimiter for the values. I'm thinking how to define that in YAML later.

```suggestion
export enum Position {
  BOTTOM_RIGHT = 'bottom-right',
  BOTTOM_CENTER = 'bottom-center',
}
```

Btw. I would call slot and "page end instead of right to support (somewhere in the future) RTL.

```suggestion
export enum Slot {
  PAGE_END = 'page-end',
  BOTTOM_CENTER = 'bottom-center',
}
```

But that's all just loud thinking. :D

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type FloatingActionButtonWithPositions = Record<
+  Position | string,
+  FloatingActionButton[]
+>;
```
christoph-jerolimov, line 25: That's an internal type I would just define in the component (props) that needs this.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export type FloatingActionButton = {
+  label: string;
+  showLabel?: boolean;
+  icon: React.ReactElement;
```
christoph-jerolimov, line 30: As mentioned above, we should allow a string here as well. Just here again to see that it's also resolved here. I think we can make it optional? There should be at least a label then. So render the label when `showLabel || !icon` ;-)

```suggestion
  icon?: string | React.ReactElement;
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
```diff
@@ -0,0 +1,49 @@
...
+    | 'error'
+    | 'info'
+    | 'inherit'
+    | 'primary'
+    | 'secondary'
+    | 'success'
+    | 'warning'
+    | string;
+  onClick?: React.MouseEventHandler;
+  url?: string;
```
christoph-jerolimov, line 44: We should also support links within the backstage instance like `/settings`. That's why I would suggest `to` here.

The Backstage `Link` component for example automatically detects external links and open them in a new page.

See also: https://github.com/backstage/backstage/blob/master/packages/core-components/src/components/Link/Link.tsx

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/utils.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+    url: 'https://github.com/xyz',
+    priority: 200,
+    excludeOnPaths: ['/test-pathname'],
+  },
+];
+
+describe('Global floating action button utils', () => {
+  it('should sort actions based on priority', () => {
+    const buttons = sortButtonsWithPriority(floatingButtons);
+    expect(buttons).toEqual([
```
christoph-jerolimov, line 64: Utils are easy to test, that's nice :partying_face:

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/utils.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+import {
+  evaluateFloatingButtonsWithPositions,
+  filterAndSortButtons,
+  sortButtonsWithPriority,
+} from './utils';
+
+const floatingButtons = [
+  {
+    color: 'success',
+    icon: <>Some Icon</>,
```
christoph-jerolimov, line 26: If you allow the string, you can rename this file to utils.test.ts and remove the React dependency. :)

christoph-jerolimov (changes requested):
I know, a lot of small comments and asks again. But also again, it looks really good so far and it's a big step forward since yesterday! There is no reason to rush, and I know some of these points are maybe already on your list. Dropped my thoughts here anyway:

workspaces/global-floating-action-button/packages/app/knip-report.md
```diff
@@ -0,0 +1,20 @@
+# Knip report
+
+## Unused dependencies (5)
+
+| Name                                                                  | Location     | Severity |
+| :-------------------------------------------------------------------- | :----------- | :------- |
+| @red-hat-developer-hub/backstage-plugin-global-floating-action-button | package.json | error    |
```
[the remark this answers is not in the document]
debsmita1: yes, I checked this is present in the knip report of the other plugins as well

workspaces/global-floating-action-button/plugins/global-floating-action-button/dev/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import React from 'react';
+import { createDevApp } from '@backstage/dev-utils';
+import {
+  globalFloatingActionButtonPlugin,
+  TestFloatingActionButtonPage,
```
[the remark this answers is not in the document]
debsmita1: The ExampleComponent page can be scrolled and has cards in it

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
debsmita1: I am using this as a return type of a util that evaluates all the buttons according to their positions

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
debsmita1: @christoph-jerolimov  I was under the impression that the  icon is mandatory, if the user wants to show a label they can do so and that will appear next to the icon

workspaces/global-floating-action-button/plugins/global-floating-action-button/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "keywords": [
+    "lifecycle:active",
+    "backstage",
+    "plugin"
+  ],
+  "homepage": "https://red.ht/rhdh",
+  "bugs": "https://github.com/redhat-developer/rhdh-plugins/issues",
+  "maintainers": [
+    "@rm3l"
```
christoph-jerolimov, line 70: Feel free to add yourself here ;)

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButtons.tsx
```diff
@@ -0,0 +1,39 @@
...
+  floatingButtons: FloatingActionButton[];
+}) => {
+  const floatingButtonMap =
+    evaluateFloatingButtonsWithPositions(floatingButtons);
+
+  return (
+    <>
+      {Object.keys(floatingButtonMap).map(fb => (
+        <FloatingButton
+          position={fb as Slot}
```
christoph-jerolimov, line 33: Hmmmm, see issue is that Object.keys returns a list of strings, not a list of Slots here.

I'm a big fan of Record, but maybe it's worth to change the `FloatingActionButtonWithPositions` type here to an array so that it's easier to handle this in the React world:

```tsx
export type FloatingActionButtonWithPositions = Array<{
  slot: Slot,
  actionButtons: FloatingActionButton[],
}>;
```

I'm not sure because this makes the util more "complex" and you might also need to use the slot as key here.

Btw, the key is missed here anyway.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export {
+  GlobalFloatingButton,
+  globalFloatingActionButtonPlugin,
+} from './plugin';
```
christoph-jerolimov, line 19: I guess you can savely export everything from the plugin and also the types:

```suggestion
export * from './plugin';
export * from './types';
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/plugin.ts
```diff
@@ -0,0 +1,36 @@
...
+ * @public
+ * Floating Button Plugin
+ */
+export const GlobalFloatingButton = globalFloatingActionButtonPlugin.provide(
+  createComponentExtension({
+    name: 'GlobalFloatingButton',
+    component: {
+      lazy: () => import('./components').then(m => m.FloatingButtons),
+    },
+  }),
```
christoph-jerolimov, line 35: I think you should eighter rename the export or the file here. So that both matches. This makes it much easier in the future to understand the plugin.

Also, please use the full path of that component in the import('...') statement so allow the build process build smaller build artifacts / chunks.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
```diff
@@ -0,0 +1,50 @@
...
+  Slot | string,
+  FloatingActionButton[]
+>;
+
+export type FlexDirection = 'row' | 'row-reverse' | 'column' | 'column-reverse';
+
+export type FloatingActionButton = {
+  label: string;
+  showLabel?: boolean;
+  icon: string | React.ReactElement;
```
christoph-jerolimov, line 32: As mentioned before, I recommend to make the icon optional:

```suggestion
  icon?: string | React.ReactElement;
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+    return 0;
+  });
+};
+
+export const filterAndSortButtons = (
+  floatingButtons: FloatingActionButton[],
+  pathname: string,
+) => {
+  const sortedButtons = sortButtonsWithPriority(floatingButtons);
+  return sortedButtons.filter(fb => {
```
christoph-jerolimov, line 61: Just a nit and I'm sure you can't measure it with our 3 floating buttons. But I would always first filter and then sort elements to reduce the number of sorting elements. :man_shrugging:

christoph-jerolimov (changes requested):
Some small nits:

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
christoph-jerolimov: Got that. You will notice that when you export all the types from types.ts, that this is a type you might want move the the util, otherwise it will be visible for all plugin users. Which is also okay I think... Keep it where it is. I mark this as resolved. :)

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/types.ts
christoph-jerolimov: I think less is more and in this case this means: It's easier for an admin to a (temporary) FAB without an icon then thinking on "where I get this icon from".

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <Slide
+          container={ref}
+          easing={{
+            enter: theme.transitions.easing.easeOut,
+            exit: theme.transitions.easing.sharp,
+          }}
+        >
+          <>
+            {fabs?.map(fb => {
+              return <FAB actionButton={fb} size="medium" key={fb.label} />;
```
christoph-jerolimov, line 104: Please add a key for react

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -0,0 +1,95 @@
...
+  },
+  'page-end': {
+    bottom: '60px',
+    right: '60px',
+  },
+  'bottom-center': {
+    bottom: '60px',
+    left: '50%',
+  },
+}));
```
christoph-jerolimov, line 42: You could use the theme, like:

```suggestion
const useStyles = makeStyles((theme) => ({
  button: {
    zIndex: 200,
    display: 'flex',
    position: 'fixed',
    maxWidth: '150px',
    gap: '10px',
  },
  'page-end': {
    bottom: theme.spacing(4),
    right: theme.spacing(4),
  },
  'bottom-center': {
    bottom: theme.spacing(4),
    left: '50%',
  },
}));
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
```diff
@@ -0,0 +1,111 @@
...
+          {...(newWindow ? { target: '_blank', rel: 'noopener' } : {})}
+          variant={actionButton.showLabel ? 'extended' : 'circular'}
+          size={size || actionButton.size || 'medium'}
+          color={actionButton.color || 'default'}
+          aria-label={actionButton.label}
+          onClick={actionButton.onClick}
+          href={actionButton.to}
+        >
+          <FabIcon icon={actionButton.icon as string} />
+          {actionButton.showLabel && (
```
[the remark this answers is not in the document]
debsmita1: Done
Screenshot 2025-01-03 at 6 51 26 PM

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
```diff
@@ -0,0 +1,111 @@
...
+          size="medium"
+          color="default"
+          onClick={handleClick}
+          aria-label="Menu"
+          variant="circular"
+        >
+          {isMenuOpen ? <CloseIcon /> : <MenuIcon />}
+        </Fab>
+      </Tooltip>
+      {isMenuOpen && (
```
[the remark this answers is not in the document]
debsmita1: I will address this issue in this ticket https://issues.redhat.com/browse/RHIDP-5398

christoph-jerolimov (approved):
Thanks @debsmita1!

This lgtm!

PR #221, opened 2024-12-23
fix(orchestrator):workflow name column values in workflow runs table …

LiorSoffer:
resolve [FLPATH-1967](https://issues.redhat.com/browse/FLPATH-1967)
image


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
```diff
@@ -55,7 +51,7 @@ export const WorkflowDetailsTabContent = ({
             <WorkflowEditor
               kind={EditorViewKind.EXTENDED_DIAGRAM_VIEWER}
               workflowId={workflowId}
-              format={workflowFormat}
+              format={workflowOverviewDTO?.format}
```
gciavarrini, line 54: what if `workflowOverviewDTO` is `undefined`?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
```diff
@@ -243,7 +243,6 @@ const RefForwardingWorkflowEditor: ForwardRefRenderFunction<
             if (format && workflowId && format !== workflowFormat) {
               const link = viewWorkflowLink({
                 workflowId: workflowId,
-                format: workflowFormat,
```
gciavarrini, line 246: do we stil need `format` in the `if` condition?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
```diff
@@ -243,7 +243,6 @@ const RefForwardingWorkflowEditor: ForwardRefRenderFunction<
             if (format && workflowId && format !== workflowFormat) {
               const link = viewWorkflowLink({
                 workflowId: workflowId,
```
batzionb, line 245: I'm pretty sure we can't just remove the format
Checking now why it's needed

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
```diff
@@ -35,11 +35,7 @@ export const WorkflowDetailsTabContent = ({
   loading,
   workflowOverviewDTO,
 }: Props) => {
-  const { workflowId, format } = useRouteRefParams(workflowRouteRef);
-  const workflowFormat = useMemo(
-    () => (format === 'json' ? 'json' : 'yaml'),
-    [format],
-  );
+  const { workflowId } = useRouteRefParams(workflowRouteRef);
```
batzionb, line 38: small improvement suggestion:
we can also take the workflow id from workflowOverviewDTO
this way the component becomes completely independent from the routing system, making it more reusable and isolated
WDYT?

batzionb:
@LiorSoffer 
I like this solution
It makes the links more concise and easy to use
Thanks!

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
batzionb: I see you take it from WorkflowOverviewDTO which is the better approach :)

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
batzionb: good point, please remove the format from the condition

batzionb: there's some kind of odd rerouting here...
this code handles a case where workflowFormat is different than the format in the URL
if this is the case, rerouting won't help
this whole code should remove

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
batzionb: I think the whole page should not show anything if workflowOverviewFormatDTO is undefined
It should be handled in WorkflowPage

batzionb: @LiorSoffer 
WDYT of WorkflowOverviewDTO property being mandatory
In case it's undefined WorkflowPage should show an error and no content

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowEditor/WorkflowEditor.tsx
batzionb: if the formats don't match we need to handle it differently
maybe show an error

batzionb: Took a look where format comes from
Backend calculates it
https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts#L194
Now that it comes from the backend, recalculating it in the FE is a waste
This whole code should be removed, including recalculating the format and handling the mismatch

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowDetailsTabContent.tsx
LiorSoffer: good idea

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/WorkflowPage.tsx
```diff
@@ -66,10 +66,12 @@ export const WorkflowPage = () => {
             loading={loading}
             canRun={canRun}
           >
-            <WorkflowDetailsTabContent
-              loading={loading}
-              workflowOverviewDTO={workflowOverviewDTO?.data}
-            />
+            {workflowOverviewDTO && (
```
batzionb, line 69: actually this might not work well, I think when in loading state workflowOverviewDTO is undefined, and this way we won't show the loading state (skeletons) that we have

batzionb: maybe better pass it, and add a more simple loading handling, if loading show <Progress>
if workflowOverviewDTO not defined, show null
make these changes in WorkflowDetailsTabContent and not here

batzionb (dismissed):

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -189,35 +191,37 @@ export const WorkflowRunsTabContent = () => {
...
-                capitalize(row.category ?? VALUE_UNAVAILABLE),
-              sorting: false,
+              render: (data: WorkflowRunDetail) =>
+                capitalize(data.category ?? VALUE_UNAVAILABLE),
             },
           ]),
-      { title: 'Started', field: 'start', customSort: applyBackendSort },
-      { title: 'Duration', field: 'duration', sorting: false },
+      { title: 'Started', field: 'start', defaultSort: 'desc' },
+      { title: 'Duration', field: 'duration' },
```
batzionb, line 222: please return sorting: false
sorting is not possible for duration

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -189,35 +191,37 @@ export const WorkflowRunsTabContent = () => {
...
-              render: (row: WorkflowRunDetail) =>
-                capitalize(row.category ?? VALUE_UNAVAILABLE),
-              sorting: false,
+              render: (data: WorkflowRunDetail) =>
+                capitalize(data.category ?? VALUE_UNAVAILABLE),
             },
           ]),
-      { title: 'Started', field: 'start', customSort: applyBackendSort },
-      { title: 'Duration', field: 'duration', sorting: false },
+      { title: 'Started', field: 'start', defaultSort: 'desc' },
```
batzionb, line 221: undo this change as well

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
LiorSoffer: should I return customSort: applyBackendSort  as well?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
batzionb: it should be customSort: applyBackendSort

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -191,33 +193,36 @@ export const WorkflowRunsTabContent = () => {
...
         ? []
         : [
             {
               title: 'Category',
               field: 'category',
-              render: (row: WorkflowRunDetail) =>
-                capitalize(row.category ?? VALUE_UNAVAILABLE),
-              sorting: false,
+              render: (data: WorkflowRunDetail) =>
+                capitalize(data.category ?? VALUE_UNAVAILABLE),
```
batzionb, line 219: return the sorting: false

batzionb (approved):

PR #223, opened 2024-12-24
chore(orchestrator): update devmode image

batzionb:
resolves broken developer environment, following sonataflow images moving from docker hub to quay


gciavarrini (approved):
/lgtm

PR #211, opened 2024-12-19
feat(orchestrator): workflow instances pagination

batzionb:
Implemented frontend pagination of worklfow instances page
Resolves https://issues.redhat.com/browse/FLPATH-1534

Contains the followin backend API fixes:

1. Fixed broken status filter - it always failed with `Unsupported operator... `
2. Added propagation of errors in case graphql query is invalid, instead of returning an empty list of instances
3. Improved error messages in case of unsupported operator 
4. Removed fetching instances total count, which can't be implemented by graphql query and the way it's done, fetching all the ids, defeats the whole purpose of the pagination
5. Since filters must contain the same values and the same field name as Sonata, the field name status was changed to state and the enum values were changed to capital, same as in sonata. This removes the need for a lot of code mapping values back and forth, and reduces the confusion of what values to pass to pagination.
6. /instances API request type included paginationInfo in two levels of the request structure

Includes also adding task "yarn dev:debug", I found myself needing to google how to run backstage in debug mode many times... so I thought this is useful

image

PR #224, opened 2024-12-24
fix(orchestrator): multiple bug fixes following pagination PR

batzionb:
Multiple bug fixes in instances table following https://github.com/redhat-developer/rhdh-plugins/pull/211

1. Each click on sort resulted in two calls to backend, one with previous sort direction and one with new one, resulting in inconsistent results. Fix is by disabling useSwr revalidateOnFocus.
2. Sorting didn't display results in correct order, since material-table overrides the order from backend with simple sort by string. The only way to resolve this was to force the additional sorting to use the same order received from backend.
3. Sort by workflow name failed because it sent to backend the field 'name' instead of 'processName'
4. Sorting was enabled for fields that aren't supported by backend: category and duration

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
@@ -36,19 +36,17 @@ import { WorkflowVariablesViewer } from './WorkflowVariablesViewer';
 export const mapProcessInstanceToDetails = (
   instance: ProcessInstanceDTO,
 ): WorkflowRunDetail => {
-  const name = instance.processName || instance.processId;
   const start = instance.start ? moment(instance.start) : undefined;
   let duration: string = VALUE_UNAVAILABLE;
   if (start && instance.end) {
     const end = moment(instance.end);
     duration = moment.duration(start.diff(end)).humanize();
   }
-
   const started = start?.toDate().toLocaleString() ?? VALUE_UNAVAILABLE;
 
   return {
     id: instance.id,
-    name,
+    processName: instance.processName || VALUE_UNAVAILABLE,
     workflowId: instance.processId,
     start: started,
     duration,
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunDetail.ts
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunDetail.ts
@@ -20,7 +20,7 @@ export interface WorkflowSuggestion {
 
 export type WorkflowRunDetail = {
   id: string;
-  name: string;
+  processName: string;
   workflowId: string;
   state?: string;
   start: string;
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
@@ -72,57 +72,9 @@ export const WorkflowRunsTabContent = () => {
   const [statusSelectorValue, setStatusSelectorValue] = useState<string>(
     Selector.AllItems,
   );
-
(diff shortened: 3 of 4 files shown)


batzionb:
@gciavarrini 
I think no changeset required since static plugins weren't released yet with the pagination

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -48,7 +48,7 @@ export const mapProcessInstanceToDetails = (
 
   return {
     id: instance.id,
-    name,
+    processName: name,
```
gciavarrini, line 51: I'm ok to have this as quick fix, but please open a jira issue to fix this in the backend. it shoudn't be FE resposnibility to properly map orchestrator objects into graphql ones.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunDetail.ts
```diff
@@ -20,7 +20,7 @@ export interface WorkflowSuggestion {
 
 export type WorkflowRunDetail = {
   id: string;
-  name: string;
+  processName: string;
```
gciavarrini, line 23: same as above

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -202,7 +153,67 @@ export const WorkflowRunsTabContent = () => {
     ),
     [statusSelectorValue],
   );
-
+  
+  const applyBackendSort = React.useCallback((item1: WorkflowRunDetail, item2: WorkflowRunDetail): number => {          
+    // Workaround for material-table applying sorting on top of backend sorting. The version we are using is too old to request a fix. 
+    // Should be resolved when upgrading backstage and all plugins to material6
```
gciavarrini, line 159: create a jira for this as well and add the link in this comment.
 I'd like to avoid to forget fixing this.

gciavarrini (approved):
This pr LGTM, but please create jira items as per my comments to track future works.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
batzionb: I think the best approach is for sonata, backend, and FE to all use the same propery names, otherwise the code becomes impossible to follow, and it's hard to understand what field names to use.
In this case the backend returns the correct field name processName, so the only fix required is for the UI to use it as is instead of converting it to "name".

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
batzionb: I'm not sure there's a point having a Jira issue opened in the backlog that is not solvable until upgrade to material6. I can turn this comment into a TODO.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
gciavarrini: The backend should decouple SonataFlow and orchestrator frontend. Frontend should rely only on OpenAPI spec. What if sonata decide to change one of the property name? We need to change both backend and frontend. I understand that is easy on the short term but I don't think it's a good practice in the long run.

batzionb: What are the chances they change their property names and entire API?
It's really confusing working with different property names, and makes the code harder to maintain. If something fails in our API we want to check if it fails there and then need to remember to change the field names...
I would appreciate more opinions on this

PR #230, opened 2025-01-01
fix(orchestrator): Implement inputs card in workflow instance page

LiorSoffer:
resolve [FLPATH-1976](https://issues.redhat.com/browse/FLPATH-1976)
image
image

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInputs.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInputs.tsx
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+
+import {
+  InfoCard,
+  Progress,
+  ResponseErrorPanel,
+  StructuredMetadataTable,
+} from '@backstage/core-components';
+
+export const WorkflowInputs: React.FC<{
+  className: string;
+  cardClassName: string;
+  value: any;
+  loading: any;
+  responseError: any;
+}> = ({ className, cardClassName, value, loading, responseError }) => {
+  const inputs = value?.data;
+  return (
+    <InfoCard
+      title="Inputs"
+      subheader={
+        !loading && !inputs && <i>The workflow instance has no inputs</i>
+      }
+      divider={false}
+      className={className}
+      cardClassName={cardClassName}
+    >
+      {loading ? <Progress /> : null}
+
+      {!loading && responseError && (
+        <ResponseErrorPanel error={responseError} />
+      )}
+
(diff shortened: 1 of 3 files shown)


batzionb:
@LiorSoffer 
The schema can't be passed from the previous page
A page needs to get all it's data independently
The instance page is navigated to from multiple places, and also needs to be able to show the data after clicking f5
The instance page should call the inputSchema API same as ExecuteWorkflowPage does 
here https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/plugins/orchestrator/src/components/ExecuteWorkflowPage/ExecuteWorkflowPage.tsx#L67

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -87,14 +93,22 @@ export const WorkflowInstancePageContent: React.FC<{
...
     };
-    if (instanceVariables.hasOwnProperty('result')) {
-      delete instanceVariables.result;
+
+    const { schema } = location.state || {};
+
+    // Ensure schema properties are an object and iterate over them
+    const properties = schema.properties || {};
+
+    for (const key of Object.keys(inputs)) {
```
batzionb, line 107: I suggest to put this in a separate function getInputsFromVariables that gets the schema and the variables

batzionb:
Please include a screenshot for when there are inputs and when there are no inputs (empty state)

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -97,6 +105,16 @@ export const WorkflowInstancePageContent: React.FC<{
       delete instanceVariables.result;
     }
   }
+  const workflowId = assessedInstance.instance.processId;
+  const instanceId = assessedInstance.instance.id;
+  const { value } = useAsync(async (): Promise<InputSchemaResponseDTO> => {
```
batzionb, line 110: we always need handling of error and loading state

batzionb: In case of error add a ResponseErrorPanel
In case of loading - add a progress bar

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -97,6 +105,16 @@ export const WorkflowInstancePageContent: React.FC<{
...
+    const res = await orchestratorApi.getWorkflowDataInputSchema(
+      workflowId,
+      instanceId,
+    );
+    return res.data;
+  }, [orchestratorApi, workflowId]);
+  const inputs = value?.data;
```
batzionb, line 117: nice way to do it
Indeed if you pass the instance id to the backend, it returns the inputs, I didn't think about this solution

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInputs.tsx
```diff
@@ -0,0 +1,56 @@
...
+  value: any;
+  loading: any;
+  responseError: any;
+}> = ({ className, cardClassName, value, loading, responseError }) => {
+  const inputs = value?.data;
+  return (
+    <InfoCard
+      title="Inputs"
+      subheader={
+        !loading && !inputs && <i>The workflow instance has no inputs</i>
```
batzionb, line 38:
```suggestion
        !loading && !error && !inputs && <i>The workflow instance has no inputs</i>
```

batzionb: I'm merging the PR for now with this issue so we can create the release

batzionb (approved):

PR #235, opened 2025-01-03
Update copyright header

04kash:

This PR updates the copyright header template to accept copyright headers with either "The RHDH Authors" or "The Backstage" (since a lot of the workflows used in this repository are from backstage/community-plugins) as valid options.

The header files for the following workspaces are also updated in the PR: bulk-import, global-header, homepage, lightspeed, marketplace, openshift-image-registry, theme.

I will update the orchestrator workplace files in a separate PR as it has few eslint errors. 

Fixes: https://issues.redhat.com/browse/RHIDP-5401

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage and RHDH Authors
+ * Copyright 2024 The Backstage Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,8 +24,7 @@ module.exports = {
     'notice/notice': [
       'error',
       {
-        mustMatch:
-          /Copyright \d{4} (The Backstage and RHDH Authors|The RHDH Authors)/,
+        mustMatch: /Copyright \d{4} (The Backstage Authors|The RHDH Authors)/,
         templateFile: path.resolve(
           // eslint-disable-next-line no-restricted-syntax
           __dirname,
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright 2024 The Backstage and RHDH Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,8 @@ module.exports = {
     'notice/notice': [
       'error',
       {
+        mustMatch:
+          /Copyright \d{4} (The Backstage and RHDH Authors|The RHDH Authors)/,
         templateFile: path.resolve(
           // eslint-disable-next-line no-restricted-syntax
           __dirname,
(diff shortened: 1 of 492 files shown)


workspaces/bulk-import/examples/template/content/index.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright 2025 The RHDH Authors
```
nickboldt, line 2: should this actuall be a range of dates? eg., 2024-2025?

.eslintrc.cjs
```diff
@@ -24,6 +24,7 @@ module.exports = {
     'notice/notice': [
       'error',
       {
+        mustMatch: /Copyright \d{4} (The Backstage Authors|The RHDH Authors)/,
```
nickboldt, line 27: should this actually be a range of dates? eg., 2024-2025? 

```suggestion
        mustMatch: /Copyright [0-9-]+ (The Backstage Authors|The RHDH Authors)/,
```

christoph-jerolimov:
Hey @04kash, I always wanted start a code-review with this words..

I'm not a lawyar 😄 but I think we should keep the first year we introduced the code, so 2024 in most cases here. It's nice that we update the author from backstage to RHDH for all code and the template to 2025.

workspaces/bulk-import/examples/template/content/index.js
christoph-jerolimov: And when its okay: is RHDH Authors really a thing or should we go one step forward and say something like "Copyright 2024 Red Hat and contributors"

rm3l:
> Hey @04kash, I always wanted start a code-review with this words..
> 
> I'm not a lawyar 😄 but I think we should keep the first year we introduced the code, so 2024 in most cases here. It's nice that we update the author from backstage to RHDH for all code and the template to 2025.

There is an interesting article titled `Copyright Notices for Open Source Projects` from the Legal team on The Source. It does not make it mandatory to include the year.

```
 [...]
We do not recommend inclusion of the year in the copyright notice.
Attempting to consistently maintain dates in copyright notices is difficult, time-consuming, and is unnecessary.
```

So I think we can maybe simplify and omit the year.

04kash:
> There is an interesting article titled Copyright Notices for Open Source Projects from the Legal team on The Source. It does not make it mandatory to include the year.

[...] We do not recommend inclusion of the year in the copyright notice. Attempting to consistently maintain dates in copyright notices is difficult, time-consuming, and is unnecessary. 

So I think we can maybe simplify and omit the year.


Yup I agree omitting the date would make things easier!

workspaces/bulk-import/examples/template/content/index.js
04kash: Changed it to Red Hat, Inc. based on the other copyright notices in the `redhat-developer` org

christoph-jerolimov (approved):
Lgtm!

christoph-jerolimov:
We have currently 20 other PRs, but it looks like most of them are version updates and bugfixes that doesn't introduce new files. There is just one feature PR opened today that needs to be updated anyway.

I guess it's acceptable to merge this change right now and see how many PRs needs to be updated with this new (eslint) requirement.

PR #244, opened 2025-01-07
feat(global-header): implemented searchBar

ciiay:

For [RHIDP-5365](https://issues.redhat.com/browse/RHIDP-5365)

screen recording

https://github.com/user-attachments/assets/0f81b568-7bf2-4da9-8110-630f5c3bcd1c

### DIFF
--- a/workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
+++ b/workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
@@ -1,0 +1,151 @@
+/*
+ * Copyright 2024 The Backstage Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import React from 'react';
+import {
+  SearchResultState,
+  SearchResultProps,
+} from '@backstage/plugin-search-react';
+import Typography from '@mui/material/Typography';
+import { Link } from '@backstage/core-components';
+import ListItem from '@mui/material/ListItem';
+import Box from '@mui/material/Box';
+import Divider from '@mui/material/Divider';
+import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
+import Autocomplete from '@mui/material/Autocomplete';
+import TextField from '@mui/material/TextField';
+import InputAdornment from '@mui/material/InputAdornment';
+import SearchIcon from '@mui/icons-material/Search';
+import { highlightMatch } from '../../utils/stringUtils';
+import styles from './SearchBar.module.css';
+import { useNavigate } from 'react-router-dom';
+
+interface SearchBarProps {
+  query: SearchResultProps['query'];
+  setSearchTerm: (term: string) => void;
+}
+export const SearchBar = (props: SearchBarProps) => {
+  const { query, setSearchTerm } = props;
+  const navigate = useNavigate();
+
+  return (
+    <SearchResultState {...props}>
+      {({ loading, error, value }) => {
+        const results = query?.term ? value?.results ?? [] : [];
+        let options: string[] = [];
+        if (results.length > 0) {
+          options = [
+            ...results.slice(0, 5).map(result => result.document.title),
+            `${query?.term}`,
+          ];
+        } else if (query?.term) {
+          options = ['No results found'];
+        }
+
+        return (
+          <Autocomplete
+            freeSolo
+            options={options}
+            loading={loading}
+            getOptionLabel={option => option ?? ''}
+            onInputChange={(_, inputValue) => setSearchTerm(inputValue)}
+            sx={{ width: '100%' }}
+            filterOptions={x => x}
+            getOptionDisabled={option => option === 'No results found'}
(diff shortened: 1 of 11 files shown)


workspaces/global-header/plugins/global-header/src/utils/stringUtils.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * Highlights the substring that matches the query term.
+ * @param text The full text to render.
+ * @param query The query term to highlight.
+ * @returns JSX.Element with highlighted matching substring.
+ */
+export const highlightMatch = (text: string, query: string) => {
+  if (!query) return <>{text}</>;
+
+  const regex = new RegExp(`(${query})`, 'i');
+  const parts = text.split(regex);
```
christoph-jerolimov, line 30: It looks really nice when it's working!

But this crashes when you enter some regex characters like `[`

Did you pick up that idea from https://github.com/backstage/backstage/blob/master/plugins/search-react/src/components/HighlightedSearchResultText/HighlightedSearchResultText.tsx ?

Can we reuse that component or check why the origin search didn't crash on that character?

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  button: { color: '#fff' },
+                }}
+              />
+            )}
+            renderOption={(renderProps, option, { index }) => {
+              if (option === query?.term && index === options.length - 1) {
+                return (
+                  <Box key="all-results" id="all-results">
+                    <Divider sx={{ my: 0.5 }} />
+                    <Link to={`/search?query=${query?.term}`} underline="none">
```
christoph-jerolimov, line 104: If you don't escape query parameters, you search can't include URL characters like `&`:

```suggestion
                    <Link to={`/search?query=${encodeURIComponent(query?.term)}`} underline="none">
```

Since this happen two times now, maybe you can create a small inline function createSearchLink, or something similar.

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+            getOptionLabel={option => option ?? ''}
+            onInputChange={(_, inputValue) => setSearchTerm(inputValue)}
+            sx={{ width: '100%' }}
+            filterOptions={x => x}
+            getOptionDisabled={option => option === 'No results found'}
+            onKeyDown={event => {
+              if (event.key === 'Enter') {
+                event.preventDefault();
+                if (query?.term) {
+                  navigate(`/search?query=${query?.term}`);
```
christoph-jerolimov, line 72: We should always escape query paramters otherwise searches like `Tom & Jerry` will not work.

```suggestion
                  navigate(`/search?query=${encodeURIComponent(query?.term)}`);
```

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+import ListItem from '@mui/material/ListItem';
+import Box from '@mui/material/Box';
+import Divider from '@mui/material/Divider';
+import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
+import Autocomplete from '@mui/material/Autocomplete';
+import TextField from '@mui/material/TextField';
+import InputAdornment from '@mui/material/InputAdornment';
+import SearchIcon from '@mui/icons-material/Search';
+import { highlightMatch } from '../../utils/stringUtils';
+import styles from './SearchBar.module.css';
```
christoph-jerolimov, line 33: This will work also with dynamic plugins? I'm fine to give it a try. :)

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
```diff
@@ -0,0 +1,151 @@
...
+                  ...params.InputProps,
+                  disableUnderline: true,
+                  startAdornment: (
+                    <InputAdornment position="start">
+                      <SearchIcon style={{ color: '#fff' }} />
+                    </InputAdornment>
+                  ),
+                }}
+                sx={{
+                  pt: '0.5rem',
```
christoph-jerolimov, line 93: For me, it feels like this padding-top breaks that the input field is vertically centered.

```suggestion
```

christoph-jerolimov (changes requested):
It looks really great!

Some small suggestions:

workspaces/global-header/plugins/global-header/src/utils/stringUtils.tsx
ciiay: The original solution doesn't fit our UI design and it uses predefined `preTag` and `postTag` to split text, which won't break the regular expression.
```
text.split(new RegExp(`(${preTag}.+?${postTag})`)); // tags are safe in regExp
```
While my solution uses user input `query` which could include special characters that break a regular expression.

I added some escape to the input text like this for the fix.
```  
const escapeRegex = (input: string) => input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
const escapedQuery = escapeRegex(query);
```

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
ciiay: Yeah, this is a much cleaner solution as long as it works in showcase.. if it doesn't we can add inline styles to override the Mui default styles.

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchBar.tsx
christoph-jerolimov: To be honest, for me, the first one looks the best. :smirk:

christoph-jerolimov (approved):
Great work. :+1:

PR #278, opened 2025-01-14
fix(ci): make config validation step run

jrichter1:
The CI step `validate config` is getting skipped every time. 

`hashFiles` function's path is relative to `GITHUB_WORKSPACE`, which point to the root of the repo. This should change the path to the respective workspace, so it actually returns something when `app-config.yaml` exists there.


.github/workflows/ci.yml
```diff
@@ -84,7 +84,7 @@ jobs:
         run: yarn fix --check
 
       - name: validate config
-        if: ${{ hashFiles('app-config.yaml') != '' }}
+        if: ${{ hashFiles(format('workspaces/{0}/app-config.yaml', matrix.workspace)) != '' }}
```
nickboldt, line 87: surprised this is needed since we already set
```
        working-directory: ./workspaces/${{ matrix.workspace }}
```

above on line 58, but this looks like a reasonable solution. did not try to run it to verify the format() syntax.

04kash: Yeah, the [github documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/setting-a-default-shell-and-working-directory#setting-default-shell-and-working-directory) mentions the same thing.

I'll merge this and test it on a few PRs.

04kash (approved):

PR #300, opened 2025-01-19
fix(orchestrator): Add retrigger option to UI using v2 API

LiorSoffer:
resolve [FLPATH-1684](https://issues.redhat.com/browse/FLPATH-1684)

https://github.com/user-attachments/assets/56184ffe-5ae2-4767-853d-edab849d4d7b

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import { LoggerService } from '@backstage/backend-plugin-api';
 
 import {
@@ -148,6 +149,7 @@ export class SonataFlowService {
 
     if (!response.ok) {
       const json = await response.json();
+      json.details = `failedNodeId:${json?.failedNodeId}`;
       this.logger.error(`Retrigger failed with: ${JSON.stringify(json)}`);
       throw new Error(
         `${await this.createPrefixFetchErrorMessage(
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import React, { useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
@@ -33,6 +34,8 @@ import {
   CircularProgress,
   Grid,
   IconButton,
+  Menu,
+  MenuItem,
   Tooltip,
 } from '@material-ui/core';
 import Snackbar from '@material-ui/core/Snackbar';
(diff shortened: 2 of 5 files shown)


workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -148,6 +149,7 @@ export class SonataFlowService {
 
     if (!response.ok) {
       const json = await response.json();
+      json.details = `failedNodeId:${json?.failedNodeId}`;
```
mareklibra, line 152: Please correct me, is that augmentation for the generic error handler to include that in the user-facing message?

If so, we can use it but please make sure we do not loose the `details` which might be potentially sent by the server.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -222,6 +236,47 @@ export const WorkflowInstancePage = ({
...
+    }
+  };
+
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+
+  const handleClick = (event: React.MouseEvent<HTMLElement>) => {
+    setAnchorEl(event.currentTarget);
+  };
+
+  const handleCloseMenue = () => {
```
mareklibra, line 270:
```suggestion
  const handleCloseMenu = () => {
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -276,18 +331,55 @@ export const WorkflowInstancePage = ({
...
+                    style={{ color: 'white' }}
                   >
                     Rerun
                   </Button>
                 </Tooltip>
+                <Menu
+                  anchorEl={anchorEl}
+                  open={Boolean(anchorEl)}
+                  onClose={handleCloseMenue}
+                  getContentAnchorEl={null}
```
mareklibra, line 358: Is that property needed?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -222,6 +236,47 @@ export const WorkflowInstancePage = ({
...
+            `Couldn't retrigger instance ${value.instance.id} of workflow ${value.instance.processName}`,
+          );
+        setIsRerunSnackbarOpen(true);
+      } finally {
+        setIsRetrigger(false);
+      }
+    }
+  };
+
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
```
mareklibra, line 264: Storing HTML elements in React state is gross. Can we use `useRef` instead?

mareklibra (changes requested):

workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
LiorSoffer: yes it is the purpose. do you think is needed to provide the user the failing node?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
LiorSoffer: which one?

LiorSoffer: `getContentAnchorEl` determines how the menu is positioned relative to the button

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
LiorSoffer: I agree

workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
mareklibra: Maybe the end-user does not care much but it can be useful when developing/stabilizing the workflow. So let's keep it.

My concern is about loosing potential content of the `json.details`, we are just blindly rewriting it here.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
mareklibra: Thanks, I see the difference now

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -222,6 +236,43 @@ export const WorkflowInstancePage = ({
...
+      setIsRetrigger(true);
+      try {
+        await orchestratorApi.retriggerInstance(
+          value.instance.processId,
+          value.instance.id,
+        );
+        restart();
+      } catch (retriggerInstanceError) {
+        if (retriggerInstanceError.toString().includes('Failed Node Id')) {
+          setRetriggerError(`Run failed again`);
```
mareklibra, line 250: Nit:
```suggestion
          setRetriggerError('Run failed again');
```

(the template strings are not needed here)

mareklibra (approved):

PR #302, opened 2025-01-20
chore(orchestrator): add sonataflow env variables

mareklibra:
Add new `orchestrator.sonataFlowService.notificationsUrl` and `orchestrator.sonataFlowService.notificationsBearerToken` app-config parameters.

In a follow-up, we will extend the `app-config.yaml` configuration to pass down a set of env variables based on wildcard name.


gciavarrini (approved):
lgtm

gciavarrini:
There is 1 New Issue because of the TODO Comment (for the env var) so i think we can merge it anyway since in the next PRs you'll address it.

PR #310, opened 2025-01-22
feat(marketplace): add new aggregation api endpoint in marketplace-backend

karthikjeeyar:
## Fixes
https://issues.redhat.com/browse/RHIDP-5566 

A new API to get aggregated data from catalog. 

Screenshot 2025-01-22 at 6 57 02 PM

## How to test

1.  run `yarn dev` from workspaces/marketplace folder.
2.  use the below curl command to hit the aggregation api endpoint.

**POST** `/aggregations`
```
curl --location 'http://localhost:7007/api/marketplace/aggregations' \
--header 'Authorization: Bearer <your_token>' \
--header 'Content-Type: application/json' \
--data '[{
    "field": "kind",
    "type": "count",
    "filter": { "spec.owner" : "redhat"}
}]'

```

## Unit tests:

marketplace-common tests:
```
EntityFilterQuery schema
  ✓ should validate an object with string keys and string values (7 ms)
  ✓ should reject an object with non-string keys or values (8 ms)
  ✓ should reject non-object types (2 ms)
  ✓ should allow an empty object (1 ms)
AggregationsSchema
  ✓ should not allow an empty array (1 ms)
  ✓ should not throw error for valid data (1 ms)
  ✓ should not allow invalid type (1 ms)

getAggregateData
  ✓ should throw error if the database client is not passed (3 ms)
  ✓ should generate query for count aggregation (2 ms)
  ✓ should handle aggregations with filters and having clauses (2 ms)
  ✓ should handle having clauses with count (1 ms)
  ✓ should use a custom base query if provided (1 ms)
  ✓ should handle orderFields (1 ms)
```

marketplace-backend tests:

```
   ✓ should throw an error when the aggregations request is invalid (39 ms)
    ✓ should return sanitized error messages (11 ms)
    ✓ should return aggregated data (10 ms)
```    


workspaces/marketplace/plugins/marketplace-common/package.json
```diff
@@ -17,7 +17,7 @@
     "directory": "workspaces/marketplace/plugins/marketplace-common"
   },
   "backstage": {
-    "role": "node-library",
+    "role": "common-library",
```
christoph-jerolimov, line 20: :+1:

workspaces/marketplace/plugins/marketplace-common/package.json
```diff
@@ -50,6 +50,8 @@
   },
   "dependencies": {
     "@backstage/backend-plugin-api": "^1.0.1",
-    "@backstage/errors": "^1.2.4"
+    "@backstage/errors": "^1.2.4",
+    "knex": "^3.0.0",
```
christoph-jerolimov, line 54: As discussed on slack: I think we shouldn't have knex as a dependency in our common package. Either we move that knex code directly into the marketplace-backend or we can create a node module for that code.

christoph-jerolimov (changes requested):
As discussed :)

workspaces/marketplace/plugins/marketplace-common/package.json
karthikjeeyar: I have moved the database code to marketplace-backend, but knex dependancy is needed here just for the aggregationApi types `Knex.QueryBuilder`

workspaces/marketplace/plugins/marketplace-common/src/types.ts
```diff
@@ -93,3 +95,36 @@ export interface MarketplaceApi {
...
   getPluginsByPluginListName(name: string): Promise<MarketplacePlugin[]>;
 }
+
+/**
+ * @public
+ */
+export interface MarketplaceAggregationApi {
+  fetchAggregatedData(
+    aggregationsRequest: AggregationsRequest,
+    baseQuery?: Knex.QueryBuilder | null,
```
christoph-jerolimov, line 105: I didn't tested it yet, but the code and tests looks really nice. :+1:

This line (and the common dependency to knex) is the only aspect that I like to understand. I guess you would like to reuse this for tables other than final_entities. But I think the structure of final_entities is really "special", so it would surprise me if the rest would work with "another table". :thinking: 

Let us have a quick call later today.

workspaces/marketplace/plugins/marketplace-backend/src/service/MarketplaceAggregationService.ts
```diff
@@ -0,0 +1,184 @@
...
+      if (agg.type === 'count') {
+        query.groupBy(`${currentAlias}.key`, `${currentAlias}.value`);
+      }
+
+      // Apply HAVING filters
+      if (agg.havingFilter) {
+        const { field, operator, value } = agg.havingFilter;
+        const rawField = field === 'count' ? dbClient.raw('COUNT(*)') : field;
+        query.having(rawField, operator, value);
+      }
```
christoph-jerolimov, line 164: Do you think this should be an array and a foreach similar to the filter above?

karthikjeeyar: Yes, we can introduce support for multiple having clauses. Also I am thinking to add a new prop for logical operator (AND or OR) to both filters and having filters so it becomes more flexible.

karthikjeeyar: Yes, we can introduce support for multiple having clauses.

workspaces/marketplace/plugins/marketplace-common/src/types.ts
karthikjeeyar: You are right!. It wouldn't work with another table. Initially I designed this function as recursive one so I was trying to build the query based on this optional param but later I changed it to iterative approach for simplicity and forgot to remove this code. Will remove this code and dependancy in both the places.

PR #312, opened 2025-01-22
fix(orchestrator):add view variables dialog

LiorSoffer:
resolve [FLPATH-2001](https://issues.redhat.com/browse/FLPATH-2001)

Screenshot from 2025-01-22 17-51-52

long list case:
Screenshot from 2025-01-22 17-51-16

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
@@ -33,6 +33,7 @@ import {
   AssessedProcessInstanceDTO,
   InputSchemaResponseDTO,
   ProcessInstanceDTO,
+  WorkflowDataDTO,
 } from '@red-hat-developer-hub/backstage-plugin-orchestrator-common';
 
 import { orchestratorApiRef } from '../../src/api/api';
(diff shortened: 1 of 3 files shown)


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -89,7 +97,11 @@ export const WorkflowInstancePageContent: React.FC<{
   );
 
   const workflowdata = assessedInstance.instance?.workflowdata;
-  let instanceVariables;
+  let instanceVariables: {
```
mareklibra, line 100:
```suggestion
  let instanceVariables: WorkflowDataDTO;
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePageContent.tsx
```diff
@@ -113,14 +125,62 @@ export const WorkflowInstancePageContent: React.FC<{
     return res.data;
   }, [orchestratorApi, workflowId]);
 
+  const [isVariablesDialogOpen, setIsVariablesDialogOpen] = useState(false);
+
+  const toggleVariablesDialog = React.useCallback(() => {
+    setIsVariablesDialogOpen(prev => !prev);
+  }, []);
+
+  const VariablesDialogContent = () => (
```
mareklibra, line 134: Nit: by moving this component out of this function and passing `instanceVariables` as its property, the code will be cleaner and the component will not be re-created with every run.

Anyway, not a big issue in that particular case.

mareklibra (changes requested):

mareklibra (approved):

mareklibra (approved):

PR #316, opened 2025-01-24
chore(orchestrator): add orchestror among pluginsWithPermission

mareklibra:
The only purpose is to simplify dev-env setup when working with RBAC

### DIFF
--- a/workspaces/orchestrator/app-config.yaml
+++ b/workspaces/orchestrator/app-config.yaml
@@ -81,8 +81,10 @@ auth:
 # permission:
 #   enabled: true
 #   rbac:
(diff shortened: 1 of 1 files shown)


workspaces/orchestrator/app-config.yaml
```diff
@@ -81,8 +81,10 @@ auth:
 # permission:
```
mareklibra, line 81: Dear reviewer, this is intentionally commented-out by default :-)
We usually do not need this feature for day-to-day development.

gciavarrini (approved):
lgtm

PR #318, opened 2025-01-24
feat(fab): add application listener and provider test plugins

christoph-jerolimov:

Resolves https://issues.redhat.com/browse/RHIDP-5691

This adds two new test plugins to test the application/listener and application/provider.

We might want to move them later into another workspace. I'll add this topic to the architecture call. That's the reason I called them `application-listener-test` and not fab-test. We might want to have that as well. But FAB components are mostly configuration so I'm not sure if that is needed.

These both plugins should be tested after they are released on npm together with https://github.com/redhat-developer/rhdh/pull/2199


debsmita1 (approved):
/lgtm

PR #329, opened 2025-01-25
fix(orchestrator): list orchestrator.workflow.use in the RBAC UI

mareklibra:
With this change, the RBAC UI can list both `orchestrator.workflow` and `orchestrator.workflow.use` permissions.

Fixes: FLPATH-2032


mareklibra:
flpath-2032 workflowUsePermission

PR #330, opened 2025-01-25
fix(orchestrator): use 'update' policy for orchestrator.workflow.use

mareklibra:
Changing the `read` policy to `update` of the `orchestrator.workflow.use` permission to closer align with other plugins.
Documentation updated.

Fixes: FLPATH-2033


mareklibra:
flpath-2033 workflowUsePolicyAsUpdate

gciavarrini (approved):
lgtm

PR #332, opened 2025-01-27
feat(marketplace): add optional sorting, filtering, pagination to marketplace api

its-mitesh-kumar:


workspaces/marketplace/plugins/marketplace-common/src/types.ts
```diff
@@ -88,7 +108,13 @@ export interface MarketplacePluginSpec extends JsonObject {
...
  */
 export interface MarketplaceApi {
-  getPlugins(): Promise<MarketplacePlugin[]>;
+  getPlugins(params: {
+    limit?: string;
+    cursor?: string;
+    sortByField?: string;
+    sortOrder?: 'asc' | 'desc';
+    searchText?: string;
+  }): Promise<MarketplacePluginWithPageInfo>;
```
christoph-jerolimov, line 117: Please rename `params` to query and export the Params type as something like `GetPluginsRequest`.

Can you also reuse the types `EntityFilterQuery` and `EntityOrderQuery` for the filter and query. That allow us to just pass these values to the catalog client.

But this will require that you parse more complex types in the router. See comment there.

Can you please also use this type from https://backstage.io/docs/reference/catalog-client.queryentitiesinitialrequest/ for the full text search:

```tsx
    fullTextFilter?: {
        term: string;
        fields?: string[];
    };
```

The `request` param should be optional.

workspaces/marketplace/plugins/marketplace-backend/src/router.ts
```diff
@@ -38,7 +39,16 @@ export async function createRouter({
...
+    const { cursor, limit, sortByField, sortOrder, searchText } = _req.query;
+    const plugins = await marketplaceApi.getPlugins({
+      cursor: typeof cursor === 'string' ? cursor : undefined,
+      limit: typeof limit === 'string' ? limit : undefined,
+      sortByField: typeof sortByField === 'string' ? sortByField : undefined,
+      sortOrder:
+        sortOrder === 'asc' || sortOrder === 'desc' ? sortOrder : SortOrder.asc,
+      searchText: typeof searchText === 'string' ? searchText : undefined,
+    });
+
```
christoph-jerolimov, line 51: As we discussed, can you please support multiple filters and orders here, like:

* GET ... `?filters[kind]=component&filters[user]=all` for filters
* GET ... `?order=name,ASC&order=title,ASC` for the order
* GET ... `?searchText=...` 

And it would be great to have an offset as well.

workspaces/marketplace/plugins/marketplace/src/api/MarketplaceBackendClient.tsx
```diff
@@ -36,7 +37,7 @@ export class MarketplaceBackendClient implements MarketplaceApi {
     this.fetchApi = options.fetchApi;
   }
 
-  async getPlugins(): Promise<MarketplacePlugin[]> {
+  async getPlugins(): Promise<MarketplacePluginWithPageInfo> {
```
christoph-jerolimov, line 40: This should use the request parameters and apply all the parameters (filter, order, ...) to the URL below.

Please use `URL` or `URLSearchParams` to escape the values.

christoph-jerolimov (changes requested):
Thanks! As discussed some improvement ideas :)

its-mitesh-kumar:
@karthikjeeyar @christoph-jerolimov I have implemented like 
We  can **encode** complex query parameter like **fullTextFilter** , **filter** and **orderFields** in frontend  and decode it in the backend. Encoding in frontend can be like this . Let me know with this approach we may have any problem .
```
let search = {
  term: "Backstage",
  fields: ["metadata.description"], // mutiple fields we can add 
}
let orderFields = {
  field: "metadata.name",
  order: "asc",
} // can be array of object also in case of multiple fields
let filter = {
  "spec.type": "backend-plugin",
  "metadata.name": "search",
}
filter= encodeURIComponent(JSON.stringify(filter))
search=encodeURIComponent(JSON.stringify(search))
orderFields=encodeURIComponent(JSON.stringify(orderFields))
const endpoint=`${baseUrl}?fullTextFilter=${search}&filter=${filter}&orderFields=${orderFields}&limit=10&offset=10`
```

workspaces/marketplace/plugins/marketplace-common/src/types.ts
```diff
@@ -96,6 +118,39 @@ export interface MarketplacePluginSpec extends JsonObject {
...
+  fields?: string[];
+};
+
+/**
+ * @public
+ */
+export type GetPluginsRequest = {
+  limit?: number;
+  offset?: number;
+  filter?: Record<string, string>;
```
karthikjeeyar, line 136: we should keep this aligned with catalog types
```suggestion
  filter?: EntityFilterQuery;
```

workspaces/marketplace/plugins/marketplace-common/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  const filter: EntityFilterQuery = query?.filter
+    ? {
+        ...(typeof query.filter === 'string'
+          ? JSON.parse(query.filter)
+          : query.filter),
+        ...RequiredFilter,
+      }
+    : RequiredFilter;
```
karthikjeeyar, line 58: You could reuse `encodeFilterParams` from `/utils/encodeQueryParams.ts` here.

workspaces/marketplace/plugins/marketplace-common/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+      const { field, order } = query.orderFields;
+      params.append('orderFields', `${field},${order}`);
+    }
+  }
+  if (query?.filter) params.append('filter', JSON.stringify(query.filter));
+  return params;
+};
+
+export const convertSearchParamsToGetPluginsRequest = (
```
karthikjeeyar, line 92: I think it will be good to extend the generic `encodeQueryParams` function to handle more fields, Instead of having it's own version for GetPluginsRequest. This way we can reuse the function for other API like PluginLists, Packages etc.

workspaces/marketplace/plugins/marketplace-common/src/api/MarketplaceCatalogClient.test.ts
```diff
@@ -93,13 +93,13 @@ describe('MarketplaceCatalogClient', () => {
 
     it('should return the plugins', async () => {
       const api = new MarketplaceCatalogClient(options);
-      const plugins = await api.getPlugins();
+      const plugins = await api.getPlugins({});
```
karthikjeeyar, line 96: argument is optional, isnt it? 🤔 
```suggestion
      const plugins = await api.getPlugins();
```

workspaces/marketplace/plugins/marketplace-common/src/api/MarketplaceCatalogClient.test.ts
```diff
@@ -93,13 +93,13 @@ describe('MarketplaceCatalogClient', () => {
...
       expect(plugins).toHaveLength(2);
     });
 
     it('should return the plugins when the auth options is not passed', async () => {
       const api = new MarketplaceCatalogClient({ ...options, auth: undefined });
-      const plugins = await api.getPlugins();
+      const plugins = await api.getPlugins({});
```
karthikjeeyar, line 102:
```suggestion
      const plugins = await api.getPlugins();
```

karthikjeeyar (dismissed):

workspaces/marketplace/plugins/marketplace-common/src/types.ts
```diff
@@ -28,6 +30,18 @@ export interface MarketplacePlugin extends Entity {
   spec?: MarketplacePluginSpec;
 }
 
+/**
+ * @public
+ */
+export interface MarketplacePluginWithPageInfo {
+  items: MarketplacePlugin[];
+  totalItems?: Number;
```
christoph-jerolimov, line 38: The type is `number` while the object is `Number`

```suggestion
  totalItems?: number;
```

workspaces/marketplace/plugins/marketplace-backend/src/router.ts
```diff
@@ -36,9 +37,16 @@ export async function createRouter({
...
+  router.get('/plugins', async (req, res) => {
+    try {
+      const query = req.query as Partial<GetPluginsRequest>;
+      const plugins = await marketplaceApi.getPlugins(query);
+      res.json(plugins);
+    } catch (error) {
+      res
+        .status(500)
+        .json({ message: 'Failed to fetch plugins', error: error.message });
+    }
```
christoph-jerolimov, line 49: I don't think that we need a custom error handler here. Backstage routes has a default handle if I'm correct

workspaces/marketplace/plugins/marketplace-common/src/types.ts
```diff
@@ -96,6 +118,39 @@ export interface MarketplacePluginSpec extends JsonObject {
...
+ */
+export interface MarketplaceApi {
+  getPlugins(
+    request?: GetPluginsRequest,
+  ): Promise<MarketplacePluginWithPageInfo>;
+  getPluginByName(name: string): Promise<MarketplacePlugin>;
+  getPluginLists(): Promise<MarketplacePluginList[]>;
+  getPluginListByName(name: string): Promise<MarketplacePluginList>;
+  getPluginsByPluginListName(name: string): Promise<MarketplacePlugin[]>;
+}
```
christoph-jerolimov, line 152: Where was the old MarketplaceApi interface defined???

workspaces/marketplace/plugins/marketplace-common/src/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+  EntityOrderQuery,
+  QueryEntitiesRequest,
+} from '@backstage/catalog-client/index';
+import { GetPluginsRequest, SortOrder } from './types';
+
+const DefaultOrderFields: EntityOrderQuery = [
+  { field: 'metadata.name', order: 'asc' },
+];
+const DefaultLimit = 20;
+const RequiredFilter = { kind: 'plugin' };
```
christoph-jerolimov, line 27: Don't know if this is needed. But please start consts with lower case characters.

workspaces/marketplace/plugins/marketplace/src/api/MarketplaceBackendClient.tsx
```diff
@@ -39,9 +42,12 @@ export class MarketplaceBackendClient implements MarketplaceApi {
...
   }
 
-  async getPlugins(): Promise<MarketplacePlugin[]> {
+  async getPlugins(
+    request?: GetPluginsRequest,
+  ): Promise<MarketplacePluginWithPageInfo> {
     const baseUrl = await this.discoveryApi.getBaseUrl('marketplace');
-    const url = `${baseUrl}/plugins`;
+    const params = convertGetPluginRequestToSearchParams(request);
+    const url = `${baseUrl}/plugins?${params.toString()}`;
```
christoph-jerolimov, line 50: Let us do the extra step and don't append the `?`  if no params is needed.

```suggestion
    const params = convertGetPluginRequestToSearchParams(request);
    const query = params.toString();
    const url = `${baseUrl}/plugins${query ? '?' : ''}${query}`;
```

christoph-jerolimov (changes requested):
Small type questions for now. We should also align that with the other utils. But its a good start so far. Please add unit tests for all utils.

workspaces/marketplace/.changeset/quiet-kiwis-admire.md
```diff
@@ -0,0 +1,7 @@
+---
+'@red-hat-developer-hub/backstage-plugin-marketplace-backend': patch
+'@red-hat-developer-hub/backstage-plugin-marketplace-common': patch
+'@red-hat-developer-hub/backstage-plugin-marketplace': patch
+---
+
+add sorting , filtering , pagination in /plugins api
```
christoph-jerolimov, line 7:
```suggestion
add optional sorting, filtering, pagination to marketplace api
```

workspaces/marketplace/plugins/marketplace-common/src/types.ts
karthikjeeyar: Its down in the end of this file :) Probably this is repeated after merge conflict.

workspaces/marketplace/plugins/marketplace-backend/src/router.ts
its-mitesh-kumar: Okay , I am removing it .

workspaces/marketplace/plugins/marketplace-common/src/utils.ts
its-mitesh-kumar: Have changed into lower case . Since we want to get plugins by default we need to pass { kind: 'plugin' }  , in `queryEntities` . I wanted to get only 20 results by default .

workspaces/marketplace/plugins/marketplace-backend/src/router.ts
its-mitesh-kumar: Removed .

workspaces/marketplace/plugins/marketplace-common/src/types.ts
its-mitesh-kumar: Done

workspaces/marketplace/plugins/marketplace-common/src/utils.ts
christoph-jerolimov: Without a default limit it will return all entities? Let us not have a limit for now since the UI doesn't support pagination at the moment.

christoph-jerolimov (approved):

PR #372, opened 2025-01-31
fix(global-header): enable theme matching

ciiay:

For [RHIDP-5364](https://issues.redhat.com/browse/RHIDP-5364)

Screen recording(updated on Feb 4th)
Testing with [RHDH PR #2263](https://github.com/redhat-developer/rhdh/pull/2263)

https://github.com/user-attachments/assets/95da30d9-86f4-4ec9-bbc3-32d93e047f12

### DIFF
--- a/workspaces/global-header/plugins/global-header/src/components/GlobalHeader.tsx
+++ b/workspaces/global-header/plugins/global-header/src/components/GlobalHeader.tsx
@@ -23,11 +23,13 @@ import { useGlobalHeaderMountPoints } from '../hooks/useGlobalHeaderMountPoints'
 import { ComponentType, GlobalHeaderComponentMountPoint, Slot } from '../types';
 import { ErrorBoundary } from '@backstage/core-components';
 import { configApiRef, useApi } from '@backstage/core-plugin-api';
+import { useTheme } from '@mui/material/styles';
 
 export const GlobalHeader = () => {
   const config = useApi(configApiRef);
   const frontendConfig = config.getOptionalConfig('dynamicPlugins.frontend');
   const frontendData = frontendConfig?.get();
+  const theme = useTheme();
 
   const allGlobalHeaderMountPoints = useGlobalHeaderMountPoints();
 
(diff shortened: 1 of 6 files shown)


ShiranHi:
Thank you @ciiay, it looks great, it's exciting to see the new header 🥳

Two thoughts: 

1. Can we ensure that the "All results" button in the search dropdown and the "All templates" button in the "Create" button dropdown use the correct shade of blue in dark mode?  

2. When switching between dark and light modes, I noticed the "Create" button disappears while displaying a "Loading templates..." message. Would it be possible to keep the "Create" button visible but disabled during this loading phase?
image

ciiay:
> Thank you @ciiay, it looks great, it's exciting to see the new header 🥳
> 
> Two thoughts:
> 
> 1. Can we ensure that the "All results" button in the search dropdown and the "All templates" button in the "Create" button dropdown use the correct shade of blue in dark mode?
> 2. When switching between dark and light modes, I noticed the "Create" button disappears while displaying a "Loading templates..." message. Would it be possible to keep the "Create" button visible but disabled during this loading phase?
> 
> image

Hi @ShiranHi , thanks for the review. Here's an updated outlined button as we discussed on Slack. I made the button border opacity to 0.5, so when hover on the button it looks brighter. This is just for preview. Let me know if I need to change it to another style.

https://github.com/user-attachments/assets/5c0ffb50-8236-4cd0-872a-a7b997badde6

And here's how the button looks when loading:
image
Comparing with the regular button:
image

ShiranHi:
Thank you @ciiay for the changes! 
For context, we've changed the button from a primary to a secondary style because some pages, like the Catalog, already have a primary button, and we want to avoid having two primary buttons on the same page (based on the [PatternFly guidelines](https://www.patternfly.org/components/button/design-guidelines/#primary-button)).  

Two questions: 
1. > And here's how the button looks when loading:

Can we disable it instead of showing a loading state? It loads quickly, so I don’t think a spinner is necessary...

2. The list of templates is quite long. Can we limit the dropdown height to a maximum of half the screen height?

ciiay:
Hi @ShiranHi , I uploaded the screen recording of the updates in the PR description. Let me know if it looks good to you, thank you 🌷

ShiranHi:
Looks great @ciiay , thank you!

christoph-jerolimov (approved):
Thanks for the UI improvement :+1:

To be honest, I'm not a fan of all this custom styles.

I expect that this will raise issues for customers with a customized (configuration) theme or a complete custom theme (own dynamic plugin).

But I like to get it in anyway since it improves the styles. But I also hope that we can improve this step by step later.

PR #373, opened 2025-01-31
feat(global-header): implemented support button in header

ciiay:

For [RHIDP-5252](https://issues.redhat.com/browse/RHIDP-5252)

- support button implementation
- notifications button implementation
- removed hardcoded colors

screen recording(updated on Feb 6th):
Tested with [RHDH PR #2263](https://github.com/redhat-developer/rhdh/pull/2263), and [theme update PR](https://github.com/redhat-developer/red-hat-developer-hub-theme/pull/53)

https://github.com/user-attachments/assets/c2a1eaa5-28b1-48eb-857e-14d7c8ba6123


workspaces/global-header/plugins/global-header/dev/index.tsx
```diff
@@ -71,7 +71,7 @@ const defaultGlobalHeaderComponentsMountPoints: GlobalHeaderComponentMountPoint[
         props: {
           icon: 'support',
           tooltip: 'Support',
-          to: '/support',
+          to: 'https://github.com/redhat-developer/rhdh/issues',
```
christoph-jerolimov, line 74: When you open that page and click on create you will notice that we redirect all our users to Jira instead.

So it should be:

```suggestion
          to: 'https://issues.redhat.com/browse/RHIDP',
```

ciiay: Good catch on my dev code. Thank you 🤝 

Hmm, `https://issues.redhat.com/browse/RHIDP` is an internal Jira page, do the users have the access to it?
`https://github.com/redhat-developer/rhdh/issues` is the one anyone can visit right?

ShiranHi:
@ciiay it looks great!
Question: Can we update the tooltip text to "Support (external site)" when the link directs to an external page?

workspaces/global-header/plugins/global-header/dev/index.tsx
christoph-jerolimov: But GH issues will be empty and users shouldn't ask questions there.

Since it's just a dev page you can maybe simple use https://developers.redhat.com/rhdh/overview ?

workspaces/global-header/.changeset/grumpy-suns-impress.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-header': major
```
christoph-jerolimov, line 2: There is still things missing. Can we go with patch or minor until we're ready?

ciiay: Will change it to minor. I thought we needed a first version out then we can improve it on top of it as patches.

workspaces/global-header/plugins/global-header/dev/index.tsx
logonoff: This link goes directly to the create issue form if that is preferrable: https://issues.redhat.com/secure/CreateIssue.jspa?pid=12335748&issueType=13263

its-mitesh-kumar:
@ciiay , Not able to see unit tests , will we be targeting it as part of another story ?

ciiay:
> @ciiay , Not able to see unit tests , will we be targeting it as part of another story ?

Will work on a separate PR. This PR needs to get in and release the first version for RHDH cluster testing asap.

workspaces/global-header/README.md
```diff
@@ -14,3 +14,23 @@ To generate knip reports for this app, run:
...
+      - type: static
+        options:
+          token: ${EXTERNAL_SERVICE_TOKEN}
+          subject: cicd-system-completion-events
+```
+
+To send notifications, you can use the following command:
+
+```sh
+curl -X POST http://localhost:7007/api/notifications -H "Content-Type: application/json" -H "Authorization: Bearer ${EXTERNAL_SERVICE_TOKEN}" -d '{"recipients":{"type":"broadcast"},"payload": {"title": "Title of boradcast message","link": "http://foo.com/bar","severity": "high","topic": "The topic"}}'
```
christoph-jerolimov, line 33: Thanks for this! :)

I read it first in the official doc and they said the bearer token should be base64 encoded. Looks like that isn't true.

Btw, I couldn't start the workspace without adding this.

You can maybe just change the default to `token: EXTERNAL_SERVICE_TOKEN`.

For this service token I think we can live with an "hard coded" token. Wdyt?

workspaces/global-header/app-config.yaml
```diff
@@ -32,6 +32,15 @@ backend:
     connection: ':memory:'
   # workingDirectory: /tmp # Use this to configure a working directory for the scaffolder, defaults to the OS temp-dir
 
+  # enable notifications backend
+  auth:
+    dangerouslyDisableDefaultAuthPolicy: true
+    externalAccess:
+      - type: static
+        options:
+          token: ${EXTERNAL_SERVICE_TOKEN}
```
christoph-jerolimov, line 41: I changed it locally to this and it just worked fine. Wdyt?

```suggestion
          token: EXTERNAL_SERVICE_TOKEN
```

workspaces/global-header/app-config.yaml
```diff
@@ -63,14 +72,36 @@ techdocs:
...
+          resolvers:
+            - resolver: usernameMatchingUserEntityName
 
 scaffolder:
   # see https://backstage.io/docs/features/software-templates/configuration for software template options
 
+# notifications plugin config
+notifications:
+  processors:
+    email:
```
christoph-jerolimov, line 94: Couldn't we live without email processor? But it's fine. I hope this doesnt' connect always to gmail and tries to send emails???

workspaces/global-header/plugins/global-header/src/components/GlobalHeader.tsx
```diff
@@ -14,138 +14,19 @@
...
-              flexItem
-              sx={{ borderColor: '#4F5255', marginX: 1 }}
-            />
-          )}
-        {renderComponents(globalHeaderEndComponentsMountPoints)}
-      </Toolbar>
-    </AppBar>
+    <GlobalHeaderComponent
+      globalHeaderMountPoints={allGlobalHeaderMountPoints ?? []}
+      supportUrl={supportUrl ?? ''}
```
christoph-jerolimov, line 29: Nit: Why not make the prop optional and pass undefined when its not configured. I personally think that defaulting to empty strings brings up a (small) risk.

workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/ProfileDropdown.tsx
```diff
@@ -46,6 +47,16 @@ export const ProfileDropdown = ({
   const identityApi = useApi(identityApiRef);
   const [user, setUser] = useState<ProfileInfo>();
   const profileDropdownMountPoints = useProfileDropdownMountPoints();
+  const [bgColor, setBgColor] = useState('#3C3F42');
+
+  useEffect(() => {
+    const container = document.getElementById('global-header');
```
christoph-jerolimov, line 53: I like these "pick up the right color" details. But getElementById doesn't feel like the react way to pick up an element, or?

Maybe we you can create a bug that we get rid of this and use a ref prop instead? But it's fine for now.

workspaces/global-header/plugins/global-header/src/components/HeaderIconButtonComponent/HeaderIconButton.tsx
```diff
@@ -35,15 +38,37 @@ export const HeaderIconButton = ({
...
+    buttonProps.target = '_blank';
+    buttonProps.rel = 'noopener noreferrer';
+  }
+
   return (
     <Tooltip title={tooltip ?? icon}>
       <IconButton
         color="inherit"
-        aria-label="help"
+        aria-label={icon}
```
christoph-jerolimov, line 61: I'm worried that the icon string isn't valuable for the user as an "alternative".

It's a good point. Maybe each icon should have a `label`  or `ariaLabel` as well that is used here?

For now I guess the tooltip is more useful then the icon. Wdyt?

workspaces/global-header/plugins/global-header/src/components/GlobalHeaderComponent.tsx
```diff
@@ -0,0 +1,148 @@
...
+              <mp.Component
+                key={`header-component-${index.toString()}`}
+                {...getDropdownButtonProps(mp.config?.key ?? index.toString())}
+                {...mp.config?.props}
+              />
+            </ErrorBoundary>
+          );
+        case ComponentType.ICON_BUTTON:
+          return (
+            <ErrorBoundary>
```
christoph-jerolimov, line 118: I saw some react warnings:

> Warning: Each child in a list should have a unique "key" prop.
> Check the render method of `GlobalHeaderComponent`. See https://reactjs.org/link/warning-keys for more information.

I guess these warnings comes from here where the error boundaries needs a key.

christoph-jerolimov (changes requested):
Thanks @ciiay !

This looks really awesome. Thanks for all the effort and thanks for adding the notification and signals plugin to the workspace. This makes it so much easier to test the plugin!

Awesome work.

I have one question anyway: Why is the default background color blue? Is it because that's the AppBar default?

Then I checked on a cluster without Software templates. I checked with @ShiranHi and she recommended to render in that case "No templates available" and hide the "All template" link. Do you want include this into this PR or should I create a bug for that?

image

And then I found some smaller issues or recommendations. Please take a look below:

workspaces/global-header/README.md
ciiay: Hi @christoph-jerolimov , the `${EXTERNAL_SERVICE_TOKEN}` is an expression for environment variables. All env variables in our existing yaml files are in this format. Better we don't use a new expression only for this file, it could be confusing.

Instead of a hardcoded value, I can attach the link to the Backstage document where explains how to generate an `EXTERNAL_SERVICE_TOKEN` value. The reason I want to keep this name is that it shows exactly what it is. Otherwise a string could be confusing, like what is that and where is that from. What do you think?

workspaces/global-header/app-config.yaml
ciiay: It will cause this error when backend starts

```
backend:start: ForwardedError: Module 'email' for plugin 'notifications' startup failed; caused by Error: Missing required config value at 'notifications.processors.email' in 'app-config.local.yaml'
```

ciiay:
Hi @ShiranHi @christoph-jerolimov , how do you like this when no templates available?
image
image

ShiranHi:
> how do you like this when no templates available?

Can the text "Use a template" be changed to "No templates available" and the second line removed?

workspaces/global-header/app-config.yaml
christoph-jerolimov: Of couse, but you can simple drop `@backstage/plugin-notifications-backend-module-email` from `packages/backend/package.json` and `packages/backend/src/index.ts`

workspaces/global-header/README.md
christoph-jerolimov: Of course it must not be called external_service_token then. This might confuse people.

I hope this is fine for you: https://github.com/redhat-developer/rhdh-plugins/pull/373/commits/f9731fe401d44f845951c322123491f9c6995301

If not please really let me know. Then I'm sorry. :)

christoph-jerolimov (approved):

PR #381, opened 2025-02-03
fix(orchestarator):Disable abort when workflow execution does not exist

LiorSoffer:
resolve [FLPATH-1080](https://issues.redhat.com/browse/FLPATH-1080)

https://github.com/user-attachments/assets/2f1e7503-e0d0-498c-ab0a-6e840402beb2

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
@@ -42,6 +42,7 @@ import Snackbar from '@material-ui/core/Snackbar';
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 import CloseIcon from '@material-ui/icons/Close';
 import ErrorIcon from '@material-ui/icons/Error';
+import { AlertTitle } from '@material-ui/lab';
 import Alert from '@material-ui/lab/Alert';
 import ArrowDropDown from '@mui/icons-material/ArrowDropDown';
 import StartIcon from '@mui/icons-material/Start';
@@ -70,11 +71,18 @@ import { WorkflowInstancePageContent } from './WorkflowInstancePageContent';
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     abortButton: {
-      backgroundColor: theme.palette.error.main,
-      color: theme.palette.getContrastText(theme.palette.error.main),
+      backgroundColor: theme.palette.error.dark,
+      color: theme.palette.getContrastText(theme.palette.error.dark),
       '&:hover': {
         backgroundColor: theme.palette.error.dark,
+        filter: 'brightness(90%)',
       },
+    },
+    modalText: {
+      fontSize: '1.1rem',
+    },
+    errorColor: {
+      color: theme.palette.error.dark,
     },
   }),
 );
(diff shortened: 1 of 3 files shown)


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -82,6 +82,9 @@ const useStyles = makeStyles((theme: Theme) =>
 export type AbortConfirmationDialogActionsProps = {
   handleSubmit: () => void;
   handleCancel: () => void;
+  isAborting: boolean;
+  canAbort: boolean;
+  classes: ReturnType<typeof useStyles>;
```
mareklibra, line 87: Can we use `const classes = useStyles();` instead?

mareklibra: yes, sort of:
```
const AbortConfirmationDialogActions = (
  props: AbortConfirmationDialogActionsProps,
) => {
const classes = useStyles();

return (<><Button>....</>)
}
```

gciavarrini:
I know that the issue is exactly about disabling the "abort" button, and this PR matches the requirement. But in terms of user experience, i think it isn't clear why the abort is unavailable. I mean, this can be confusing for the user. 
Can we consider showing a different popup when the workflow can't be aborted? or to give some info to help the user?

gabriel-farache:
> i think it isn't clear why the abort is unavailable

@gciavarrini I agree with you that having the button disappears is a bit unexpected but I think it's obvious that the disappearance is caused by the abortion of the workflow so it cannot be aborted anymore

My personal preference would have been to grey-out the button so the layer does not change, the button is still there but it is not actionable anymore. But I am no UX designer :)

> Can we consider showing a different popup when the workflow can't be aborted?

Having a clickable button then upon the action showing a pop-up saying "No you can't do that, the workflow is not running" is a little not-nice to me: I hate pop-up as they are interrupting the browsing flow
Having a tool tip on the disabled/greyed-out button stating that would be a good idea though

LiorSoffer:
There are two "Abort" buttons:

The first one opens the "Are you sure?" confirmation dialog.

The second one (red button) actually aborts the instance.
In the video, the red "Abort" button becomes disabled because the process completes before the user clicks it. This is visible under "Results," where the status changes from "Running" to "Completed".

When an instance isn't abortable:
The first "Abort" button is removed from the screen.
If the confirmation dialog is already open, the red "Abort" button becomes disabled.
I can ask Shiran to disable the first button instead of removing it. What do you think? @gciavarrini @gabriel-farache

LiorSoffer:
There are two "Abort" buttons:
-The first one opens the "Are you sure?" confirmation dialog.
-The second one (red button) actually aborts the instance.

When an instance isn't abortable:
The first "Abort" button is removed from the screen.
If the confirmation dialog is already open, the red "Abort" button becomes disabled.

In the video, the red "Abort" button becomes disabled because the process completes before the user clicks the button.
This is visible under "Results," where the status changes from "Running" to "Completed" (in red).
What do you think should be the behavior in this case? @ShiranHi

ShiranHi:
Thank you, I have some comments: 

Two small comments:
1. Modal UI ([design](https://www.figma.com/design/XjdTCb4vDQ7L4Jz1TdHKGq/RHDH---Workflow-Orchestrator?node-id=4930-2317&t=aMQa66LsIMCxB8RG-1)):
- Title should be "Abort workflow run"
- The text on the modal should not be bold but medium 
- It seems the red color is not the right shade
- We should add padding below the buttons 

2. When users click "Abort" on the modal we should close it

LiorSoffer:
@ShiranHi Thank you for the comments.
In this case the user didn't click abort on the modal :) This video demonstrates a case where the user opens the modal and is contemplating whether to abort the instance or not. Meanwhile, the run completes, causing the abort button to become disabled.

ShiranHi:
@LiorSoffer go it! 
In case, I think we need to inform users on the modal what happened, like [here](https://www.figma.com/design/XjdTCb4vDQ7L4Jz1TdHKGq/RHDH---Workflow-Orchestrator?node-id=5311-3448&t=CNIrthsK0STjoV2a-1).

LiorSoffer:
@ShiranHi So, the title has been changed, the text is bigger, padding has been added, and the information now appears when the instance isn't abortable.

The button is in `theme.palette.error.main` color, but on hover, it changes to `theme.palette.error.dark` color, which is closer to the design. I noticed that in other parts of the plugin, we also use the main shade. What do you think?

ShiranHi:
> @ShiranHi So, the title has been changed, the text is bigger, padding has been added, and the information now appears when the instance isn't abortable.

Thank you, could you please add a screenshot/screen recording of it?

LiorSoffer:
@mareklibra Am I allowed to change these? `theme.palette.error.main` `theme.palette.error.dark`

mareklibra:
> Am I allowed to change these?

Of course you are. Just be prepared for more reviewers and opinions.

LiorSoffer:
@Shiran I fixed the colors and padding plus aligned the icon with the text and action buttons

ShiranHi:
> @shiran I fixed the colors and padding plus aligned the icon with the text and action buttons

Thank you! Could you please add a screenshot so I can see it?

LiorSoffer:
@ShiranHi the video is already updated :)

ShiranHi:
> @ShiranHi the video is already updated :)

Thank you! Looks good

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -70,36 +71,89 @@ import { WorkflowInstancePageContent } from './WorkflowInstancePageContent';
...
+  return (
+    <>
+      <Button
+        onClick={props.handleSubmit}
+        variant="contained"
+        className={classes.abortButton}
+        startIcon={props.isAborting ? <CircularProgress size="1rem" /> : null}
+        disabled={props.isAborting || !props.canAbort}
+      >
+        {' '}
```
mareklibra, line 136: Is this `{' '}` an automatic addition by prettier? Let;'s remove it.

mareklibra (commented):
LGTM, just a nit-pick

mareklibra (approved):

PR #386, opened 2025-02-04
fix(orchestrator):implement filters

LiorSoffer:
resolve [FLPATH-1989](https://issues.redhat.com/browse/FLPATH-1989)

https://github.com/user-attachments/assets/508eae5c-2852-4bc8-a178-a0e07ace903f


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -88,25 +109,86 @@ export const WorkflowRunsTabContent = () => {
...
+
+    let startedFilter: FieldFilter | undefined = undefined;
+
+    if (startedSelectorValue !== Selector.AllItems) {
+      let dateRange: [string, string] | undefined = undefined;
+
+      const currentDate = new Date();
+      const endOfToday = new Date(currentDate);
+      endOfToday.setHours(23, 59, 59, 999);
+
```
gciavarrini, line 121: If I'm not wrong, `new Date()` uses the tz in which the system is running. Can you please double check if it's ok?
what if the user tz is different from system tz?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowRunsTabContent.tsx
```diff
@@ -88,25 +109,86 @@ export const WorkflowRunsTabContent = () => {
...
+        filters,
       };
-    } else if (statusFilter) {
-      return statusFilter;
-    } else if (workflowIdFilter) {
-      return workflowIdFilter;
     }
-    return undefined;
-  }, [statusSelectorValue, workflowId]);
+    return filters[0] || undefined;
```
gciavarrini, line 190: why `filters[0]`? can't be there more than one valid filter?

mareklibra: This looks good to me since `if (filters.length > 1)` above handles that case

mareklibra (approved):

PR #387, opened 2025-02-05
fix(orchestrator): improved devMode, added podman and macos support

gautamkrishnar:
improved devMode to make multi platform development easier

- Added Podman support (opinionated, people can use Podman. Docker is still kept as default) 
- Added M1 (ARM) Mac support
- Added DB persistence in sonata flow dev mode image (optional). This will improve the startup time of the dev mode container. 
- Moved temp directory to `.devModeTemp` folder under the `backend` plugin directory to maintain podman support. Podman does not have access to `/tmp` by default.

### DIFF
--- a/workspaces/orchestrator/app-config.yaml
+++ b/workspaces/orchestrator/app-config.yaml
@@ -124,14 +124,26 @@ dynamicPlugins:
   frontend: {}
 orchestrator:
   sonataFlowService:
+    # uncomment the next line to use podman instead of docker
+    # runtime: podman
     baseUrl: http://localhost
     port: 8899
     autoStart: true
+    # If you're using a MacBook with M-series chips, uncomment the line below to use ARM images.
+    # To pull the image, you'll first need to authenticate with the registry.
+    # For more information, visit: https://red.ht/410mMNU.
+    # container:  registry.redhat.io/openshift-serverless-1/logic-swf-devmode-rhel8:1.35.0
     # notsecret
     notificationsBearerToken: bXljdXJscGFzc3dkCg==
     notificationsUrl: http://host.docker.internal:7007
+    # uncomment the next line if you are using podman:
+    # notificationsUrl: http://host.containers.internal:7007
+    # By default the Dev Mode workflows are Ephemeral. If you want to persist the data across restarts,
+    # uncomment the next two lines, read more: https://www.rhdhorchestrator.io/blog/devmode-vs-prodmode/
+    # persistence:
+    #  path: ./.devModeTemp/db_persistence # this will be under packages/backend
     workflowsSource:
       gitRepositoryUrl: https://github.com/parodos-dev/backstage-orchestrator-workflows
-      localPath: /tmp/orchestrator/repository
+      localPath: ./.devModeTemp/repository # this will be under packages/backend
   dataIndexService:
     url: http://localhost:8899
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
@@ -34,6 +34,7 @@ import { executeWithRetry } from './Helper';
 
 const SONATA_FLOW_RESOURCES_PATH =
   '/home/kogito/serverless-workflow-project/src/main/resources';
+const DEFAULT_SONATAFLOW_RUNTIME = 'docker';
 
 interface LauncherCommand {
   command: string;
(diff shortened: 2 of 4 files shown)


workspaces/orchestrator/app-config.yaml
```diff
@@ -124,14 +124,25 @@ dynamicPlugins:
...
+    # To pull the image, you'll first need to authenticate with the registry.
+    # For more information, visit: https://red.ht/410mMNU.
+    # container:  registry.redhat.io/openshift-serverless-1/logic-swf-devmode-rhel8:1.35.0
     # notsecret
     notificationsBearerToken: bXljdXJscGFzc3dkCg==
     notificationsUrl: http://host.docker.internal:7007
+    # uncomment for podman:
+    # notificationsUrl: http://host.containers.internal:7007
+    persistence:
+      path: ./.devModeTemp/db_persistence # this will be under packages/backend
```
gciavarrini, line 143: Unsure about the `persistance`. According to [our docs](https://www.rhdhorchestrator.io/blog/devmode-vs-prodmode/), devMode has no persistence. @masayag wdyt?

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
```diff
@@ -155,10 +159,17 @@ export class DevModeService {
       '-v',
       `${resourcesAbsPath}:${SONATA_FLOW_RESOURCES_PATH}:Z`,
     );
+    if (!fs.existsSync(persistenceAbsPath)) {
+      fs.mkdirSync(persistenceAbsPath, { recursive: true });
+    }
+    launcherArgs.push(
+      '-v',
+      `${persistenceAbsPath}:${DEFAULT_SONATAFLOW_PERSISTENCE_PATH}:Z`,
+    );
```
gciavarrini, line 168: what if i want to use a custom persistence path? It seems that, if the `persistenceAbsPath` exists then the `DEFAULT_SONATAFLOW_PERSISTENCE_PATH` won't be used in any case. Can you please clarify?

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
```diff
@@ -155,10 +159,17 @@ export class DevModeService {
...
     launcherArgs.push('-e', 'KOGITO.CODEGEN.PROCESS.FAILONERROR=false');
     launcherArgs.push(
       '-e',
-      `QUARKUS_EMBEDDED_POSTGRESQL_DATA_DIR=${this.connection.persistencePath}`,
+      `QUARKUS_EMBEDDED_POSTGRESQL_DATA_DIR=${DEFAULT_SONATAFLOW_PERSISTENCE_PATH}`,
```
gciavarrini, line 172: need help to understand here: why we don't consider `persistenceAbsPath` anymore?

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
```diff
@@ -197,7 +208,7 @@ export class DevModeService {
     const persistencePath =
       config.getOptionalString(
         'orchestrator.sonataFlowService.persistence.path',
-      ) ?? DEFAULT_SONATAFLOW_PERSISTENCE_PATH;
+      ) ?? './.devModeTemp/db_persistence';
```
gciavarrini, line 211: I'd like to keep the default as variable instead of using an hardcode path. Is it possible?

workspaces/orchestrator/app-config.yaml
masayag: In dev mode, the postgres server is an embedded one by default.
But it is there, and workflow runs can survive pod restart.
So this property just exposes it - which seems good to me.

gautamkrishnar: Thanks for confirming @masayag i will make some changes to make it optional so that developers can chose what they want

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
gautamkrishnar: Great idea, will do that change.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DevModeService.ts
gautamkrishnar: @gciavarrini Great idea, will do that change.

gciavarrini (approved):
lgtm

gciavarrini:
@masayag PTAL

mareklibra (approved):

masayag (approved):

gautamkrishnar:
Thanks for the review team 🎉

PR #397, opened 2025-02-10
fix(orchestrator):Standardise the page loading animation

LiorSoffer:
resolve [FLPATH-772](https://issues.redhat.com/browse/FLPATH-772)

I made loading times deliberately longer for this video , unless it's too fast to see:

https://github.com/user-attachments/assets/85e1d132-c8f2-4918-bdb0-c112fd9b70db

### DIFF
--- a/workspaces/orchestrator/.changeset/yellow-humans-share.md
+++ b/workspaces/orchestrator/.changeset/yellow-humans-share.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Standardise the page loading animation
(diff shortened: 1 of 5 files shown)


gciavarrini:
@LiorSoffer need help to understand this change. 
Reading the jira issue it seems that when moving between tabs (workflows and all runs) they expect to see the loading bar on the top. Can you clarify? Can you add a before/after screen recording to understand better the change? 🙏🏻

LiorSoffer:
@gciavarrini your'e right I forgot to add the video which I sent to Shiran, now it's added.
The idea is that If we already know the page layout, it's better to display it rather than showing a blank screen. 
For example, if data rows are still loading, it's better to show the row titles and spinner.

gciavarrini (approved):
lgtm

workspaces/orchestrator/packages/backend/.devModeTemp/repository
ciiay, line 1: Hi @LiorSoffer , is this `Submodule repository added at 47d3b5` change intentionally? It causes ci failures, such as

https://github.com/redhat-developer/rhdh-plugins/actions/runs/13269506218

PR #403, opened 2025-02-11
fix(orchestrator):[UI] Layout errors with workflow details screen

LiorSoffer:
resolve [FLPATH-2066](https://issues.redhat.com/browse/FLPATH-2066)

Screenshot from 2025-02-12 18-11-57
Screenshot from 2025-02-12 18-13-13
Screenshot from 2025-02-12 18-13-45

### DIFF
--- a/workspaces/orchestrator/.changeset/popular-months-watch.md
+++ b/workspaces/orchestrator/.changeset/popular-months-watch.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Update the workflow details screen to match the latest Figma design
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
@@ -200,6 +200,7 @@ export const WorkflowInstancePage = ({
   const handleAbort = React.useCallback(async () => {
     if (value) {
       setIsAborting(true);
+
       try {
         await orchestratorApi.abortWorkflowInstance(value.instance.id);
         restart();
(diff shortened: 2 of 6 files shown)


LiorSoffer:
@gciavarrini  @ShiranHi  we're using `AboutField` from '@backstage for the label and value.
Regarding the bottom padding, I  reduced the card's height maybe it should be even smaller , I'll tag Shiran :)

ShiranHi:
> @ShiranHi we're using `AboutField` from '@backstage for the label and value.

I'm not sure what `AboutField` means 🤔

I agree with the comment from @gciavarrini about the paddings. Also, there seems to be plenty of space for the description on the right, can we make it wider to reduce scrolling?

ShiranHi:
Thank you Yi, it looks better now!
A question - Is there a way to make the height adjustable based on the description text to avoid scrolling, unless the text is very long and a max-height is applied?

LiorSoffer:
@ShiranHi Yes :) I used max of `15 rem` which is relative to the root element's font size.

gciavarrini (approved):
lgtm

gciavarrini:
@LiorSoffer Must be cherry-picked into `orchestrator-1.4` branch

workspaces/orchestrator/.changeset/popular-months-watch.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+[UI] Layout errors with workflow details screen
```
mareklibra, line 5: This message will be part of the changelog, one of many. I think this will not be clear to the reader without knowing wider context. A clear sentence will be better.

mareklibra (approved):

mareklibra:
LGTM, just rebase is needed

PR #405, opened 2025-02-12
fix(theme): Fixing fontColor error while customizing rhdh appearance

its-mitesh-kumar:
Fixed [RHIDP-5902](https://issues.redhat.com/browse/RHIDP-5902)


christoph-jerolimov (approved):
Thanks for that fix

PR #407, opened 2025-02-12
fix(orchestrator):remove workflows repo from backend

LiorSoffer:


christoph-jerolimov (approved):
I think the failing tests here are "okay". That change should solve that issue. We can merge it anyway when someone else from orchestrator is fine with this change.

jkilzi (approved):
LGTM. OK, to merge this as is

christoph-jerolimov:
Thanks for confirmation. Done. Let see if this helps. :)

christoph-jerolimov:
Yep, it worked even for old PRs that wasn't rebased.

Thanks for the quick solution @LiorSoffer and feeback @jkilzi

PR #418, opened 2025-02-16
chore(orchestrator): upgrade backstage to 1.36

mareklibra:
The orchestrator-development Backstage is upgraded to the latest 1.36-next.

We need this to unblock testing and development of new features. Once stable 1.36.0 is released (in a few weeks), we will upgrade one more time.


workspaces/orchestrator/packages/backend/Dockerfile
```diff
@@ -40,6 +40,7 @@ WORKDIR /app
 # Copy files needed by Yarn
 COPY --chown=node:node .yarn ./.yarn
 COPY --chown=node:node .yarnrc.yml ./
+COPY --chown=node:node backstage.json ./
```
mareklibra, line 43: SonarCloud is complaining here. I do not see any risk keeping it as it is. Or it should be fixed systematically for the rest of the Dockerfile as well.

PR #425, opened 2025-02-17
orchestrator: add endpoint for calling mutation ExecuteAfter

ydayagi:
DO NOT MERGE BEFORE https://github.com/redhat-developer/rhdh-plugins/pull/306. This one depends on it.


workspaces/orchestrator/plugins/orchestrator-backend/src/service/api/v2.ts
```diff
@@ -141,9 +141,16 @@ export class V2 {
...
+    if (typeof instance.variables === 'string') {
+      variables = JSON.parse(instance?.variables);
+    } else {
+      variables = instance.variables;
+    }
+
+    if (includeAssessment && variables?.orchestratorAssessmentInstanceId) {
       assessedByInstance = await this.orchestratorService.fetchInstance({
-        instanceId: instance.businessKey,
+        instanceId: variables.orchestratorAssessmentInstanceId as string,
```
mareklibra, line 153: Do we need the `as string` here? I guess not thanks to the check on line 151

workspaces/orchestrator/plugins/orchestrator-backend/src/service/api/v2.ts
```diff
@@ -173,10 +179,15 @@ export class V2 {
     }
     const executionResponse = await this.orchestratorService.executeWorkflow({
       definitionId: workflowId,
-      inputData:
-        executeWorkflowRequestDTO.inputData as ProcessInstanceVariables,
+      definitionVersion: definition.version,
+      inputData: {
```
mareklibra, line 183: This part looks suspicious, can you please confirm it is as expected?

When looking at changes around https://github.com/redhat-developer/rhdh-plugins/pull/425/files#diff-dfd9240b67dcd5abbff3560a8d0f341802d928a817c13e2350a999bb2751d5fdR588 , there seems to be just an addition of fields but the change here puts `executeWorkflowRequestDTO.inputData` into a sub-property.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/api/v2.ts
ydayagi: line 151 helps for the case that it is undefined not for the type. because it is of type unknown (the field orchestratorAssessmentInstanceId) I get compile error. so, yes, it is needed.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/api/v2.ts
ydayagi: indeed this change is required. this is the new implementation in sonataflow. if you have workflowdata field then it is treated as the input to the workflow execution and all the rest is stored without using it. If you do not have workflowdata then the whole variables object is treated as the input to the workflow execution. putting the input in the workflowdata field enables us to pass orchestratorAssessmentInstanceId into variables and having it stored and read when querying the workflow instance/execution/run.

PR #258, opened 2025-01-08
feat(fab): update fab content

debsmita1:
**Resolves:**
https://issues.redhat.com/browse/RHIDP-5398

**GIF/Screenshots:**

Added slide transition
https://github.com/user-attachments/assets/1b27522e-c0db-48d4-8a4c-ab623b442552

The below screenshot shows the following:
- Show an icon or an icon with a label, but not only a label
- Truncating text longer than 20 characters
- Have set default colors of the sub-menu buttons to be [Secondary](https://www.patternfly.org/design-foundations/colors/#brand-colors) `-pf-tglobalbackgroundcolorsecondary-default` , text to be `--pf-t--global--text--color--regular` and icon color to be `--pf-t--global--icon--color--regular`
- Kept 16px from the bottom and the left and right sides. Keep the button on the bottom-right or bottom left and remove middle alignment.
- Buttons leading to external links has icon to indicate that the link opens externally

Screenshot 2025-01-08 at 8 31 23 PM

Added instructions to recommend users to use only filled icons from the [Material Design library](https://fonts.google.com/icons)

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [x] Added or Updated documentation
- [x] Tests for new functionality and regression tests for bug fixes
- [x] Screenshots attached (for UI changes)

PR #427, opened 2025-02-17
fix(fab): fixing css issues

its-mitesh-kumar:
https://github.com/user-attachments/assets/fb2e5f21-4f60-4c2f-a4a9-63b3c9d2003a

Fixes https://issues.redhat.com/browse/RHIDP-5482


workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
```diff
@@ -58,6 +58,10 @@ const FABLabel = ({
             ...marginStyle,
             color: '#151515',
             order: 2,
+            textTransform: 'lowercase', // Convert entire text to lowercase
+            '&::first-letter': {
+              textTransform: 'uppercase', // Capitalize first letter
+            },
```
christoph-jerolimov, line 64: I don't think we should transform the label at all, and esp this way. The only option I see is that we want reset the default transformation.

christoph-jerolimov (changes requested):
Hi @its-mitesh-kumar, can you please add screenshots before and after to your PR.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
its-mitesh-kumar: As per requirement **Change the FAB text to sentence case, ex:  "Bulk import" instead of "BULK IMPORT"** , we need to transform the lable , yes better to do with javascript .

christoph-jerolimov: Yes and no. Maybe the wording wasn't exact. But instead of applying another set of text transform we should use no text transform here. Which means a simple `textTransform: 'unset'`

MUI by default converts all Buttons with text-transform to uppercase. With PatternFly we don't follow this rule.

In best case, we would fix that (override this default text-transform rule) in our theme. If customers want use the MUI theme, or their own theme, it should be up to the theme if the button is "UPPERCASE" or "Normal Case".

But there is no general rule how to handle for example this buttons:

* Create a ticket
* ServiceNow
* Open RHDH documentation

MUI solution works because it simple says everything is uppercase.

Our goal is that we want keep what the user (admin) configures.

If you like, you can also take a look into the theme. But with our short timeframe its also fine to apply it (first) here. I would prefer later a theme solution and remove this special rule here again.

christoph-jerolimov: @debsmita1 just fyi, I didn't reviewed the rest of the changes. :)

debsmita1 (changes requested):
Fab with text labels are appearing like this

Screenshot 2025-02-19 at 6 56 42 PM

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
its-mitesh-kumar: @christoph-jerolimov I'm updating `textTransform` to `'unset'`. Should we also create a ticket for the theme changes?

its-mitesh-kumar:
@ShiranHi @debsmita1 isux looks fine for you . 

https://github.com/user-attachments/assets/3e5081bd-882d-499d-87ab-2a748b3c1a88

workspaces/global-floating-action-button/packages/app/src/components/Root/Root.tsx
```diff
@@ -98,14 +98,14 @@ export const Root = ({ children }: PropsWithChildren<{}>) => {
           },
           {
             slot: Slot.BOTTOM_LEFT,
-            icon: <LibraryBooks />,
-            label: 'Api Docs',
-            toolTip: 'Api Docs',
+            icon: <ExtensionIcon />,
```
logonoff, line 101: nit: API is an acronym that should be in capitals

```suggestion
            icon: <ExtensionIcon />,
            label: 'APIs',
            toolTip: 'APIs',            
```

debsmita1:
As discussed on a call, my commit https://github.com/redhat-developer/rhdh-plugins/pull/427/commits/cc908815dc6b9225cdc5e477ee1da7f621f958a6 has some changes where I have tried Slide with Fade MUI components

christoph-jerolimov:
@debsmita1 can you please take a new look and check with @invincibleJai  if this is a 1.5.x or 1.6.0 change? Thanks.

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
```diff
@@ -122,39 +124,39 @@ export const FAB = ({
...
-          variant={
-            actionButton.showLabel || isExternal || !actionButton.icon
-              ? 'extended'
-              : 'circular'
+      <Fab
+        {...(newWindow ? { target: '_blank', rel: 'noopener' } : {})}
+        className={className}
+        style={{
+          color: iconColor ? iconColor : '#1f1f1f',
+          backgroundColor: !actionButton.color && !className ? 'white' : '',
```
debsmita1, line 132: Can you undo this change and test ?
I tried it myself and it looks like this change is not needed

debsmita1 (changes requested):
It looks much better now
Could you work on adding some delay to transition ?

its-mitesh-kumar:
@ShiranHi , Now transition looks much smooth , can you review and add lgtm 

https://github.com/user-attachments/assets/b2a33cd7-b9db-48e5-8e5d-ed34dd18607b

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/CustomFab.tsx
```diff
@@ -68,18 +71,20 @@ const FABLabel = ({
...
+    </>
   );
 };
 
-export const FAB = ({
+export const CustomFab = ({
   actionButton,
   size,
   className,
+  iconColor,
```
debsmita1, line 82: instead of adding a new prop to this Component you could add `iconColor` property to the `FloatingActionButton` type

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FloatingButton.tsx
```diff
@@ -78,20 +61,23 @@ export const FloatingButton = ({
...
     >
-      {fabs.length > 1 ? (
-        <FABWithSubmenu fabs={fabs} slot={slot} />
-      ) : (
-        <FAB actionButton={fabs[0]} />
-      )}
+      <CustomFab
+        actionButton={{ color: 'info', ...fabs[0] }}
+        iconColor="white"
+      />
```
debsmita1, line 80:
```suggestion
       <CustomFab
        actionButton={{ color: 'info', iconColor: 'white', ...fabs[0] }}
      />
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FABWithSubmenu.tsx
```diff
@@ -59,50 +73,65 @@ export const FABWithSubmenu = ({
...
+      {fabs?.map(fb => {
+        return (
+          <Slide
+            key={fb.label}
+            direction="up"
+            container={containerRef?.current}
+            in={isMenuOpen}
+            mountOnEnter
+            unmountOnExit
+            timeout={300}
```
debsmita1, line 122:
```suggestion
            timeout={500}
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/CustomFab.tsx
```diff
@@ -122,39 +124,39 @@ export const FAB = ({
...
-          style={{ color: '#1f1f1f' }}
-          variant={
-            actionButton.showLabel || isExternal || !actionButton.icon
-              ? 'extended'
-              : 'circular'
+      <Fab
+        {...(newWindow ? { target: '_blank', rel: 'noopener' } : {})}
+        className={className}
+        style={{
+          color: iconColor || '#1f1f1f',
```
debsmita1, line 131:
```suggestion
          color: actionButton.iconColor || '#1f1f1f',
```

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/CustomFab.tsx
```diff
@@ -122,39 +124,39 @@ export const FAB = ({
...
+          backgroundColor: !actionButton.color && !className ? 'white' : '',
```
debsmita1, line 132: why is this needed ? and why can't we use `getColor` util to evaluate the background color for the `color` prop in line no 140 ?

debsmita1 (changes requested):

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/CustomFab.tsx
its-mitesh-kumar: Because the `color` prop of the MUI Fab component expects a value from the predefined set ("primary", "secondary", "error", "info", "success", "warning", "inherit") and does not accept arbitrary colors like "white"

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FABWithSubmenu.tsx
```diff
@@ -15,40 +15,54 @@
...
-    paddingTop: '10px',
+    zIndex: 205,
     color:
       theme && Object.keys(theme).length > 0
         ? theme.palette.grey[500]
         : '#9e9e9e',
   },
   menuButtonStyle: {
-    color: '#1f1f1f',
+    color: 'white',
```
debsmita1, line 50: Based on Christoph's comment [here](https://github.com/redhat-developer/rhdh-plugins/pull/258#discussion_r1908474925)

We can use something like
```
color: theme?.palette?.getContrastText(theme.palette.background.default),
```

With this, the icon color changes based on the theme to evaluate the readable text color

In light mode
Screenshot 2025-03-13 at 6 00 23 PM


In dark mode
Screenshot 2025-03-13 at 6 00 39 PM


@ShiranHi Your thoughts on this ?
cc @christoph-jerolimov

ShiranHi: That's great, but can we keep the icon white for both light and dark modes?

debsmita1: We can by hardcoding it
@christoph-jerolimov are you okay with this ?

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/CustomFab.tsx
```diff
@@ -58,6 +58,9 @@ const FABLabel = ({
             ...marginStyle,
             color: '#151515',
             order: 2,
+            textTransform: 'none',
+            whiteSpace: 'nowrap',
+            textOverflow: 'ellipsis',
```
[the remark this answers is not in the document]
its-mitesh-kumar: Now Removed . 
Screenshot 2025-03-17 at 5 48 02 PM (2)

debsmita1 (approved):
/lgtm

Thank you @its-mitesh-kumar !

workspaces/global-floating-action-button/plugins/global-floating-action-button/src/components/FAB.tsx
christoph-jerolimov: Yes a ticket for the theme would be great. Thanks.

christoph-jerolimov (approved):

PR #444, opened 2025-02-25
test: run playwright dev mode tests on CI

jrichter1:

- adds CI steps to run playwright tests in dev mode for workspaces that contain a playwright config file
- updates playwright to 1.50.1 across the board to avoid issues with new chrome headless mode
- fixes a test in `homepage` workspace because it lands on a different page than it expected

https://issues.redhat.com/browse/RHIDP-6140

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,8 +115,6 @@ jobs:
       - name: run playwright tests
         if: ${{ hashFiles(format('workspaces/{0}/playwright.config.ts', matrix.workspace)) != '' }}
         run: yarn playwright test
-        env:
-          DEVMODE: true
 
       - name: ensure clean working directory
         run: |
--- a/workspaces/global-header/playwright.config.ts
+++ b/workspaces/global-header/playwright.config.ts
@@ -17,10 +17,6 @@
 import { defineConfig } from '@playwright/test';
 import { generateProjects } from '@backstage/e2e-test-utils/playwright';
 
-const devmode = process.env.DEVMODE
-  ? true
-  : !process.env.PLAYWRIGHT_URL && !process.env.CI;
-
 /**
  * See https://playwright.dev/docs/test-configuration.
  */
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -108,6 +108,16 @@ jobs:
       - name: test changed packages
         run: yarn backstage-cli repo test --coverage --maxWorkers=3
 
+      - name: install playwright
+        if: ${{ hashFiles(format('workspaces/{0}/playwright.config.ts', matrix.workspace)) != '' }}
+        run: npx playwright install --with-deps chromium chrome
```
christoph-jerolimov, line 113: I think we should use yarn here

.github/workflows/ci.yml
```diff
@@ -108,6 +108,16 @@ jobs:
...
+
+      - name: run playwright tests
+        if: ${{ hashFiles(format('workspaces/{0}/playwright.config.ts', matrix.workspace)) != '' }}
+        run: npx playwright test
+        env:
+          DEVMODE: true
```
christoph-jerolimov, line 119: Why do we need another env here? Isn't CI good enough and DEVMODE is something that someone might use locally?

christoph-jerolimov (commented):
Thanks @jrichter1, that's awesome! Some thoughts:

.github/workflows/ci.yml
jrichter1: I could use some opinions here. The status quo in playwright configs is that `CI` disables dev mode. I'm trying to run dev mode on CI. Should I just flip the condition then?

jrichter1: I have a feeling you'll like the new version more (since it was your idea) :laughing:

christoph-jerolimov (approved):
I could see in the CI jobs that Playwright was installed successfully and run successfully for all these touched workspaces.

It's a nice improvement. I think we should merge this! Any other opinions or approvals?

PR #452, opened 2025-02-27
Customizable (per-user) home-page grid layout

rafaeltuelho:

This PR introduces the Backstage [Customizable home page](https://github.com/backstage/backstage/blob/master/plugins/home/README.md#customizable-home-page) capability to RHDH dynamic homepage plugin.

NOTE: This change does not affect the original `HomePage` as it introduces this new capability through new components. So, the user can choose its HomePage implementation via the plugin configuration (dynamic routes).

Sample configuration using the new `DynamicCustomizableHomePage` component:
```yaml
    pluginConfig:
      dynamicPlugins:
        frontend:
          red-hat-developer-hub.backstage-plugin-dynamic-home-page:
            dynamicRoutes:
              - path: /
                importName: DynamicCustomizableHomePage # instead of the original DynamicHomePage
            mountPoints:
              - mountPoint: home.page/cards
                importName: SearchBar
                config:
                  layouts:
                    xl: { w: 10, h: 2, x: 1, y: 3 }
                    lg: { w: 10, h: 2, x: 1, y: 3 }
                    md: { w: 10, h: 2, x: 1, y: 3 }
                    sm: { w: 10, h: 2, x: 1 }
                    xs: { w: 12, h: 2 }
                    xxs: { w: 12, h: 2 }
```

### screenshots

* https://github.com/user-attachments/assets/2938181e-0157-4f1c-8478-377d0abde52f


christoph-jerolimov:
Hi @rafaeltuelho ,

thanks for this change. It's great to have both options. Since this is optional its good to have it. The CI job fails because a report.api.md must be regenerated. Please run this in the workspace:

```
yarn tsc && yarn build:api-reports:only
```

Do you think it's possible that you contribute also that the default grid are used from the admin configuration??

rafaeltuelho:
> Hi @rafaeltuelho ,
> 
> thanks for this change. It's great to have both options. Since this is optional its good to have it. The CI job fails because a report.api.md must be regenerated. Please run this in the workspace:
> 
> ```
> yarn tsc && yarn build:api-reports:only
> ```
> 
> Do you think it's possible that you contribute also that the default grid are used from the admin configuration??

Sure, I can take a look at it.

rafaeltuelho:
@christoph-jerolimov, concerning your suggestion

>... that the default grid is used from the admin configuration?

I spent some time looking for ways to allow switching between the two HomePage implementations via Admin view, and I realized it would not be straightforward. I would need to access the declared DynamicRoutes config via `Scalprum` like in https://github.com/redhat-developer/rhdh/blob/main/packages/app/src/components/DynamicRoot/ScalprumRoot.tsx
but many of them are not exposed to external consumption. 

I guess for now, allowing the userto  switch via configuration would be a good start.

christoph-jerolimov:
Hi @rafaeltuelho,

thanks for this contribution! It looks nice and works fine for me. :+1: 

Fyi: I've added two commits to your branch.

* The first one extract the duplicated code for the `Header` and uses that from the old `HomePage` and your new `CustomizableHomePage`.
* The 2nd adds a simple new doc "User customization" how to enable your feature. I've mentioned there that this is a dev preview. We will not document that feature in our official RHDH doc. For that, this feature isn't working "well enough". I think it would be great to load the default cards that an admin configured. So that the default shows what an admin configures, right? It should also show under the Add option all available cards.

I know that this additional work is not straight forward. But is there any chance that you will continue on that? :)

PR #470, opened 2025-03-04
feat(marketplace-cli): Create marketplace-cli subcommand to generate CSV files from YAML files

logonoff:
implements https://issues.redhat.com/browse/RHIDP-6231


logonoff:
WIP PR to implement the CSV subcommand.

@christoph-jerolimov am I on the right track/reading from the right files?

christoph-jerolimov:
> WIP PR to implement the CSV subcommand.
> 
> @christoph-jerolimov am I on the right track/reading from the right files?

Yeah that's a really good direction. :+1:

Please extend this so that you read all YAMLs from theplugins and packages folder and see if we can have a combination of both.

We don't know the exact data we want export, but it would be great if you can show for each plugin (contains the most metadata) show the related frontend and backend package. Incl. the npm package name and version.

These are linked via their backstage catalog entity names.

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  /** The columns to be included in the CSV. */
+  private readonly columns: {
+    /**
+     * The column key to be used in the CSV header. It must not contain commas,
+     * newlines, quotes, or anything else that could break the CSV format. It
+     * does not have to be unique.
+     */
+    key: string;
+    /** The getter function for each cell. The return value must be castable to a string */
+    getter: (plugin: T) => any;
```
christoph-jerolimov, line 40: tiny: if this CSVGenerator is generic, try to avoid "plugin" code or variables here.

I would also recommend to avoid any. Can it maybe return a `string | number`?

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+
+          // read each individual plugin from the file
+          const plugins = fileContent
+            .split('\n---\n')
+            .filter(Boolean)
+            .map((plugin: string) => {
+              return YAML.parse(plugin);
+            })
+            .filter(isEntity);
+          return plugins;
```
christoph-jerolimov, line 143: Hey, the YAML library supports parsing multiple docs via `YAML.parseAllDocuments`.  Each document should have something like toJSON.

```suggestion
          const plugins = YAML.parseAllDocuments(fileContent)
            .map((doc => doc.toJSON())
            .filter(isMarketplacePlugin)
          return plugins;
```

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+ * @param packageName The name of the package to fetch
+ * @returns The package.json object
+ */
+const fetchNpmPackageJson = async (packageName: string, version: string) => {
+  const response = await fetch(
+    `https://registry.npmjs.org/${packageName}/${version}`,
+  );
+
+  return await response.json();
+};
```
christoph-jerolimov, line 160: This is a "nice idea" but there is no need to fetch external information. All information should be part of the Plugin or Package YAML.

Esp. the descriptions and version strings. You can drop that code.

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+
+type CombinedPackage = {
+  yaml: MarketplacePackage;
+  pkg: PackageJson;
+};
```
christoph-jerolimov, line 165: The idea is good, but could be something like this:

```suggestion
type CombinedPackage = {
  plugin: MarketplacePlugin;
  frontendPackage?: MarketplacePackage;
  backendPackage?: MarketplacePackage;
  otherPackages?: MarketplcaePackage[];
};
```

The plugin and packages YAMLs contains many information and you should parse them, find the relationships and render "grouped information". The main entity == one row should be the plugin.

But a plugin can link zero, one or more packages.

For example:

https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/plugins/kubernetes.yaml#L91-L93
https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/packages/backstage-plugin-kubernetes.yaml#L4
https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/packages/backstage-plugin-kubernetes-backend.yaml#L4

So the CLI should load all plugins and all packages. At the end, you should iterate over all plugins and try to find related Packages.

The `spec.backstage.role` identifies if the Package is a a frontend or backend plugin:

https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/packages/backstage-plugin-kubernetes.yaml#L22-L23
https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/packages/backstage-plugin-kubernetes-backend.yaml#L22-L23

And then you can display in the CSV the Plugin name, description (and so one), and the frontend and backend package version.

Does this make sense to you?

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+  }
+
+  const yamls = parseYamls(pluginsYamlPath.split(','));
+
+  /** The headers we want to include in the CSV */
+  const headers = [
+    'name',
+    'title',
+    'description',
+    'developer',
```
christoph-jerolimov, line 180: We used author and developer in some YAMLs. The final version uses "author"

```suggestion
    'author',
```

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,241 @@
...
+    'categories',
+    'lifecycle',
+    'icon',
```
christoph-jerolimov, line 183: We used "icon" with an URL while developing the plugin. In the current version we use inline base64 icons. I guess we should skip this information until we have another soliution.  

```suggestion
```

christoph-jerolimov (changes requested):
Hi @logonoff, good work so far. I hope the comments below helps:

workspaces/marketplace/examples/plugins/tekton.yaml
```diff
@@ -14,6 +14,8 @@ spec:
   developer: Red Hat
   lifecycle: production
   owner: redhat
+  packages:
+    - 'backstage-community-plugin-tekton'
```
christoph-jerolimov, line 18: Can you rebase and revert this change. It's already added at the end of the file.

workspaces/marketplace/packages/cli/src/commands/export-csv/index.ts
```diff
@@ -0,0 +1,273 @@
...
+    if (isMarketplacePackage(yaml)) {
+      packages[yaml.metadata.name] = yaml;
+      packageCSV.addRow(yaml);
+    } else if (isMarketplacePlugin(yaml)) {
+      pluginCSV.addRow(yaml);
+    }
+  }
+
+  /** Generate the final CSV */
+  const finalCSV = `${packageCSV.generate()}${pluginCSV.generate(false)}`;
```
christoph-jerolimov, line 264: Okay. Two different "tables" in the same CSV doesn't look right to me.

Can you change that to have a merged table as we discussed.

With columns for

* Frontend package
* Frontend version
* Backend package
* Backend version
* Other packages

christoph-jerolimov (changes requested):

PR #483, opened 2025-03-07
feat(marketplace): Add marketplace providers to add entities into the catalog

karthikjeeyar:

This PR adds three new marketplace providers to add entities into the catalog.  The providers reads the plugins, package and collection entity metadata from RHDH image and adds to the catalog.

```
        catalog.addEntityProvider(new MarketplacePackageProvider(taskRunner));
        catalog.addEntityProvider(new MarketplacePluginProvider(taskRunner));
        // catalog.addEntityProvider(new MarketplaceCollectionProvider(taskRunner)); // not registered for 1.5 
```

## Screenshots:

image

## Unit tests

image

### DIFF
--- a/workspaces/marketplace/.changeset/shaggy-zoos-fetch.md
+++ b/workspaces/marketplace/.changeset/shaggy-zoos-fetch.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-catalog-backend-module-marketplace': patch
+---
+
+Add marketplace providers to add entities into catalog
--- a/workspaces/marketplace/package.json
+++ b/workspaces/marketplace/package.json
@@ -11,7 +11,7 @@
     "start-backend": "yarn workspace backend start",
     "build:backend": "yarn workspace backend build",
     "tsc": "tsc",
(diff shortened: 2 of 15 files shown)


gashcrumb (approved):
very nice!

karthikjeeyar:
Ran `yarn dedupe` and pushed the changes.

gashcrumb (approved):

karthikjeeyar:
Thank you @gashcrumb @christoph-jerolimov for the review.


> Just one thing I personally would do different: instead of finding a parent "marketplace" workspace, I would make the path configurable from the app-config.yaml. Like this:

>      local-provider:
>        globs:
>          - /marketplace/catalog/entities/**/*.yaml

Currently, the `/marketplace` folder is not colocated in the root src folder, it is placed in the Operating system's root folder in the RHDH image (not sure if this is intentional). 



We can look at adding all the configuration options for both `provider` and `processor` in 1.5.x as there can be more configuration options like relative glob paths, provider/processor is enabled or disabled, scheduler frequency config, skipping or filtering a kind via config etc.

 Considering that we have to think through all the configurations before we add them and the limited time we have to fix & test this issue, I am merging and releasing this PR with the current changes. Thanks.

PR #495, opened 2025-03-13
feat(global-header): changing create to self-service

its-mitesh-kumar:
Fixing : [Rename the 'Create' that appears on the Global header ](https://issues.redhat.com/browse/RHIDP-6164)

Screenshot from RHDH-Plugin
Screenshot 2025-03-13 at 6 54 02 PM (2)

ScreenShot from RHDH
Screenshot 2025-03-13 at 8 35 14 PM

### DIFF
--- a/workspaces/global-header/.changeset/quiet-bulldogs-battle.md
+++ b/workspaces/global-header/.changeset/quiet-bulldogs-battle.md
@@ -1,5 +1,5 @@
 ---
-'@red-hat-developer-hub/backstage-plugin-global-header': patch
+'@red-hat-developer-hub/backstage-plugin-global-header': minor
 ---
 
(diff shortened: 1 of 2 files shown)


christoph-jerolimov:
@its-mitesh-kumar @ciiay  I changed the changeset level from patch to minor so that this results in global-header 1.1.

So that we have RHDH 1.5 = global-header 1.0 and RHDH 1.6 = global-header 1.1 (or higher if we have more minor changes)

christoph-jerolimov (approved):
I think this is fine. The alternative might be adding translation for it... ;)

Let us keep this simple until we want share this with BCP maybe in the future. The I think we should have as part of the configuration (dynamic config props for example)

PR #501, opened 2025-03-17
fix(orchestrator):Limit access to workflow instances to initiators only

LiorSoffer:
First 2 commits  #306 and #349 
resolve [FLPATH-1916](https://issues.redhat.com/browse/FLPATH-1916)


workspaces/orchestrator/entities/users.yaml
```diff
@@ -8,3 +8,14 @@ spec:
     displayName: Marek Libra
     email: foo@xyzbar.cz
   memberOf: []
+---
```
ydayagi, line 11: please do not use your name in users file

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -798,9 +825,42 @@ function setupInternalRoutes(
...
+
+        let filters: Filter | undefined = requestFilters;
+
+        if (!isUserAuthorizedForInstancesAdminView) {
+          const EntityFilter: NestedFilterNested = {
+            operator: 'EQ',
+            value: entity,
+            field: 'initiatorEntity',
+          };
+          const nestedEntityFilter: NestedFilter = {
```
ydayagi, line 845: the names are confusing. i would use this name for the one above and call this one variablesFilter

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -798,9 +825,42 @@ function setupInternalRoutes(
...
+        const isUserAuthorizedForInstancesAdminView: boolean = // This permission will let user see All instances (including ones others created)
+          await isUserAuthorizedForInstancesAdminViewPermission(
+            req,
+            permissions,
+            httpAuth,
+          );
+
+        const requestFilters = getRequestFilters(req);
+
+        let filters: Filter | undefined = requestFilters;
```
ydayagi, line 837: you dont need to define the type for the variable

workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
```diff
@@ -221,9 +251,13 @@ function convertToBoolean(value: any): boolean {
 function formatValue(
   fieldName: string,
   fieldValue: any,
-  fieldDef: IntrospectionField,
+  fieldDef: IntrospectionField | undefined,
   type: ProcessType,
 ): string {
+  if (!fieldDef) {
+    return `"${fieldValue}"`;
+  }
```
gciavarrini, line 259: if `fieldDef` is undefined, it means that there is no such field in the introspection field list and so we don't know its format (i.e. string, date, ...).
Converting to string can be wrong.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts
```diff
@@ -270,7 +270,7 @@ export class DataIndexService {
     const graphQlQuery = buildGraphQlQuery({
       type: 'ProcessInstances',
       queryBody:
-        'id, processName, processId, businessKey, state, start, end, nodes { id }, variables, parentProcessInstance {id, processName, businessKey}',
+        'id, processName, processId, state, start, end, nodes { id }, variables, parentProcessInstance {id, processName, businessKey}',
```
gciavarrini, line 273: to understand: why removing businessKey?

workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
```diff
@@ -153,16 +178,21 @@ export function buildFilterCondition(
...
-    throw new Error(`Can't find field "${filters.field}" definition`);
-  }
+  let fieldDef;
 
-  if (!isOperatorAllowedForField(filters.operator, fieldDef, type)) {
-    const allowedOperators = supportedOperatorsByType[fieldDef.type.name] || [];
-    throw new Error(
-      `Unsupported operator ${filters.operator} for field "${fieldDef.name}" of type "${fieldDef.type.name}". Allowed operators are: ${allowedOperators.join(', ')}`,
-    );
+  if (!isNested) {
```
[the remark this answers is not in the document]
ydayagi: @gciavarrini  this is a change in my PR not Lior's. anyohw, you must have it that way. the filter relies on Introspection but not all filters have Introspection. In our case we don't have it. it a filter by field not in the schema. this is the only way to do it without creating 2 handleBinaryOperator functions.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts
ydayagi: this is a different PR. it is a requirement by piotr.

PR #510, opened 2025-03-18
fix(orchestrator): workflow result - add pending as a result

LiorSoffer:
resolve [FLPATH-2177](https://issues.redhat.com/browse/FLPATH-2177)
image

### DIFF
--- a/workspaces/orchestrator/.changeset/wet-glasses-dance.md
+++ b/workspaces/orchestrator/.changeset/wet-glasses-dance.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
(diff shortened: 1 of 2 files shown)


mareklibra:
@LiorSoffer Have you noticed any limitations how the `pending` state is reported by the backend?
Does something work differently than expected?
Any known differences between SonataFlow versions regarding that feature?

LiorSoffer:
@mareklibra moved it to draft, Moti said pending status isn't supported by Sonataflow for workflows.

mareklibra:
@LiorSoffer do we have already support from backend for this? If not, let's keep the PR as Draft to prevent merging it.

LiorSoffer:
This PR was closed because pending status isn't supported by Sonataflow for workflows.

PR #564, opened 2025-03-25
fix(orchestrator): Add Workflow Status to WorkflowsTable

LiorSoffer:
resolve [FLPATH-2201](https://issues.redhat.com/browse/FLPATH-2201)
Screenshot from 2025-03-25 13-34-26

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowsTable.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowsTable.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
@@ -20,8 +21,11 @@ import { Link, TableColumn, TableProps } from '@backstage/core-components';
 import { useRouteRef } from '@backstage/core-plugin-api';
 import { usePermission } from '@backstage/plugin-permission-react';
 
+import { Box, makeStyles, Tooltip } from '@material-ui/core';
 import FormatListBulleted from '@material-ui/icons/FormatListBulleted';
 import PlayArrow from '@material-ui/icons/PlayArrow';
+import TaskAltOutlinedIcon from '@mui/icons-material/TaskAltOutlined';
+import WarningAmberOutlinedIcon from '@mui/icons-material/WarningAmberOutlined';
 
 import {
   capitalize,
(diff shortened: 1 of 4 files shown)


workspaces/orchestrator/plugins/orchestrator/src/dataFormatters/WorkflowOverviewFormatter.ts
```diff
@@ -32,6 +33,7 @@ export interface FormattedWorkflowOverview {
   readonly category: string;
   readonly description: string;
   readonly format: WorkflowFormatDTO;
+  readonly availablity?: string;
```
mareklibra, line 36: Can this be of any other value then Available and Unavailable?
If not, I would keep that as a boolean and handle the text elsewhere.

LiorSoffer: It's originally a boolean, for my understanding this formatted purpose is to costume the data for the user so why do it somewhere else?

mareklibra: If so, please move the strings into constants, so the decision logic in the WorkflowsTable.tsx will be more stable.

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowsTable.tsx
```diff
@@ -13,15 +13,21 @@
...
+
 import React, { useCallback, useEffect, useMemo, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { Link, TableColumn, TableProps } from '@backstage/core-components';
 import { useRouteRef } from '@backstage/core-plugin-api';
 import { usePermission } from '@backstage/plugin-permission-react';
 
+import { Box, makeStyles, Tooltip } from '@material-ui/core';
+// import ErrorOutlineOutlinedIcon from '@material-ui/icons/ErrorOutlineOutlined';
```
mareklibra, line 25: Can you removed this, please?

workspaces/orchestrator/plugins/orchestrator/src/dataFormatters/WorkflowOverviewFormatter.ts
```diff
@@ -32,8 +33,18 @@ export interface FormattedWorkflowOverview {
...
 }
 
+export const available = 'Available';
```
mareklibra, line 39: To keep the same code-style, can this be named:
```suggestion
export const AVAILABLE = 'Available';
```

And I keep up to your decision if you prefer to keep it in this file or moving to `constants.ts`

mareklibra (changes requested):

workspaces/orchestrator/plugins/orchestrator/src/dataFormatters/WorkflowOverviewFormatter.ts
LiorSoffer: ohhh now I understand what  "please move the strings into constants" means, which way you prefer?

mareklibra: the constants.ts file :-)

mareklibra (approved):

PR #590, opened 2025-04-01
feat(marketplace): add tech preview notice to header

logonoff:
fixes https://issues.redhat.com/browse/RHIDP-6084

Adds a "Tech preview" banner to each page in the marketplace UI

after:

deviated from the mockup a bit because:

a) it better matches with OCP console

image

b) a banner component doesn't exist in MUI

c) having a div above the `Page` component introduces some issues with the container scrolling

### DIFF
--- a/workspaces/marketplace/.changeset/six-nails-drop.md
+++ b/workspaces/marketplace/.changeset/six-nails-drop.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-marketplace': minor
+---
+
+Add a Tech preview banner to the UI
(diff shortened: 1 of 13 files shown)


christoph-jerolimov:
I've asked @christophe-f on this and his response was:

> Let's keep this for later. If we go that path, we will have to update all plugins.  [...]
>
> Also, a page like the overview on the entity or homepage with several plugins will have all these banners.
> 
> I like the idea, but we have more important things to deliver.

So thanks @logonoff for that work but I think we should have a more generic solution in rhdh instead of the extensions plugin. Closing this for now.

christophe-f:
Forgot to mention that this plugin will be a community plugin.

PR #595, opened 2025-04-02
fix(orchestrator): fix view variables screen

LiorSoffer:
resolve [FLPATH-2202](https://issues.redhat.com/browse/FLPATH-2202)

image


ShiranHi:
Hey @LiorSoffer , it looks great! 
I'm wondering if we can reduce the size of the copy button so it will be like [here](https://www.patternfly.org/components/code-block) and if can add a bit of padding above and below the text inside the code block.

mareklibra:
Thanks, it's a big improvement.

mareklibra (approved):

PR #601, opened 2025-04-03
fix(orchestrator): run button should match figma

LiorSoffer:
resolve [https://issues.redhat.com/browse/FLPATH-2234](FLPATH-2234)
image


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstancePage.tsx
```diff
@@ -377,7 +377,15 @@ export const WorkflowInstancePage = ({
                     // }
                     style={{ color: 'white' }}
                   >
-                    Rerun
+                    {value.instance.state ===
+                    ProcessInstanceStatusDTO.Active ? (
+                      <>
+                        <CircularProgress color="inherit" size="0.75rem" />{' '}
```
mareklibra, line 383: No gap is too little, one is just right, two are too much.

Issue #607, opened 2025-04-04
Bug: Insights plugin


matdtr:
Hi, 

I was testing the Insights plugin, i'm on community backstage version 1.33.5 and node 18.19.0. Installation works good but navigating to the page both using identityApi or without is failing with:

```
Cannot read properties of undefined (reading '0')
Call Stack
 Users
  node_modules/@red-hat-developer-hub/backstage-plugin-adoption-insights/dist/components/Users/Users.esm.js:56:65
```

How to solve?

karthikjeeyar:
@matdtr Can you provide some more information to reproduce the issue.


How did you install the plugin using OCI images or steps mentioned in this [README file](https://github.com/redhat-developer/rhdh-plugins/tree/main/workspaces/adoption-insights/plugins/adoption-insights#installation) and what was the login mechanism used ?

karthikjeeyar:
Recommended Backstage version should be 1.35.1 for this plugin.

matdtr:
executed:

```
yarn --cwd packages/backend add @red-hat-developer-hub/backstage-plugin-adoption-insights-common
yarn --cwd packages/backend add @red-hat-developer-hub/backstage-plugin-adoption-insights-backend
yarn --cwd packages/app add @red-hat-developer-hub/backstage-plugin-analytics-module-adoption-insights
yarn --cwd packages/app add @@red-hat-developer-hub/backstage-plugin-adoption-insights
```

login based on EntraID, testing locally. 

I'm not yet to v1.35.1 that's why testing locally for the moment

karthikjeeyar:
@matdtr Can you please inspect the network calls in the browser to see if the requests are passing? specifically a request  `/event?type=total_users`

matdtr:
> [@matdtr](https://github.com/matdtr) Can you please inspect the network calls in the browser to see if the requests are passing? specifically a request `/event?type=total_users`

that specific one is failing with 500

karthikjeeyar:
Interesting only that call fails, I assume you have already added the backend [configuration](https://github.com/redhat-developer/rhdh-plugins/tree/main/workspaces/adoption-insights/plugins/adoption-insights-backend#configuration) in `app-config.yaml`

Have you configured any databases? or using default better-sqlite (in-memory) db?

matdtr:
Using on app-config.local.yaml

```  
analytics:
    adoptionInsights:
[...]

PR #608, opened 2025-04-05
fix(adoption-insights): fix subquery alias in total_users endpoint to support older PG versions

karthikjeeyar:
## Fixes:

https://github.com/redhat-developer/rhdh-plugins/issues/607     

## Problem:

 The failing API contains a sub query without an alias (alias was wrongly placed in the query), since sub-query alias was made optional in Postgres 16 onwards so it is not an issue with latest Postgres db versions, but it errors out in Postgres 15 or below. RHDH setup uses 
 
 This PR fixes database query failure for the earlier versions of Postgres. i.e Postgres 15 or below

**Before:**

before

---

**After:**

after


rohitkrai03 (approved):
/approve

PR #609, opened 2025-04-06
fix(orchestrator):Add option to view input schema from workflows table

LiorSoffer:
resolve [FLPATH-2248](https://issues.redhat.com/browse/FLPATH-2248)
Orchestrator page folder was created as well

image
image
image

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/InputSchemaDialog.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/InputSchemaDialog.tsx
@@ -1,0 +1,128 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/InputSchemaDialog.tsx
```diff
@@ -0,0 +1,121 @@
...
+      {inputSchema?.inputSchema === undefined ? (
+        'No input schema is defined for this workflow'
+      ) : (
+        <CodeSnippet
+          text={JSON.stringify(inputSchema, null, 2)}
+          language="json"
+          showLineNumbers
+          showCopyCodeButton
+          customStyle={{
+            color: activeThemeId === 'dark' ? '#abb2bf' : 'd3d3d3',
```
mareklibra, line 70: can we take these colors from the theme palette instead of hardcoding them?

LiorSoffer: These are the colors Shiran told me to use. I can get similar colors with theme.palette.grey[<some number>]
what do you think?

mareklibra: It's better to take it from the palette for both consistency and maintainability, especially if the colors are similar

PR #611, opened 2025-04-07
feat(orchestrator):  capturing the auth tokens from available SCM systems after user logsIn and then send it to backend-plugin

ijogi-redhat:


workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -104,10 +111,39 @@ export class OrchestratorClient implements OrchestratorApi {
     const defaultApi = await this.getDefaultAPI();
     const reqConfigOption: AxiosRequestConfig =
       await this.getDefaultReqConfig();
+    // We know the injected ScmIntegrationsApi instance is actually a ScmIntegrationRegistry with .list()
+    const integrations = (this.scmIntegrationsApi as any).list?.();
+    const authTokens: { provider: string; token: string }[] = [];
+    for (const integration of integrations) {
+      const provider = integration.type;
+      const host = integration.config.apiBaseUrl || integration.config.host;
```
gabriel-farache, line 119: The default `apiBaseUrl` value is `https://api.github.com/` and when it's using this URL to fetch the token, it fails.
Using the `host` alone make it works fine (so `https://github.com/`).

Is there a reason to use `apiBaseUrl`  instead of `host`?
Or is it a github related behaviour?
cc @mareklibra if you know anything about that

workspaces/orchestrator/.changeset/scm-token-propagation.md
```diff
@@ -0,0 +1,6 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-backend': major
+'@red-hat-developer-hub/backstage-plugin-orchestrator': major
+---
+
+[FLPATH-2224] : capturing the auth tokens from available SCM systems after user logsIn and then send it to backend-plugin.
```
gabriel-farache, line 6: I think we want to avoid using internal reference to jira in the upstream

gabriel-farache (commented):
Tested locally and works fine as long as `host` is used and not `apiBaseUrl` OR if `apiBaseUrl` is configured with same value as `host` in app config

Also, could you rename the PR title with something like
> feat(orchestrator): capturing the auth tokens from available SCM systems after user logsIn and then send it to backend-plugin

instead?

Same with the commits (and if you could squash them too :) )

good job!

workspaces/orchestrator/.changeset/scm-token-propagation.md
ijogi-redhat: Removed the internal reference to the Jira.

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -104,10 +111,38 @@ export class OrchestratorClient implements OrchestratorApi {
...
+        });
+
+        if (credentials?.token) {
+          authTokens.push({
+            provider,
+            token: credentials.token,
+          });
+        }
+      } catch (e) {
+        console.warn(`No token available for ${provider}`, e);
```
gciavarrini, line 134: no need to re-throw the error?

gciavarrini (changes requested):

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -104,10 +111,38 @@ export class OrchestratorClient implements OrchestratorApi {
     const defaultApi = await this.getDefaultAPI();
     const reqConfigOption: AxiosRequestConfig =
       await this.getDefaultReqConfig();
+    // We know the injected ScmIntegrationsApi instance is actually a ScmIntegrationRegistry with .list()
```
mareklibra, line 114: This seems to be risky. The injected `this.scmIntegrationsApi` is of `ScmIntegrationsApi` type. It contains just a single static factory function `fromConfig` which requires input parameter. So not the `.list()`.

How do we know that we are actually receiving a different object of `ScmIntegrationRegistry` type, already built one?
Btw. the `ScmIntegrationRegistry` is internal type only (not exported).

We need to be clear here and rely on Backstage supported API only. The risk of changes between versions is high.

mareklibra (commented):
This is a new area for me as well.

Wouldn't explicit iteration over all possible providers work as well/better? Their list is not long and changing very slowly. We can keep adding support as upstream adds it.

I mean to explicitly name those in the `plugins.ts`, sort of:
```
googleAuthApi: googleAuthApiRef,
```

instead of using the ScmIntegration API

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
ijogi-redhat: @mareklibra I am also new to this. As far as I have researched on using explicit auth apis like githubAuthApi, gitlabAuthApi, googleAuthApi, etc., forces the user to re-authenticate with a popup for each provider unless the access token is already cached and available. The backstage will only store identity token once the user logsin from signin page. So, this is the reason for me to use scmAuthApi.It attempts to retrieve the token silently (no popup)
and if the token is already present (i.e., from the sign-in session), it returns the same.

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
ijogi-redhat: @gciavarrini no need to re-throw the error here as the backend-plugin will itself throw 400 bad error if no auth tokens are present. We just want to silently log in the error.

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
ijogi-redhat: @mareklibra I have refactored the approach as suggested by you.Kindly take a look. Thanks !

workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -111,28 +114,24 @@ export class OrchestratorClient implements OrchestratorApi {
...
+      console.warn('GitHub token not available', e);
+    }
+    // Explicit GitLab token check
+    try {
+      const gitlabToken = await this.gitlabAuthApi?.getAccessToken?.();
+      if (gitlabToken) {
+        authTokens.push({ provider: 'gitlab', token: gitlabToken });
       }
+    } catch (e) {
+      console.warn('GitLab token not available', e);
```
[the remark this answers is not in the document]
ijogi-redhat: Makes sense. I have refactored the code accordingly and used Promise.allSettled instead of Promise.all.If GitLab is missing or throws, you still keep a valid GitHub token.We can log individual rejections without short-circuiting the whole batch.

gciavarrini (approved):
lgtm

mareklibra:
lgtm

PR #613, opened 2025-04-07
fix(orchestrator):Add status icons

LiorSoffer:
resolve [FLPATH-2253](https://issues.redhat.com/browse/FLPATH-2253)

image

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstanceStatusIndicator.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstanceStatusIndicator.tsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,7 +18,10 @@ import React from 'react';
 import { Link } from '@backstage/core-components';
 import { useRouteRef } from '@backstage/core-plugin-api';
 
-import DotIcon from '@material-ui/icons/FiberManualRecord';
+import { CircularProgress } from '@material-ui/core';
+import CheckCircleOutlinedIcon from '@material-ui/icons/CheckCircleOutlined';
+import ErrorOutlineOutlinedIcon from '@material-ui/icons/ErrorOutlineOutlined';
+import HourglassEmptyOutlinedIcon from '@material-ui/icons/HourglassEmptyOutlined';
 
 import {
   capitalize,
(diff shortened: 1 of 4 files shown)


workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstanceStatusIndicator.tsx
```diff
@@ -43,16 +47,50 @@ export const WorkflowInstanceStatusIndicator = ({
     return VALUE_UNAVAILABLE;
   }
 
+  let icon: React.ReactNode;
+  let title: string = '';
+  switch (status) {
+    case ProcessInstanceStatusDTO.Active:
+      icon = <CircularProgress size="1.15rem" />;
```
mareklibra, line 54:
```suggestion
      icon = <CircularProgress size="1.15rem"  className={iconColor} />;
```
Maybe there is no obvious change but it will make the code coherent.

mareklibra (changes requested):
Can you please check `WorkflowProgressNode.tsx` as well and fix it in case of significant inconsistencies?

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstanceStatusIndicator.tsx
LiorSoffer: it bothers me as well thanks :)

LiorSoffer:
@ShiranHi what do you think?

LiorSoffer:
@ShiranHi yes! , regarding the pending icon  I couldn't find the exact one on https://v4.mui.com/components/material-icons/, is there something else there which is more suitable?

LiorSoffer:
@ShiranHi please take a look

LiorSoffer:
@ShiranHi @mareklibra np, icon is fixed

ShiranHi:
> @ShiranHi , can we choose something from https://mui.com/material-ui/material-icons/ ? We should not mix icons from other sources.

Yes, we can use the HourglassEmpty icon

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowInstanceStatusIndicator.tsx
```diff
@@ -43,16 +46,50 @@ export const WorkflowInstanceStatusIndicator = ({
...
+    case ProcessInstanceStatusDTO.Active:
+      icon = <CircularProgress size="1.15rem" className={iconColor} />;
+      title = 'Running';
+      break;
+    case ProcessInstanceStatusDTO.Completed:
+      icon = <CheckCircleOutlinedIcon className={iconColor} />;
+      title = 'Completed';
+      break;
+    case ProcessInstanceStatusDTO.Suspended:
+      icon = <b className={iconColor}>--</b>;
```
mareklibra, line 61: The last one, I promise: We use 3 dashes on other places (`---`). I don't have any preference what to choose but we should align.

LiorSoffer: We use 3 dashes for unavailable value  , I think suspended is a different scenario than unavailable
@ShiranHi @mareklibra what do you think?

mareklibra: I would align it.

ShiranHi: Let's use the [WarningAmberOutlined](https://mui.com/material-ui/material-icons/?query=warning&theme=Outlined&selected=WarningAmberOutlined) icon for Unavailable status as [in the design](https://www.figma.com/design/XjdTCb4vDQ7L4Jz1TdHKGq/Workflow-Orchestrator?node-id=4687-1354&t=IytQq1dU2SyA5BrP-1).

LiorSoffer: sorry, I was unclear. there's not an Unavailable status,
Where are cases in the app when we are missing values for example (last run of workflow which was never executed, image added) in these cases we use 3 dashes.
image

ShiranHi: ah, got it. Thanks for the clarification! 
We use just two dashes (--) to indicate this state.

mareklibra (approved):

PR #626, opened 2025-04-08
fix(ai-experience): links should open pages in the same tab

karthikjeeyar:

This PR contains following changes:
- Fix the Link component import so the page transitions are smooth.
- removed "target=_blank" property from the UI to open links in the same page.

### DIFF
--- a/workspaces/ai-integrations/plugins/ai-experience/src/components/LearnSection/CardWrapper.tsx
+++ b/workspaces/ai-integrations/plugins/ai-experience/src/components/LearnSection/CardWrapper.tsx
@@ -18,8 +18,10 @@ import React from 'react';
 import Box from '@mui/material/Box';
 import CardContent from '@mui/material/CardContent';
 import Typography from '@mui/material/Typography';
-import Button from '@mui/material/Button';
 import ArrowForwardIcon from '@mui/icons-material/ArrowForward';
+import { Link } from '@backstage/core-components';
+import { makeStyles, useTheme } from '@material-ui/core';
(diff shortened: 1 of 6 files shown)


workspaces/ai-integrations/plugins/ai-experience/src/components/LearnSection/CardWrapper.tsx
```diff
@@ -28,12 +30,27 @@ interface CardWrapperProps {
   buttonLink: string;
 }
 
+const getStyles = makeStyles((theme: Theme) => ({
```
Eswaraiahsapram, line 33: Shall we use `sx prop` or `useTheme + Theme`. So we can use `@mui`.

karthikjeeyar: `sx` prop is not available in the Link component because it is now imported from `backstaeg/core-components`. More over the recommended approach for overriding styles is using makeStyles with theme.

Eswaraiahsapram (approved):
/lgtm

workspaces/ai-integrations/plugins/ai-experience/src/components/LearnSection/CardWrapper.tsx
karthikjeeyar: As discussed in slack with @Eswaraiahsapram , this Link component is needed for smooth page transitions and it doesn't contain `sx` prop.

rohitkrai03 (approved):
/approve

PR #633, opened 2025-04-10
fix(orchestrator): fix bug in workflow result -backport #632

LiorSoffer:


gciavarrini (approved):
Not a react expert but LGTM

mareklibra (approved):

PR #644, opened 2025-04-11
fix(ai-integrations): add null check for catalogKeys

johnmcollier:

This PR adds a null check for `catalogKeys` when we retrieve the key list from the catalog bridge. This ensures that we don't error out checking for properties on a null object.

I've also updated the workspace readme to include instructions for building the plugin archive


gabemontero (commented):
is a unit test onerous to add ?

johnmcollier:
> is a unit test onerous to add ?

Updated - just made it easy. Our entity provider tests are pretty light right now, but I tested this by verifying it doesn't error out, and that we never call `fetchModelCatalogFromKey`

gabemontero (approved):
/lgtm

gabemontero:
I'll hit the merge button since it is clean at the moment @johnmcollier

PR #448, opened 2025-02-26
fix(fab): fix fab position to append to root

debsmita1:
## Hey, I just made a Pull Request!

Resolves:

https://issues.redhat.com/browse/RHIDP-6042

Updated fab position so that it appends to `Backstage-Root`

**GIF:**

https://github.com/user-attachments/assets/0afa7f75-43ef-4375-9adc-15a3face9e0a

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ ] Added or Updated documentation
- [x] Tests for new functionality and regression tests for bug fixes
- [x] Screenshots attached (for UI changes)

PR #652, opened 2025-04-14
fix(fab): render fab at all times

debsmita1:
Part of: https://github.com/redhat-developer/rhdh-plugins/pull/448


workspaces/global-floating-action-button/.changeset/tidy-glasses-jump.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-floating-action-button': patch
+---
+
+render fab on document body when the target element is not found
```
logonoff, line 5: the changeset says that the fab should render on the document body, but the code would render it on the `main` element. 

is this correct?

debsmita1: ```document.querySelector('[class^="BackstagePage-root"]') or document.querySelector('main'); ```
decides the target element. If both of them returns null it should render on the document body. 

I overlooked removing this check https://github.com/redhat-developer/rhdh-plugins/pull/652/files#diff-84de97e74333e2bd4b6f22dfa470b675538294ffd8007b90d04ef134ae69c543L83 which prevented the fab button from rendering when the classname was not found

so, ACC to me the changeset is correct

logonoff (approved):

PR #686, opened 2025-04-23
feat(homepage): add onboarding, entities and templates section cards

Eswaraiahsapram:

Fixes - [RHDHPAI-771](https://issues.redhat.com/browse/RHDHPAI-771)

## Homepage
We've started work on new homepage cards in RHDH, such as the Onboarding Card, Entity Section Card, and Template Card, to provide users with better visibility and quicker access to important catalog content.

### What’s Included in This PR?
- Homepage UI components which matches [Figma](https://www.figma.com/design/6ORS1mhUaxeyTBjMmE95Rn/Home-page-redesign?node-id=1452-2329&t=LqoMyr8xXUDZmUVL-0)

1. Onboarding section 
> The onboarding Section Card is designed to help users quickly discover onboarding resources within RHDH. Positioned prominently on the homepage, it serves as a starting point for onboarding, exploring docs, and deepening knowledge of core RHDH features

2. Entity section card
> Entity Section Card is a UI component designed for the new RHDH homepage experience. It provides a visually engaging section intended to promote key catalog-driven entities or content.

3. Templates section card
> The Template Section Card enables users to quickly explore and initiate software templates in RHDH.

## Screenshots
image

image

image

image

## How to test
I used this command to install the plugin into my local rhdh:

```
cd rhdh-plugins/workspaces/homepage/plugins/dynamic-home-page
npx --yes @janus-idp/cli package export-dynamic-plugin --dynamic-plugins-root  <path-to-RHDH>/dynamic-plugins-root --dev
```

And this configuration in my `app-config.local.yaml`:
```yaml
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
  frontend:
    red-hat-developer-hub.backstage-plugin-dynamic-home-page:
      dynamicRoutes:
        - path: /
          importName: DynamicHomePage
      mountPoints:
        - mountPoint: home.page/cards
          importName: OnboardingSection
          config:
            layouts:
              xl: { w: 12, h: 5 }
              lg: { w: 12, h: 5 }
              md: { w: 12, h: 5 }
              sm: { w: 12, h: 5 }
              xs: { w: 12, h: 7 }
              xxs: { w: 12, h: 12 }
        - mountPoint: home.page/cards
          importName: EntitySection
          config:
            layouts:
              xl: { w: 12, h: 6 }
              lg: { w: 12, h: 6 }
              md: { w: 12, h: 6 }
              sm: { w: 12, h: 6 }
              xs: { w: 12, h: 6 }
              xxs: { w: 12, h: 14.5 }
        - mountPoint: home.page/cards
          importName: TemplateSection
          config:
            layouts:
              xl:  { w: 12, h: 5 }
              lg:  { w: 12, h: 5 }
              md:  { w: 12, h: 5 }
              sm:  { w: 12, h: 5 }
              xs:  { w: 12, h: 5 }
              xxs: { w: 12, h: 14 }
```


ShiranHi:
Looks great, thank you @Eswaraiahsapram !

I have a couple of questions:
1. I believe we talked about removing the page header with the "Welcome back" text, since we already have a welcome message right below it ("Good afternoon, <name>"). Can we make that change?
2. Would it be possible to update the name on the first card to show only the first name?
3. The buttons on the first card ("Go to Tech Docs" and "Go to Catalog") seem to have quite a bit of padding, can we tighten that up?
image0

benwilcock:
One thing I noticed, that I think is different from the mockups, is that the users full forename and surname is used in the "Good Afternoon XXX" section. I preferred firstname only as we already have the users full name in the Global Header.

Eswaraiahsapram:
> Looks great, thank you @Eswaraiahsapram !
> 
> I have a couple of questions:
> 
> 1. I believe we talked about removing the page header with the "Welcome back" text, since we already have a welcome message right below it ("Good afternoon, "). Can we make that change?
> 2. Would it be possible to update the name on the first card to show only the first name?
> 3. The buttons on the first card ("Go to Tech Docs" and "Go to Catalog") seem to have quite a bit of padding, can we tighten that up?
> 
> image0

@ShiranHi , thank you for reviewing. 

1. "Welcome back" is from Header component. This learn section is a Card where dynamically we can add it. cc- @rohitkrai03 @karthikjeeyar 
2. Done (please refer latest images)
3. Done

workspaces/homepage/plugins/dynamic-home-page/src/api/ModelCatalogApiClient.ts
```diff
@@ -0,0 +1,91 @@
...
+import {
+  CatalogApi,
+  EntityFilterQuery,
+  GetEntitiesRequest,
+  GetEntityFacetsRequest,
+} from '@backstage/catalog-client';
+import type { Entity } from '@backstage/catalog-model';
+
+export interface AIModel extends Entity {
+  spec: any;
```
karthikjeeyar, line 26: please add `type` here

workspaces/homepage/plugins/dynamic-home-page/src/api/ModelCatalogApiClient.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+}
+
+export interface ModelCatalogApi {
+  getModels(
+    request: GetEntitiesRequest,
+  ): Promise<QueryEntitiesResponse<AIModel>>;
+  getTemplates(
+    request: GetEntitiesRequest,
+  ): Promise<QueryEntitiesResponse<any>>;
```
karthikjeeyar, line 44: Add type for template and use it in the API implementation as well.

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useTemplates.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { modelCatalogApiRef } from '../api';
+
+export const useTemplates = () => {
+  const [isLoading, setIsLoading] = React.useState<boolean>(true);
+  const [data, setData] = React.useState<any>();
+
+  const modelServiceApi = useApi(modelCatalogApiRef);
+
+  const getModelCatalogs = React.useCallback(async () => {
```
karthikjeeyar, line 29:
```suggestion
  const getTemplates = React.useCallback(async () => {
```

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useTemplates.tsx
```diff
@@ -0,0 +1,50 @@
...
+    return await modelServiceApi
+      .getTemplates({ limit: 4 })
+      .then(response => setData(response));
+  }, [modelServiceApi]);
+
+  const { error, loading } = useAsyncRetry(async () => {
+    return await getModelCatalogs();
```
karthikjeeyar, line 36:
```suggestion
    return await getTemplates();
```

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useModels.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react';
+
+import { useApi } from '@backstage/core-plugin-api';
+import useAsyncRetry from 'react-use/lib/useAsyncRetry';
+
+import { modelCatalogApiRef } from '../api';
+
+export const useModels = () => {
+  const [isLoading, setIsLoading] = React.useState<boolean>(true);
+  const [data, setData] = React.useState<any>();
```
karthikjeeyar, line 25:
```suggestion
  const [data, setData] = React.useState<QueryEntitiesResponse<AIModel>>();
```

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useTemplates.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react';
+
+import { useApi } from '@backstage/core-plugin-api';
+import useAsyncRetry from 'react-use/lib/useAsyncRetry';
+
+import { modelCatalogApiRef } from '../api';
+
+export const useTemplates = () => {
+  const [isLoading, setIsLoading] = React.useState<boolean>(true);
+  const [data, setData] = React.useState<any>();
```
karthikjeeyar, line 25: please add the newly added template type here as well.
```suggestion
  const [data, setData] = React.useState<QueryEntitiesResponse<Template>>();
```

karthikjeeyar (commented):
Here are some Questions/observations while testing this locally:

1. when there are no AI models registered yet and if the user closes the explore card then there is a big gap between the cards, can we have a empty state in this case?

https://github.com/user-attachments/assets/b5940171-e23c-4afa-819b-18b88ba05bdb


2. The heading of this section reads "Explore your software catalog", I would expect to see all the items in my catalog `eg: API, Component, Resource etc` in this section, but right now it is only fetching AI model resources, is this intentional?
3. IMO, we should have different cards one for all catalog items for RHDH usecase (Explore your Software catalog) and one more for AI homepage "Explore your AI models" so we could use easily switch between general catalog section vs ai model catalog based on some configuration later. 

4. Sections isn't responsive when running it in RHDH

image

karthikjeeyar:
Thanks @ShiranHi for clarification.

> We’re splitting which card is shown based on whether customers have purchased the AI flavor of RHDH. If they have, they’ll see AI content, if not they’ll see catalog entities as described in point 2. If users haven’t purchased the AI flavor, there’s no need to display the "Explore your AI models" card.


yeah we could have a single card that will react based on the AI flavour (prop or config value) to show AI content or else by default it will show all the catalog entities. @Eswaraiahsapram  I think this is not implemented yet in this PR so Is this something we will have in this PR or will be a followup PR?

rohitkrai03:
> Thanks @ShiranHi for clarification.
> 
> > We’re splitting which card is shown based on whether customers have purchased the AI flavor of RHDH. If they have, they’ll see AI content, if not they’ll see catalog entities as described in point 2. If users haven’t purchased the AI flavor, there’s no need to display the "Explore your AI models" card.
> 
> yeah we could have a single card that will react based on the AI flavour (prop or config value) to show AI content or else by default it will show all the catalog entities. @Eswaraiahsapram I think this is not implemented yet in this PR so Is this something you are planning to add in this PR or will be a followup PR?

@karthikjeeyar The homepage plugin should eventually be able to react based on flavors but that's not planned for this release. For now this PR or story should only focus on bringing in the new design laungage for homepage to the core homepage. We don't need to replace the AI homepage work that we did with this plugin as of now.

karthikjeeyar:
>The homepage plugin should eventually be able to react based on flavors but that's not planned for this release. For now this PR or story should only focus on bringing in the new design laungage for homepage to the core homepage. We don't need to replace the AI homepage work that we did with this plugin as of now.

Thanks @rohitkrai03, that clears my doubt. So the API should be just returning all the `kinds` from the catalog for eg: `API`, `Resource` and `Component` and not just `Resource` kind, right?  API is currently named as modelCatalogApi and one of the methods is  `getModels`, instead we should have `getCatalogEntities` to return all the items in catalog.

Eswaraiahsapram:
> Thanks @rohitkrai03, that clears my doubt. So the API should be just returning all the `kinds` from the catalog for eg: `API`, `Resource` and `Component` and not just `Resource` kind, right? API is currently named as modelCatalogApi (seems like ai-focussed API) and one of the methods is `getModels`, we should also have `getCatalogEntities` to return all the items in catalog.

@karthikjeeyar if we fetch all kinds we end up templates as well. As Rohit suggestion I removed only spec.type as of now.

workspaces/homepage/plugins/dynamic-home-page/src/components/ModelSection/ModelSection.tsx
```diff
@@ -0,0 +1,270 @@
...
+      setIsRemoveFirstCard(true);
+    }, 500);
+  };
+
+  const { data, error, isLoading } = useModels();
+  const models = data?.items ?? [];
+
+  const params = new URLSearchParams({
+    'filters[kind]': 'resource',
+    'filters[type]': 'ai-model',
```
karthikjeeyar, line 92:
```suggestion

```

karthikjeeyar (changes requested):
@Eswaraiahsapram Since this components, APIs are serving core homepage usecase, I dont think we should name anything related to AI here, please rename the Components, APIs, type etc where-ever you are now using keyword `AI` or `Model`

workspaces/homepage/plugins/dynamic-home-page/src/api/ModelCatalogApiClient.ts
```diff
@@ -0,0 +1,86 @@
...
+      ...filter,
+    },
+  });
+
+  async getModels(
+    request: GetEntitiesRequest,
+  ): Promise<QueryEntitiesResponse<Entity>> {
+    const response = await this.catalogApi.queryEntities(
+      this.appendFilter(request, {
+        kind: 'Resource',
```
karthikjeeyar, line 70: This will only return `Resource` entities but "Explore your software catalog" section talks about browsing Components, Resources, Systems and APIs in your organisation.

Can you please make this an array with other kinds as well ?

```suggestion
        kind: [ 'Component', 'API', 'Resource', 'System'],
```

rohitkrai03 (changes requested):

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,268 @@
...
+            )}
+            {entities?.slice(0, isRemoveFirstCard ? 4 : 2).map((item: any) => (
+              <Grid
+                item
+                xs={12}
+                md={isRemoveFirstCard ? 3 : 3.5}
+                key={item.metadata.name}
+              >
+                <CardWrapper
+                  link={`/catalog/default/resource/${item.metadata.name}`}
```
karthikjeeyar, line 182: `kind` is hardcoded and it redirects you to 404 page.

```suggestion
                  link={`/catalog/default/${item.kind}/${item.metadata.name}`}
```

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/CardWrapper.tsx
```diff
@@ -0,0 +1,101 @@
...
+              WebkitBoxOrient: 'vertical',
+              WebkitLineClamp: 8,
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+            }}
+          >
+            {description}
+          </Typography>
+        </Box>
+        <TagList tags={tags} />
```
karthikjeeyar, line 95: In the [design](https://www.figma.com/design/6ORS1mhUaxeyTBjMmE95Rn/Home-page-redesign?node-id=1452-1757&t=wigM7YCFmStp7pcM-0), every card starts with coloured badge denoting the `kind` of the entity.

Design:
image


 Can we show the `kind` as the first badge and then followed by tags (if available)? Currently we can't say which type of entity is showing up.

image

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,269 @@
...
+            )}
+            {entities?.slice(0, isRemoveFirstCard ? 4 : 2).map((item: any) => (
+              <Grid
+                item
+                xs={12}
+                md={isRemoveFirstCard ? 3 : 3.5}
+                key={item.metadata.name}
+              >
+                <EntityCard
+                  link={`/catalog/default/${item.kind}/${item.metadata.name}`}
```
karthikjeeyar, line 182: nit: However this link works, still it is good to be consistent with url's having lowercase string.
```suggestion
                  link={`/catalog/default/${item.kind.toLocaleLowerCase('en-US')}/${item.metadata.name}`}
```

karthikjeeyar (commented):
Thanks @Eswaraiahsapram for the followup. Verified locally works fine. 


Could you please add tests wherever it is possible?

karthikjeeyar (approved):
Thanks @Eswaraiahsapram for the adding tests and followup. 

/lgtm

karthikjeeyar:
cc: @christoph-jerolimov

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
```diff
@@ -0,0 +1,98 @@
...
+        border: theme => `1px solid ${theme.palette.grey[300]}`,
+        borderRadius: 3,
+      }}
+    >
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          fontSize: '1.5rem',
+          fontWeight: '500',
```
christoph-jerolimov, line 52: Can we use a h2 or a Typography tag here?

This is visually a headline, then we should also use a markup that makes it more a11y.

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateSection.tsx
```diff
@@ -0,0 +1,185 @@
...
+        padding: '24px 24px 0 24px',
+        border: muiTheme => `1px solid ${muiTheme.palette.grey[300]}`,
+        borderRadius: 3,
+      }}
+    >
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          fontSize: '1.5rem',
```
christoph-jerolimov, line 176: Can we use a h2 or a Typography tag here?

This is visually a headline, then we should also use a markup that makes it more a11y.

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,269 @@
...
+        padding: '24px 24px 0 24px',
+        border: muitheme => `1px solid ${muitheme.palette.grey[300]}`,
+        borderRadius: 3,
+      }}
+    >
+      <div
+        style={{
+          display: 'flex',
+          alignItems: 'center',
+          fontSize: '1.5rem',
```
christoph-jerolimov, line 260: Can we use a h2 or a Typography tag here?

This is visually a headline, then we should also use a markup that makes it more a11y.

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateSection.tsx
```diff
@@ -0,0 +1,185 @@
...
+          fontWeight: '500',
+        }}
+      >
+        Explore Templates
```
christoph-jerolimov, line 180: @ShiranHi Should we go with "Explore Templates" or something like "Explore Self-service" or "Explore Self-service catalog" or "Explore Self-service options"?

Because we renamed the sidebar item and page title to "Self-service"

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/TagList.tsx
```diff
@@ -0,0 +1,94 @@
...
+      <Chip
+        key={kind}
+        label={
+          <Typography sx={{ fontSize: '0.8rem', fontWeight: 400 }}>
+            {kind}
+          </Typography>
+        }
+        sx={{
+          backgroundColor: KINDS[kind.toLocaleUpperCase() as KindKeys]?.fill,
+        }}
```
christoph-jerolimov, line 52: Can you set also a (text) color when we use really specific background color here. Maybe simply black?

It looks okaysh on the light theme

Screenshot From 2025-05-07 11-09-12

but on dark theme its difficult to read

Screenshot From 2025-05-07 11-09-03

workspaces/homepage/plugins/dynamic-home-page/src/utils/constants.ts
```diff
@@ -0,0 +1,58 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+export const LEARNING_SECTION_ITEMS = [
+  {
+    title: 'Get started',
+    description: 'Learn about Red Hat Developer Hub.',
+    buttonText: 'Go to Tech Docs',
+    buttonLink:
+      'https://docs.redhat.com/en/documentation/red_hat_developer_hub/',
```
christoph-jerolimov, line 22: @ShiranHi Go to Tech Docs confused me here because its not a "tech docs" its our RHDH Documentation. But its fine if we want stay with that.

christoph-jerolimov (changes requested):
Hi @Eswaraiahsapram @karthikjeeyar 

I added a bunch of comments.

It's not required to make the boxes more flexibel right now. But imo it would be great if they are better responsible.

Which means: If the configuration uses h: 5, the space should be used at least for the "Card border". But its okay for the first version.

I suggested also some cleanups that should result in less code...

And finally I want bring up the issue that the learning path (plugin) could be disabled. The card shows a hard coded (broken) link in that case. We should make that more flexibel before we enable this card in a customer setup by default.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { LEARNING_SECTION_ITEMS } from '../../utils/constants';
+import useGreeting from '../../hooks/useGreeting';
+
+export const OnboardingSection = () => {
+  const greeting = useGreeting();
+  const identityApi = useApi(identityApiRef);
+
+  const { value: profile } = useAsync(() => identityApi.getProfileInfo());
+  const fullName = profile?.displayName ?? '';
+  const firstName = fullName.split(' ')[0];
```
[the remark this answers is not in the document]
christoph-jerolimov: cc @karthikjeeyar @debsmita1 @ciiay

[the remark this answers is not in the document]
karthikjeeyar: Good to know 👍🏻

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateSection.tsx
ShiranHi: "Explore Templates" should work as it is easier to understand, especially for new users. @benwilcock what do you think?

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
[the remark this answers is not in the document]
christoph-jerolimov: We're still evaluating the exact solution. Looks like some configurations result in a good user profile, some other in a good user entity, but neither alone is reliable. We didn't decided yet what we do in the header.

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+  kind,
+}) => {
+  return (
+    <Box
+      sx={{
+        border: theme => `1px solid ${theme.palette.grey[400]}`,
+        borderRadius: 3,
+        overflow: 'hidden',
+      }}
+    >
```
christoph-jerolimov, line 49: Please use a Card instead. You can remove the custom borderRadius since our latest theme added this bigger border radius already. If a customer uses their own theme, there is no reason that the home page cards enforces that.

Can we also make the content scrollable instead of saying "overflow hidden" when the card is too small?

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+  kind,
+}) => {
+  return (
+    <Box
+      sx={{
+        border: theme => `1px solid ${theme.palette.grey[400]}`,
+        borderRadius: 3,
+        overflow: 'hidden',
+      }}
+    >
```
christoph-jerolimov, line 44: Please use a Card instead. You can remove the custom borderRadius since our latest theme added this bigger border radius already. If a customer uses their own theme, there is no reason that the home page cards enforces that.

Can we also make the content scrollable instead of saying "overflow hidden" when the card is too small?

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
```diff
@@ -0,0 +1,97 @@
...
+              overflow: 'hidden',
+              textOverflow: 'ellipsis',
+              fontSize: '0.875rem',
+              fontWeight: '500',
+            }}
+          >
+            {title}
+          </Link>
+        </Box>
+        <Box sx={{ pt: 2, height: '175px', overflow: 'hidden' }}>
```
christoph-jerolimov, line 76: Can't we use a more flexible height on the cards?

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const isExtraSmallScreen = useMediaQuery(theme.breakpoints.down(900));
+  let smallScreenWidth;
+  if (isExtraSmallScreen) {
+    smallScreenWidth = 266;
+  } else if (isSmallScreen) {
+    smallScreenWidth = 115;
+  } else {
+    smallScreenWidth = 180;
+  }
+  const imageWidth = isMediumScreen ? smallScreenWidth : 266;
```
christoph-jerolimov, line 67: Is there no more responsive way to this?

I'm not saying css clamp needs to be used, but might a good fit, wdyt?

Maybe @logonoff can help with other ideas? :)

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,261 @@
...
+  };
+
+  const { data, error, isLoading } = useEntities();
+
+  const entities = data?.items ?? [];
+
+  const params = new URLSearchParams({
+    limit: '20',
+  });
+  const catalogEntityLink = `/catalog?${params.toString()}`;
```
christoph-jerolimov, line 83: No reason to enforce a limit here or?

It's simpler code and still working fine if you link below to '/catalog'.

Please remove this:

```suggestion
```

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,261 @@
...
+
+  return (
+    <Box
+      component={Paper}
+      sx={{
+        padding: '24px 24px 0 24px',
+        border: muitheme => `1px solid ${muitheme.palette.grey[300]}`,
+        borderRadius: 3,
+      }}
+    >
```
christoph-jerolimov, line 246: Does it looks the same if you use a Card here?

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+  return (
+    <Box
+      component={Paper}
+      sx={{
+        padding: '24px',
+        border: muiTheme => `1px solid ${muiTheme.palette.grey[300]}`,
+        borderRadius: 3,
+      }}
+    >
```
christoph-jerolimov, line 53: Does it look the same if you use a Card here?

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useTemplates.ts
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useApi } from '@backstage/core-plugin-api';
+import { catalogApiRef } from '@backstage/plugin-catalog-react';
+
+import useAsync from 'react-use/lib/useAsync';
+
+export const useTemplates = () => {
+  const entityServiceApi = useApi(catalogApiRef);
```
christoph-jerolimov, line 22: keep it simple:

```suggestion
  const catalogApi = useApi(catalogApiRef);
```

workspaces/homepage/plugins/dynamic-home-page/src/hooks/useEntities.ts
```diff
@@ -0,0 +1,38 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import { useApi } from '@backstage/core-plugin-api';
+import { catalogApiRef } from '@backstage/plugin-catalog-react';
+
+import useAsync from 'react-use/lib/useAsync';
+
+export const useEntities = () => {
+  const entityServiceApi = useApi(catalogApiRef);
```
christoph-jerolimov, line 22: keep it simple

```suggestion
  const catalogApi = useApi(catalogApiRef);
```

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateSection.tsx
```diff
@@ -0,0 +1,169 @@
...
+    );
+  } else {
+    content = (
+      <Box sx={{ padding: '16px 10px 10px 0' }}>
+        <React.Fragment>
+          <Grid container spacing={1} alignItems="stretch">
+            {templates?.items.map((item: any) => (
+              <Grid item xs={12} md={3} key={item.title}>
+                <TemplateCard
+                  link={`/create/templates/default/${item.metadata.name}`}
```
christoph-jerolimov, line 86: This link works only when the template uses the default namespace. Please take care of `item.metadata.namespace`.

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntitySection.tsx
```diff
@@ -0,0 +1,261 @@
...
+            )}
+            {entities?.slice(0, isRemoveFirstCard ? 4 : 2).map((item: any) => (
+              <Grid
+                item
+                xs={12}
+                md={isRemoveFirstCard ? 3 : 3.5}
+                key={item.metadata.name}
+              >
+                <EntityCard
+                  link={`/catalog/default/${item.kind.toLocaleLowerCase('en-US')}/${item.metadata.name}`}
```
christoph-jerolimov, line 179: This link also ignores that the item could be in another namespace then `default`.

Maybe its even better to pass the entity into your `EntityCard` and use backstage `EntityRefLink` to create the link around the card (or in the card).

christoph-jerolimov (changes requested):

karthikjeeyar:
I see scrollbars are now added in cards, but can we also add section to be scrollable as well? so that even when user sets a really small height (h: 3) the section should be still viewable via scroll (now it is hidden).

image

karthikjeeyar:
Verified locally in RHDH with latest theme (`"@red-hat-developer-hub/backstage-plugin-theme": "0.8.1"`), looks good to me


image

/lgtm

workspaces/homepage/plugins/dynamic-home-page/src/utils/constants.ts
benwilcock: The idea is to have it go to TechDocs. I've asked our docs folks to produce some limited end-user focused documentation in TechDocs format and this button would take you to it. https://issues.redhat.com/browse/RHDHPAI-602

There's a demo on the summit cluster of how it should really work using these techdocs (not official) https://github.com/benwilcock/rhdh-techdocs/blob/main/rhdh-catalog-info.yaml

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
```diff
@@ -0,0 +1,109 @@
...
+import HomePageIllustration from '../../images/homepage-illustration-1.svg';
+import { LEARNING_SECTION_ITEMS } from '../../utils/constants';
+import useGreeting from '../../hooks/useGreeting';
+
+export const OnboardingSection = () => {
+  const greeting = useGreeting();
+  const { displayName } = useUserProfile();
+
+  const fullName = displayName ?? '';
+  const firstName = fullName.split(' ')[0];
```
logonoff, line 35: i don't think it's safe to assume that a first name will be one word:
- e.g.,: https://academia.stackexchange.com/questions/78746/how-to-get-others-to-understand-that-my-first-name-consists-of-two-words
- some chinese names, e.g.,: https://en.wikipedia.org/wiki/Loh_Kean_Yew (the first name is `Kean Yew`)

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+            pb: 2,
+          },
+          backgroundColor: 'transparent',
+        }}
+      >
+        <Box sx={{ overflow: 'hidden' }}>
+          <EntityRefLink
+            entityRef={entity}
+            style={{
+              display: '-webkit-box',
```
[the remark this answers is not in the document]
Eswaraiahsapram: Thanks @logonoff for the suggestion. I checked the MDN docs for the non-vendored `line-clamp` property. Unfortunately, browser support for the standard `line-clamp` is still limited - especially in Firefox and some chromium browsers.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
Eswaraiahsapram: @ShiranHi, could you please take a look and suggest how we might better handle this case?

karthikjeeyar:
/lgtm cancel

I just noticed this comment from Christoph is still not resolved.

> And finally I want bring up the issue that the learning path (plugin) could be disabled. The card shows a hard coded (broken) link in that case. We should make that more flexibel before we enable this card in a customer setup by default.

If there is a way to identify if the learning path is enabled or not, we should use it conditionally render that "Learning Path" card.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
ShiranHi: @Eswaraiahsapram we have the same discussion [here](https://github.com/redhat-developer/rhdh-plugins/pull/781#discussion_r2089097559). We can use the full name. @benwilcock Let me know if you have any concerns about that.

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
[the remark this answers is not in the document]
logonoff: it appears integer values such as `line-clamp: 3;` have full browser support but the keyword `line-clamp: none;` doesn't. can you double check?

karthikjeeyar (approved):
Thanks @Eswaraiahsapram, I can now see the full name and rounded buttons in the onBoarding cards.

image
/lgtm

invincibleJai:
Verified the changes in RHDH with plugin dist and provided config, can see below

image

invincibleJai:
@logonoff would you like to take another look, or it has +1 from your side?

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+  title,
+  description,
+  buttonText,
+  buttonLink,
+  target,
+}) => {
+  return (
+    <Box>
+      <CardContent sx={{ backgroundColor: 'transparent' }}>
+        <Box
```
logonoff, line 42: can we use a MUI typography variant here?

logonoff (commented):
overall looks great! just have a few final comments

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+          >
+            {title}
+          </EntityRefLink>
+        </Box>
+        <Box
+          sx={{
+            pt: 2,
+            height: '175px',
+            overflow: 'auto',
+            '$::-webkit-scrollbar': {
```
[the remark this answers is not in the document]
Eswaraiahsapram: I think it's better to completely hide the scrollbar to maintain a cleaner and more consistent UI. @ShiranHi @rohitkrai03 @karthikjeeyar 
image

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
```diff
@@ -0,0 +1,100 @@
...
+              display: 'none',
+            },
+            scrollbarWidth: 'none',
+          }}
+        >
+          <Typography variant="body2" paragraph>
+            {description}
```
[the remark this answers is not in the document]
Eswaraiahsapram: Good question - this is currently just a plain description field, and we’re not expecting markdown content here. It’s mainly for displaying a short summary about the entity.

@ShiranHi / @rohitkrai03  / @karthikjeeyar - could you please suggest whether we should support markdown rendering for the entity description?

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+            overflow: 'hidden',
+            textOverflow: 'ellipsis',
+          }}
+        >
+          {description}
+        </Box>
+        <Button
+          component={RouterLink}
+          variant="outlined"
+          color="primary"
```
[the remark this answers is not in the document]
Eswaraiahsapram: From my end its working fine

image

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
```diff
@@ -0,0 +1,106 @@
...
+      <Typography
+        variant="h3"
+        sx={{
+          display: 'flex',
+          alignItems: 'center',
+          fontWeight: '500',
+          fontSize: '1.5rem',
+        }}
+      >
+        {`${greeting} ${displayName ?? 'Guest'}!`}
```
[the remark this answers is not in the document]
Eswaraiahsapram: @logonoff , I think this is coming from global header (Welcome back). It was going to replace with new header where we don't have that greetings from global header.

@ShiranHi @karthikjeeyar

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+            margin: 0,
+          }}
+        >
+          {title}
+        </Box>
+        <Box
+          component="div"
+          sx={{
+            fontSize: '1rem',
+            fontWeight: 500,
```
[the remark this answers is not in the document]
Eswaraiahsapram: @logonoff - Good question. IMO this is more over subtitle, where it saying `Learn..., Explore...`. Better this text we can keep it semibolded. This is [Figma](https://www.figma.com/design/6ORS1mhUaxeyTBjMmE95Rn/Home-page-redesign?node-id=1452-2289&m=dev) we have same design.

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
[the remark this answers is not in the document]
ShiranHi: Yes, removing the scrollbar definitely looks better. Could we also add an ellipsis ("...") at the end of the text to visually indicate that there's more content available (just like in [the design](https://www.figma.com/design/6ORS1mhUaxeyTBjMmE95Rn/Home-page-redesign?node-id=1452-1757&t=j9n64QkgQ839VQDl-1))?

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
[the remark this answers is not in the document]
ShiranHi: Good catch! Yes, having text formatting to improve readability would be really helpful.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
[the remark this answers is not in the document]
ShiranHi: We removed the home page header to prevent this redundant welcome message ([design](https://www.figma.com/design/6ORS1mhUaxeyTBjMmE95Rn/Home-page-redesign?node-id=1452-1757&t=dVrAF9elPw9WWHF4-1)). Can we hide it in case users use this new home page?

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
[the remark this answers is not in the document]
Eswaraiahsapram: @ShiranHi , since we’ve enabled scroll for longer content, applying an ellipsis (...) isn't feasible in this case. ellipsis is typically used when text is visually truncated, not when scroll is available.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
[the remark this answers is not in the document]
ShiranHi: ok, thank you for checking. @benwilcock FYI

workspaces/homepage/plugins/dynamic-home-page/src/components/EntitySection/EntityCard.tsx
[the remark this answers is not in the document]
ShiranHi: Right, I suggest removing the scroll bat and adding ellipsis instead

workspaces/homepage/plugins/dynamic-home-page/src/components/TemplateSection/TemplateSection.tsx
benwilcock: Let's go with templates. We can always change it if feedback suggests it's too hard to follow.

benwilcock:
BTW, I don't think it should show full name, only first name - because we already have full name in the global header, so it's repetitive.

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingSection.tsx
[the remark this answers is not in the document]
benwilcock: Where it says "Good morning Guest" I think <guest> should be the users first name, not their full name (as this is already in the global header drop down).

Eswaraiahsapram:
> BTW, I don't think it should show full name, only first name - because we already have full name in the global header, so it's repetitive.

@benwilcock FYI - https://github.com/redhat-developer/rhdh-plugins/pull/686#discussion_r2089099865

As per the discussion above, the priority for displaying the name should be `spec.profile.displayname`, followed by  `displayName`. Currently, we don;t have a reliable setup to extract the first name(please refer above link)

workspaces/homepage/plugins/dynamic-home-page/src/components/OnboardingSection/OnboardingCard.tsx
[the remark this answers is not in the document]
Eswaraiahsapram: @logonoff please reopen if this is not matching with the Figma or any questions on this. Thank you

workspaces/homepage/plugins/dynamic-home-page/src/images/homepage-illustration-1.svg
logonoff, line 1: is this file still used?

logonoff (commented):
can you double check if this a bug? or did I configure RHDH incorrectly

https://github.com/user-attachments/assets/e8d89273-b2e0-4ea4-b92a-32f2914bfded

Eswaraiahsapram:
> can you double check if this a bug? or did I configure RHDH incorrectly
> 
>  Screencast.From.2025-05-20.11-28-21.mp4

Hi @logonoff , @ShiranHi could you please have a look. Regards Software catalog Info card (When user visits the page we are showing in all visits)

ShiranHi:
> can you double check if this a bug? or did I configure RHDH incorrectly

Yes, that's correct. Once users click to close this card, it should not reappear.

logonoff (approved):
/lgtm

Eswaraiahsapram:
> Yes, that's correct. Once users click to close this card, it should not reappear.

@ShiranHi / @logonoff  , Thanks for the comment! I've updated the logic to store the dismissal state in `sessionStorage`, so the card will remain hidden only until the browser session ends. Once the user closes and reopens the browser (starting a new session), the card will be shown again. This aligns with the requirement that the card should not reappear until the session ends.

Let me know if you'd prefer to persist this state longer (e.g., using `localStorage` or an expiry timestamp).

benwilcock:
I think it should be longer than the session.

ShiranHi:
> I think it should be longer than the session.

I agree, can we change that?

invincibleJai (approved):
/approve

rohitkrai03 (approved):
/approve

PR #698, opened 2025-04-28
feat(ai-integrations): Provide support for techdocs when present on the Model Catalog JSON object

johnmcollier:

This PR updates the model catalog plugin to add the techdocs annotation (`backstage.io/techdocs-ref`) to the generated `ai-model` Resource entities when they are generated.

### DIFF
--- a/workspaces/ai-integrations/plugins/catalog-backend-module-model-catalog/src/clients/ModelCatalogGenerator.ts
+++ b/workspaces/ai-integrations/plugins/catalog-backend-module-model-catalog/src/clients/ModelCatalogGenerator.ts
@@ -122,6 +122,20 @@ export function GenerateModelResourceEntities(
         `component:${modelServer.name}`,
       );
     }
+
+    // Handle any annotations present on the model resource
+    if (model.annotations !== undefined) {
+      // Add the techdocs annotation to the resource if present
+      if (model.annotations.TechDocs !== '') {
+        const techdocsUrl: string = model.annotations.TechDocs;
+        if (modelResourceEntity.metadata.annotations === undefined) {
+          modelResourceEntity.metadata.annotations = {};
+        }
+        modelResourceEntity.metadata.annotations[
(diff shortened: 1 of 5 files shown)


workspaces/ai-integrations/plugins/catalog-backend-module-model-catalog/src/clients/ModelCatalogGenerator.ts
```diff
@@ -122,6 +122,20 @@ export function GenerateModelResourceEntities(
         `component:${modelServer.name}`,
       );
     }
+
+    // Handle any annotations present on the model resource
+    if (model.annotations !== undefined) {
+      // Add the techdocs annotation to the resource if present
+      if (model.annotations.TechDocs !== '') {
```
gabemontero, line 129: should we trim for spaces and then check for the empty string

gabemontero: also, should we enforce that the annotation is a http/https URL

gabemontero (commented):
just a couple of suggestions

workspaces/ai-integrations/plugins/catalog-backend-module-model-catalog/src/clients/ModelCatalogGenerator.ts
johnmcollier: > should we trim for spaces and then check for the empty string

Yes, we should do that given the free form nature of the input textbox in RHOAI

> also, should we enforce that the annotation is a http/https URL

IMO, this should be handled in the bridge if we do this.

gabemontero: ok then ... I'm working on PR to make the polling interval / push to backstage configurable ... I'll include it there

gabemontero (approved):

PR #700, opened 2025-04-29
fix(orchestrator): Add workflow status

LiorSoffer:
resolve [FLPATH-2301](https://issues.redhat.com/browse/FLPATH-2301)

**Available** 

Screenshot from 2025-04-29 13-40-52

Screenshot from 2025-04-29 13-41-01

Screenshot from 2025-04-29 13-41-17

**Unavailable**

Screenshot from 2025-04-29 13-25-54
Screenshot from 2025-04-29 13-26-36
Screenshot from 2025-04-29 13-39-39

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
@@ -121,13 +121,9 @@ export class OrchestratorService {
     instanceId: string;
     cacheHandler?: CacheHandler;
   }): Promise<ProcessInstance | undefined> {
-    const { instanceId, cacheHandler } = args;
+    const { instanceId } = args;
     const instance = await this.dataIndexService.fetchInstance(instanceId);
-    const isWorkflowAvailable = this.workflowCacheService.isAvailable(
-      instance?.processId,
-      cacheHandler,
-    );
-    return isWorkflowAvailable ? instance : undefined;
+    return instance;
   }
 
   // SonataFlow Service Wrapper
@@ -218,15 +214,13 @@ export class OrchestratorService {
 
   public async fetchWorkflowOverview(args: {
     definitionId: string;
-    cacheHandler?: CacheHandler;
   }): Promise<WorkflowOverview | undefined> {
-    const { definitionId, cacheHandler } = args;
-    const isWorkflowAvailable = this.workflowCacheService.isAvailable(
-      definitionId,
-      cacheHandler,
-    );
-    return isWorkflowAvailable
-      ? await this.sonataFlowService.fetchWorkflowOverview(definitionId)
-      : undefined;
+    const { definitionId } = args;
+    const isWorkflowAvailable =
+      this.workflowCacheService.isAvailable(definitionId);
+    const overview =
+      await this.sonataFlowService.fetchWorkflowOverview(definitionId);
+    if (overview) overview.isAvailable = isWorkflowAvailable; // workflow overview is avaiable but the workflow itself is not
+    return overview;
   }
 }
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/RunButton.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/RunButton.tsx
@@ -30,7 +30,7 @@ import {
 import { usePermissionArrayDecision } from '../../hooks/usePermissionArray';
 import { executeWorkflowRouteRef, workflowRouteRef } from '../../routes';
 
-export const RunButton = () => {
+export const RunButton = ({ isAvailable }: { isAvailable?: boolean }) => {
   const { workflowId } = useRouteRefParams(workflowRouteRef);
   const navigate = useNavigate();
   const executeWorkflowLink = useRouteRef(executeWorkflowRouteRef);
@@ -44,6 +44,14 @@ export const RunButton = () => {
       orchestratorWorkflowUseSpecificPermission(workflowId),
     ]);
 
+  let tooltipText = '';
+  if (!canRun) {
+    tooltipText = 'User not authorized to execute workflow.';
+  } else if (!isAvailable) {
+    tooltipText =
+      'The workflow is currently down or in an error state. Running it now may fail or produce unexpected results.';
+  }
+
   return (
     <Grid item container justifyContent="flex-end" xs={12} spacing={2}>
       <Grid item>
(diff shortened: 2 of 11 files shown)


workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
```diff
@@ -121,13 +121,9 @@ export class OrchestratorService {
     instanceId: string;
     cacheHandler?: CacheHandler;
   }): Promise<ProcessInstance | undefined> {
-    const { instanceId, cacheHandler } = args;
+    const { instanceId } = args;
     const instance = await this.dataIndexService.fetchInstance(instanceId);
-    const isWorkflowAvailable = this.workflowCacheService.isAvailable(
-      instance?.processId,
-      cacheHandler,
-    );
```
gciavarrini, line 129: need help to understand this change: it checked if the instance is one of the available from the cache service (an eventually not so updated view of the data index values). So adding the `Workflow status` how does impact this?

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
```diff
@@ -221,12 +217,20 @@ export class OrchestratorService {
...
+        cacheHandler,
+      );
+    } catch (e) {
+      // Log the error but continue fetching the overview
+      console.warn(`Workflow ${definitionId} availability check failed`, e);
+      isWorkflowAvailable = false;
+    }
+    const overview =
+      await this.sonataFlowService.fetchWorkflowOverview(definitionId);
+    if (overview) overview.isAvailable = isWorkflowAvailable;
```
gciavarrini, line 233: if we can retrieve the `overview` (that is `overview` is defined) but `isWorkflowAvailable` was false, than `overview.isAvailable` is false... why? I expected to see ` overview.isAvailable=true`

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
gciavarrini: i understand your point but
```if (!isAvailable && cacheHandler === 'throw') {
      throw new Error(
        `Workflow service "${definitionId}" not available at the moment`,
      );
    }```
if cache handler is set to 'skip' (default value) than it doesn't throw the error. I'm worried that removing it can impact the performance. Can you please double-check?

Comment (LiorSoffer) [reply workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts:129]: I think the isWorkflowAvailable check is unnecessary here for fetching an instance

File: workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts (hunk 2 above)
Comment (LiorSoffer) [reply workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts:233]: It's confusing, it means that the workflow overview is available but the workflow itself isn't, 
maybe I can write it clearer

File: workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts (hunk 1 above)
Comment (LiorSoffer) [reply workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts:129]: we call fetch instance with cache handler set to "throw" so isAvailable throws an error 
image

If I understand your concern correctly, you're worried that we're not using the Cache but the WorkflowCacheService is caching only the availability status of workflow to reduce the number of times we actually ping  the workflow service.

So by removing this line, we might even improve performance because we avoid the additional isAvailable check 
image

File: workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts (hunk 2 above)
Comment (LiorSoffer) [reply workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts]: I'll add a clarifying comment, I'd rather avoid changing the API

File: workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts [hunk 3]
```diff
@@ -221,12 +217,18 @@ export class OrchestratorService {
...
-    return isWorkflowAvailable
-      ? await this.sonataFlowService.fetchWorkflowOverview(definitionId)
-      : undefined;
+    let isWorkflowAvailable;
+    try {
+      isWorkflowAvailable = this.workflowCacheService.isAvailable(
+        definitionId,
+        cacheHandler,
+      );
+    } catch (e) {
```
mareklibra, line 226:
```suggestion
    } catch {
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/RunButton.tsx
```diff
@@ -44,15 +44,25 @@ export const RunButton = () => {
       orchestratorWorkflowUseSpecificPermission(workflowId),
     ]);
 
+  let tooltipText = '';
+  if (!canRun) {
+    tooltipText = 'User not authorized to execute workflow.';
+  } else {
```
mareklibra, line 50:
```suggestion
  } else if (!isAvailable) {
```

workspaces/orchestrator/plugins/orchestrator/src/components/WorkflowPage/RunButton.tsx
```diff
@@ -44,15 +44,25 @@ export const RunButton = () => {
...
+    if (!isAvailable)
+      tooltipText =
+        'The workflow is currently down or in an error state. Running it now may fail or produce unexpected results.';
+    else tooltipText = '';
```
mareklibra, line 54: this `else` is not needed.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
mareklibra: @gciavarrini thanks for spotting that and raising your concern, it looks suspicious. 
But semantically, it seems to be correct. Wdyt?

mareklibra (commented):
Just nitpicks, otherwise LGTM

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
batzionb: I suggest to not pass cachehandler and to remove try catch.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
batzionb: Since now we are providing all data in the data index regardless of if the workflow is available or not, we don't need to handle here the case of the workflow not being available, we should just return it.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
batzionb: This comment is unrelated to the changes in this PR

batzionb (approved):

PR #742, opened 2025-05-07
fix(orchestrator): avoid forced SCM login at workflow run

mareklibra:
This patch fixes the need to log into all so far supported SCM providers before executing any workflow.

Follow-up for #611 .

This part is further planed to be updated within FLPATH-2314.
In a nutshell, we plan to request selected SCM logins during the input parameters collection and then silently collect them at Run the way as #611 suggests.

Before:

Screenshot From 2025-05-07 08-53-25

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
+++ b/workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2024 The Backstage Authors
+ * Copyright Red Hat, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,7 +13,9 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import {
+  AuthRequestOptions,
   DiscoveryApi,
   IdentityApi,
   OAuthApi,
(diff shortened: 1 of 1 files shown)


workspaces/orchestrator/plugins/orchestrator/src/api/OrchestratorClient.ts
```diff
@@ -116,17 +118,18 @@ export class OrchestratorClient implements OrchestratorApi {
       await this.getDefaultReqConfig();
     const authTokens: { provider: string; token: string }[] = [];
 
+    const authRequestOptions: AuthRequestOptions = { optional: true };
     /** Build one promise per provider (guard against missing APIs) */
     const tokenPromises = [
       this.githubAuthApi
         ? this.githubAuthApi
-            .getAccessToken?.()
+            .getAccessToken?.(undefined, authRequestOptions)
```
gciavarrini, line 126: Doubt here.
According to the `getAccessToken` description "[....] The returned promise can be rejected, but only if the user rejects the login request." Do you think we need to add a `.catch(....)` to handle this scenario?
This comment applies to GitLab part as well

mareklibra: With this change, the login form is never displayed, so the user can not reject it.

This code is kind of "collector" of tokens which are already in the cache.
We will add option to login into selected SCMs within FLPATH-2314.

gciavarrini (approved):

PR #746, opened 2025-05-07
fix(sandbox-plugin): make sandbox plugin private

karthikjeeyar:

Mark Sandbox plugin as private as it is not supposed to be published to npm.


mfrancisc (approved):
Thanks!

PR #765, opened 2025-05-12
fix(sandbox): fix page title

mfrancisc:
It contains the following fixes:

- fix the page title:
image

- add back the background color for the Activities page in light mode 

- improve custom error message for phone verification, as suggested in https://github.com/redhat-developer/rhdh-plugins/pull/752/files#r2084893090

### DIFF
--- a/workspaces/sandbox/plugins/sandbox/src/components/SandboxHeader.tsx
+++ b/workspaces/sandbox/plugins/sandbox/src/components/SandboxHeader.tsx
@@ -30,6 +30,7 @@ export const SandboxHeader: React.FC<SandboxHeaderProps> = ({ pageTitle }) => {
   const theme = useTheme();
   return (
     <Header
+      pageTitleOverride={pageTitle}
       title={
         <Typography
           color="textPrimary"
--- a/workspaces/sandbox/plugins/sandbox/src/api/RegistrationBackendClient.tsx
+++ b/workspaces/sandbox/plugins/sandbox/src/api/RegistrationBackendClient.tsx
@@ -140,9 +140,15 @@ export class RegistrationBackendClient implements RegistrationService {
     });
 
     if (!response.ok) {
-      throw new Error(
-        'Invalid phone number. Please verify the country code and number format, then try again.',
-      );
+      const error: CommonResponse = await response.json();
+      // handle backend error messages and make it more user-friendly
+      if (error?.message.includes("Invalid 'To' Phone Number")) {
+        throw new Error(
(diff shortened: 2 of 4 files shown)


workspaces/sandbox/plugins/sandbox/src/components/SandboxHeader.tsx
```diff
@@ -30,6 +30,7 @@ export const SandboxHeader: React.FC<SandboxHeaderProps> = ({ pageTitle }) => {
   const theme = useTheme();
   return (
     <Header
+      pageTitleOverride={pageTitle}
```
Lucifergene, line 33: As per the Slack conversation, I have reviewed the previous PR which caused this regression. 
You can refer to that https://github.com/redhat-developer/rhdh-plugins/pull/752

mfrancisc: The issue should be fixed now. I've added also a quick unit test for the page title.

Lucifergene: This logic wont work all the time. Please see my prev implementation. I have some more logic.
Please revert the Sandboxheader and SandboxPage as mentioned in the prev. PR. 
They should have nothing to do with the Grid.

mfrancisc: Why would not work all the time ? Could you elaborate more on this 😕 

If I'm not mistaken, the previous logic was:
```
pageTitleOverride={!showSubTitle ? 'Activities' : 'Home'}
```
 which sets the title to either `Activities` or `Home` based on the fact that there is a subtitle or not. Which doesn't make much sense IMO because:
 
 1 - the `Powered by RHDH` subtitle should always be displayed in both Home and Activities pages. This also fixes the issues with the menu lines and the title line not being aligned , see issue 2 from https://redhat-internal.slack.com/archives/C085PFGDLGY/p1746716005823379?thread_ts=1746541820.107489&cid=C085PFGDLGY 
 ```
 Can the banner line up with the line in the bottom of the menu like on the home page
 ```
 

2. The old logic assumes the Home page has a subtitle while the Activities has not, which is not so flexible. What happens if we add another page ( third page )? 
 
 Unless there something I'm missing, `pageTitle` is the value we should use for this purpose ( that's also implicit in the variable name :) ) .

Lucifergene:
lgtm :)

alexeykazakov (approved):

PR #766, opened 2025-05-12
fix(bulk-import): fix import api error message for generic error

debsmita1:

Resolves:

https://issues.redhat.com/browse/RHIDP-7478

Screenshot:

Before:
Screenshot 2025-05-12 at 4 07 21 PM

After:
Screenshot 2025-05-12 at 9 44 53 PM

### DIFF
--- a/workspaces/bulk-import/.changeset/violet-planes-greet.md
+++ b/workspaces/bulk-import/.changeset/violet-planes-greet.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-bulk-import': patch
+---
+
+fix generic error when pr creation fails
--- a/workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositories.tsx
+++ b/workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositories.tsx
@@ -19,7 +19,6 @@ import Alert from '@mui/material/Alert';
 import AlertTitle from '@mui/material/AlertTitle';
 import FormControl from '@mui/material/FormControl';
 import { useFormikContext } from 'formik';
-import { get } from 'lodash';
 
 import { AddRepositoriesFormValues, PullRequestPreviewData } from '../../types';
 import { PreviewFileSidebar } from '../PreviewFile/PreviewFileSidebar';
@@ -45,6 +44,9 @@ export const AddRepositories = ({ error }: { error: any }) => {
     });
     setOpenDrawer(false);
   };
+
+  const errorMessage = error?.error?.message && JSON.parse(error.error.message);
+
   return (
     <>
       <FormControl fullWidth>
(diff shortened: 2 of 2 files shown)


ShiranHi:
Thank you @debsmita1 , it is better when we give more context. Is there a way to present the message in a more-friendly way?

christoph-jerolimov:
Maybe we can use an ErrorPanel or ResponesErrorPanel?

https://backstage.io/storybook/?path=/story/data-display-responseerrorpanel--with-title

christoph-jerolimov (approved):
I approve this because this is general fine.

Maybe we can switch to an ErrorPanel or ResponesErrorPanel from Backstage to align it with other errors? Wdyt?

https://backstage.io/storybook/?path=/story/data-display-responseerrorpanel--with-title

debsmita1:
> Thank you @debsmita1 , it is better when we give more context. Is there a way to present the message in a more-friendly way?

@ShiranHi I parsed the response to show the message in a better way and updated the screenshot in the description
About giving more context, there can be various reasons why the API might fail. So, in my opinion we at least show what the error is and let the user edit the selected repositories or do the needful. Let me know what you think

ShiranHi:
> > Thank you @debsmita1 , it is better when we give more context. Is there a way to present the message in a more-friendly way?
> 
> @ShiranHi I parsed the response to show the message in a better way and updated the screenshot in the description About giving more context, there can be various reasons why the API might fail. So, in my opinion we at least show what the error is and let the user edit the selected repositories or do the needful. Let me know what you think

Thank you, it looks better. I see there are several errors, can we have each one of them in a separate line with bullet points to make it easier to digest?

debsmita1:
> > > Thank you @debsmita1 , it is better when we give more context. Is there a way to present the message in a more-friendly way?
> > 
> > 
> > @ShiranHi I parsed the response to show the message in a better way and updated the screenshot in the description About giving more context, there can be various reasons why the API might fail. So, in my opinion we at least show what the error is and let the user edit the selected repositories or do the needful. Let me know what you think
> 
> Thank you, it looks better. I see there are several errors, can we have each one of them in a separate line with bullet points to make it easier to digest?

@ShiranHi This error is more like a message with multiple layers of context, like what caused this error, what is the root cause, and so on

ShiranHi:
> > > > Thank you @debsmita1 , it is better when we give more context. Is there a way to present the message in a more-friendly way?
> > > 
> > > 
> > > @ShiranHi I parsed the response to show the message in a better way and updated the screenshot in the description About giving more context, there can be various reasons why the API might fail. So, in my opinion we at least show what the error is and let the user edit the selected repositories or do the needful. Let me know what you think
> > 
> > 
> > Thank you, it looks better. I see there are several errors, can we have each one of them in a separate line with bullet points to make it easier to digest?
> 
> @ShiranHi This error is more like a message with multiple layers of context, like what caused this error, what is the root cause, and so on

I'm wondering if we have any options to style this type of message a little differently. Could we potentially explore some alternative visual treatments for it, perhaps something [like this](https://www.figma.com/design/09Iltd57RiuqOtKWkjy8MH/Bulk-import?node-id=1482-4355&t=XX7Nv5G3qTey6nPL-1)?

debsmita1:
@ShiranHi Got the idea, but it is a bit difficult to map technical errors like this to customized messages. 'Cause while trying to import from 100+ repositories, the API might fail for various reasons, it may not necessarily fail for this reason everytime

ShiranHi:
> @ShiranHi Got the idea, but it is a bit difficult to map technical errors like this to customized messages. 'Cause while trying to import from 100+ repositories, the API might fail for various reasons, it may not necessarily fail for this reason everytime

In case we have the list of failure reasons/messages we show, I can create each with this kind of a format to help our users read it easily. But maybe this is for another ticket.

debsmita1:
> I approve this because this is general fine.
> 
> Maybe we can switch to an ErrorPanel or ResponseErrorPanel from Backstage to align it with other errors? Wdyt?
> 
> https://backstage.io/storybook/?path=/story/data-display-responseerrorpanel--with-title

I missed addressing this comment, so added it as part of https://issues.redhat.com/browse/RHIDP-7501

PR #779, opened 2025-05-14
fix(Sandbox): Bug fixes observed during Bug Hunting

Lucifergene:
Jira: https://issues.redhat.com/browse/RHDHPAI-798

This pull request introduces several changes to improve error handling, enhance testing, and fix minor bugs in the sandbox plugin. Key updates include expanded error handling in the `RegistrationBackendClient`, new unit tests for backend clients, and bug fixes in the `PhoneVerificationModal` component.

### Error Handling Improvements:
* Expanded error handling in `RegistrationBackendClient` to cover additional backend error messages, such as invalid short codes, invalid mobile numbers, and phone numbers already in use.

### Testing Enhancements:
* Added comprehensive unit tests for `AnsibleBackendClient`, covering methods like `getAAP`, `createAAP`, `unIdleAAP`, and `deleteAAPCR`. These tests simulate various scenarios, including successful responses, conflict errors, and other failures.
* Added unit tests for `RegistrationBackendClient`, testing methods like `getRecaptchaAPIKey`, `getSignUpData`, `initiatePhoneVerification`, `completePhoneVerification`, and `verifyActivationCode`. The tests include mock responses for different cases, such as invalid phone numbers and activation codes.

### Bug Fixes:
* Fixed a typo in the `PhoneVerificationModal` component by renaming `setphoneSubmitError` to `setPhoneSubmitError` for consistency. [[1]](diffhunk://#diff-75607753fa158b5a68d33c6873f3d6fede6b47643ff0ca7d9ad7f399108ff5fcL47-R47) [[2]](diffhunk://#diff-75607753fa158b5a68d33c6873f3d6fede6b47643ff0ca7d9ad7f399108ff5fcL69-R72) [[3]](diffhunk://#diff-75607753fa158b5a68d33c6873f3d6fede6b47643ff0ca7d9ad7f399108ff5fcL78-R81)
* Resolved an issue in `PhoneVerificationModal` where error states were not being reset properly when the modal was closed. Added logic to reset `phoneSubmitError`, `verificationCodeError`, and `codeResent` states on modal close.
* Introduced a delay before calling `refetchUserData` in `PhoneVerificationModal` to ensure data consistency after completing phone verification.


workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/__tests__/SandboxCatalogBanner.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+        </SandboxContext.Provider>
+      </ThemeProvider>,
+    );
+  };
+
+  it('renders loading skeleton when loading is true', () => {
+    renderWithProviders({ loading: true });
+    const skeletons = screen.getAllByTestId('MuiSkeleton-root');
+    expect(skeletons).toHaveLength(3);
+  });
```
mfrancisc, line 134: not sure I fully understand what is this testing 😕 .
I mean what are these skeletons , and why do we have 3 ?

workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/__tests__/SandboxCatalogBanner.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+      company: 'ACME Corp',
+      status: { ready: true, reason: '', verificationRequired: false },
+    };
+    renderWithProviders({ userData });
+    expect(
+      screen.getByText(`Welcome, ${userData.givenName}`),
+    ).toBeInTheDocument();
+  });
+
+  it('renders verification required message when needed', () => {
```
mfrancisc, line 165: the title of the test doesn't seem to reflect what the test does. 
We don't have any `verification required message` if I'm not mistaken. 

Also the user should just see the Try it button spinning in the case that verification is required, if I'm not mistaken.

workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/__tests__/SandboxCatalogBanner.test.tsx
```diff
@@ -0,0 +1,258 @@
...
+        },
+      },
+      pendingApproval: true,
+    });
+    expect(
+      screen.getByText('Please wait for your trial to be approved.'),
+    ).toBeInTheDocument();
+  });
+
+  it('renders trial expiry message when user has end date', () => {
```
mfrancisc, line 211: let's also check that it doesn't render the message when `endDate` is not set

mfrancisc (commented):
Looks Good 👍 
I have only few minor comments

Thanks a lot for these fixes and for the additional unit tests 🏅

workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/__tests__/SandboxCatalogBanner.test.tsx
Lucifergene: the 3 lines in the banner that shows up when the name and date is loading. 
Skeleton is a MUI component.

workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/SandboxCatalogBanner.tsx
```diff
@@ -114,6 +117,9 @@ export const SandboxCatalogBanner: React.FC = () => {
                       fontWeight: 400,
                       fontSize: { xs: '1rem', sm: '1.25rem', md: '1.5625rem' },
                     }}
+                    key={`user-status-${
+                      userData.endDate || ''
+                    }-${verificationRequired}-${pendingApproval}`}
```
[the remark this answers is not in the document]
Lucifergene: By concatenating these values into a single string key, you're ensuring that React will re-render this component whenever any of those values change, since a change in any value produces a different key.

This approach is a common pattern when you need a component to re-render based on multiple dependencies, rather than just an array index or ID. It's particularly useful for conditional rendering states that aren't directly tied to data IDs.

workspaces/sandbox/plugins/sandbox/src/components/Modals/PhoneVerificationModal.tsx
```diff
@@ -86,8 +89,10 @@ export const PhoneVerificationModal: React.FC<PhoneVerificationModalProps> = ({
       setVerificationCodeError(undefined);
       setLoading(true);
       await registerApi.completePhoneVerification(otp.join(''));
-      refetchUserData();
       handleClose();
+      setTimeout(() => {
+        refetchUserData();
+      }, 2000);
```
karthikjeeyar, line 95: arbitrary delay in this case may lead to inconsistencies if the backend operation takes longer than 2 seconds for some reason. Better way could be using polling or retries (with max retry attempts) to ensure completion of the events.

Lucifergene: This is actually an intentional bug fix.

I noticed that right after signing up, it takes a second or two before signupData starts returning the actual user info.
Previously, there was no timeout, so the request would immediately return a 404 — user not found — even though the sign-up had just completed.

It seems the backend takes around 2 seconds to fully propagate the sign-up process and create the user record.

karthikjeeyar: I understand the delay is intentional, but my concern is what if the backend takes slightly more than 2 seconds, the above logic will again return 404 (user not found). but lets say, If you retry fetching for every second for 5 times (max attempts), as soon as you get the response you can stop retrying and it always guarantees you get the data (it could be even on the first try (~1s) or on the fourth try (~4s).

mfrancisc: +1 

In addition we should leave the spinner on the button , so it makes it clear for the User that the operation is in progress, avoding confusion, and avoiding extra clicks again on the Try it button.

workspaces/sandbox/plugins/sandbox/src/components/SandboxCatalog/SandboxCatalogBanner.tsx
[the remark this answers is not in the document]
mfrancisc: Thanks for explaining!

workspaces/sandbox/plugins/sandbox/src/components/Modals/PhoneVerificationModal.tsx
Lucifergene: Done

mfrancisc (approved):

karthikjeeyar (approved):
/lgtm

PR #781, opened 2025-05-14
feat(global-header): extract the first-name from the user's displayname

debsmita1:

**Resolves:**

https://issues.redhat.com/browse/RHIDP-7522

Screenshot 2025-05-14 at 7 48 31 PM

### DIFF
--- a/workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/ProfileDropdown.tsx
+++ b/workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/ProfileDropdown.tsx
@@ -100,7 +100,7 @@ export const ProfileDropdown = ({ layout }: ProfileDropdownProps) => {
                   mr: '1rem',
                 }}
               >
-                {displayName}
+                {displayName.split(' ')[0]}
               </Typography>
             </>
           )}
--- a/workspaces/global-header/.changeset/nine-deers-play.md
+++ b/workspaces/global-header/.changeset/nine-deers-play.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-header': patch
+---
+
+extract the first-name from the user's displayname


workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/ProfileDropdown.tsx
```diff
@@ -100,7 +100,7 @@ export const ProfileDropdown = ({ layout }: ProfileDropdownProps) => {
                   mr: '1rem',
                 }}
               >
-                {displayName}
```
logonoff, line 103: i don't think it's safe to assume that a first name will be one word:
- e.g.,: https://academia.stackexchange.com/questions/78746/how-to-get-others-to-understand-that-my-first-name-consists-of-two-words
- some chinese names, e.g.,: https://en.wikipedia.org/wiki/Loh_Kean_Yew (the first name is `Kean Yew`)

debsmita1: @ShiranHi, any thoughts on this ? 
I still believe we should show the "displayName" as is

ShiranHi: Good point! 
Is that the same as what we have on the homepage card? 
I think there's enough space in the global header, so let's use the full name.

debsmita1: yes, @Eswaraiahsapram is working on it

Would you have any concerns if we close this ticket for now, as we are already displaying the full name ? 
We can certainly revisit it later if required , or if the customers raise any further issues.

debsmita1:
Create a new PR related to this . See https://github.com/redhat-developer/rhdh-plugins/pull/792
so closing this one

PR #787, opened 2025-05-15
chore(orchestrator): make OrchestratorFormContext singleton in dynamic plugins

mareklibra:
Fixes: FLPATH-2343

Contrary to the static plugins, the former way of creating a react context  lead to two instances created by two plugins at runtime when used in RHDH dynamic plugins.

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-form-api/report.api.md
+++ b/workspaces/orchestrator/plugins/orchestrator-form-api/report.api.md
@@ -4,14 +4,12 @@
 
 ```ts
 
-/// <reference types="react" />
-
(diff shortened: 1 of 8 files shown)


batzionb:
@mareklibra with this new approach the user developing a plugin needs to provide this context even if he isn't using it
can we use the rjsf form context instead?

batzionb (approved):

batzionb:
> @mareklibra with this new approach the user developing a plugin needs to provide this context even if he isn't using it can we use the rjsf form context instead?

Just for the record - we agreed to merge this now and refactor later

PR #798, opened 2025-05-18
fix(orchestrator):force pre-check for workflow availability before co…

LiorSoffer:
resolve [FLPATH-2055](https://issues.redhat.com/browse/FLPATH-2055)

https://github.com/user-attachments/assets/36eb7b73-6e23-4a3f-b575-e3c6db89d133


batzionb:
@LiorSoffer can you add a screen shot of the dialog?
Do we want a dialog and not just an alert?

batzionb:
@LiorSoffer 
Let's do the following:
1. the execute page itself should handle this instead of each page with the run button
2. implement a new endpoint that actually pings sonata and displays the error message it returns

PR #800, opened 2025-05-19
feat(lightspeed): add configurable sample prompts in lightspeed

karthikjeeyar:
## Fixes
https://issues.redhat.com/browse/RHDHPAI-488

## Description

Sample prompts can be configured by platform engineers, three random prompts will be used by the user configured prompts and if no user prompts are configured then it will pick from the [default prompts](https://github.com/redhat-developer/rhdh-plugins/compare/main...karthikjeeyar:add-sample-prompts?expand=1#diff-ead336386fe9d8d6c984ac30ac0dceb43a688358438875f983e18812b922dc91). 
     
 configration:    
  ```yaml
  lightspeed:
    prompts:
      - title: 'Write an article'
        message: Can you write an article on event driven architecture?
      - title: 'Set up a Tekton pipeline for app deployment'
        message: 'Can you help me automate the deployment of my application using Tekton pipelines?'
  ```   

Note: Prompts are picked based on the `questionValidation` setting. If `questionValidation` is set to `true` then RHDH, openshift, tekton specific prompts are used.

## Screenshots:

image

     

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed/config.d.ts
+++ b/workspaces/lightspeed/plugins/lightspeed/config.d.ts
@@ -43,7 +43,7 @@ export interface Config {
      * @visibility frontend
      */
     questionValidation?: boolean;
-    samplePrompts?: Array</**
+    prompts?: Array</**
      * @visibility frontend
      */
     {
--- a/workspaces/lightspeed/plugins/lightspeed/config.d.ts
+++ b/workspaces/lightspeed/plugins/lightspeed/config.d.ts
@@ -43,5 +43,21 @@ export interface Config {
      * @visibility frontend
      */
     questionValidation?: boolean;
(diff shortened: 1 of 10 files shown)


workspaces/lightspeed/plugins/lightspeed/config.d.ts
```diff
@@ -38,5 +38,21 @@ export interface Config {
        */
       url: string;
     }>;
+    samplePrompts?: Array</**
```
yangcao77, line 41: should we call those `prmopt`? or `query`?
as later on, IIRC,  we may support dynamic system prompt as well

karthikjeeyar: Since it is an array, I updated it to `prompts`.

PR #801, opened 2025-05-19
chore(orchestrator): remove orchetrator's context in favor of RJSF form's one

mareklibra:
Follow-up for #787


workspaces/orchestrator/docs/extensibleForm.md
```diff
@@ -1,3 +1,5 @@
+TODO: This is outdated
```
mareklibra, line 1: I will update once we agree on the changes here

batzionb:
Originally the props were passed from OrchestratorFormWrapper to FormComponent using a context because I couldn't think of a better way to pass them, but the context contains implementation details like numStepsInMultistepSchema that plugin developers aren't supposed to know about.
So I'm wondering what is it in the context that caused it to be created twice as a dynamic plugin, and what is special about it that is different from any other react context?

batzionb:
I do know that contexts can't be shared in a micro frontend, so I guess that's the problem

mareklibra:
The issue is related to  micro-frontend only. With statically linked plugins, this is not happening since the former line:
https://github.com/redhat-developer/rhdh-plugins/pull/787/files#diff-b7e6807c6b87b52742257e380eb7b5dbe871974c0b24b36e319c04adc23dceb1L43

is executed just once.

The RHDH dynamic plugins recently do not have any mechanism how to share contexts.

Since we used to have the context in the share `orchestrator-form-api` module, such "module-level" code was reevaluated once when the plugin is loaded and once when the orchestrator plugin is loaded, each time leading to a different identity of the resulting object (context).

There are a few listed cases where the RHDH devs provided sharing of the context but the list is not extendable, plugins need to find another way.

Options considered:
- yet another API/ApiRef just for the context
- the factory method producing singleton (first implementation of this PR)
- reusing the RJSF context for our data (recent implementation).

I think it is not an issue if the widgets can see more data, like the `numStepsInMultistepSchema`.

mareklibra:
>  if there's no decorator plugin it won't work the whole form is broken
We can fix it, I havn't run the default API implementation so far

mareklibra:
> I'm looking at other approaches

Honestly, the context singleton seems like the easiest way.

batzionb:
I suggest to pass the data from OrchestratorFormWrapper to the decorator with props
And the widgets plugin itself creates the formContext to be used in it's own widgets

batzionb:
The main issue IMO is not the extra props being exposed but the plugin developer needs to know to pass the context because of our implementation details

mareklibra:
> The main issue IMO is not the extra props being exposed but the plugin developer needs to know to pass the context because of our implementation details

Any non-trivial set of widgets will need it anyway.

Moreover, it's just one prop per whole library, one line, a few letters. Straightforward and if forgotten then clearly announced.

We can play with the `OrchestratorFormContextProps` - to split it. Just the `FormComponent` from the `OrchestratorFormWrapper.tsx` needs the whole set. The widgets just a small subset.

batzionb:
> > The main issue IMO is not the extra props being exposed but the plugin developer needs to know to pass the context because of our implementation details
> 
> Any non-trivial set of widgets will need it anyway.
> 
> Moreover, it's just one prop per whole library, one line, a few letters. Straightforward and if forgotten then clearly announced.
> 
> We can play with the `OrchestratorFormContextProps` - to split it. Just the `FormComponent` from the `OrchestratorFormWrapper.tsx` needs the whole set. The widgets just a small subset.

OK let's stay with the singleton

batzionb:
We can close this PR

batzionb (approved):

PR #802, opened 2025-05-19
fix(orchestrator): add endpoint to force-check availability of a given workflow

LiorSoffer:
resolve [FLPATH-2056](https://issues.redhat.com/browse/FLPATH-2056)


workspaces/orchestrator/plugins/orchestrator-common/src/generated/client/api.ts
```diff
@@ -1530,6 +1575,15 @@ export const DefaultApiFactory = function (configuration?: Configuration, basePa
         getWorkflowsOverview(searchRequest?: SearchRequest, options?: any): AxiosPromise<WorkflowOverviewListResultDTO> {
             return localVarFp.getWorkflowsOverview(searchRequest, options).then((request) => request(axios, basePath));
         },
+        /**
+         * Returns true if the workflow service is up for the given workflow workflow ID.
```
batzionb, line 1579:
```suggestion
         * Returns true if the workflow service is up for the given workflow workflow ID.
```
```suggestion
         * Returns true if the workflow service is up for the given workflow ID.
```

batzionb (approved):
Looks good :)
Just one nit
And PTAL at sonar complaints

workspaces/orchestrator/plugins/orchestrator-backend/src/service/OrchestratorService.ts
```diff
@@ -223,4 +223,17 @@ export class OrchestratorService {
     if (overview) overview.isAvailable = isWorkflowAvailable; // workflow overview is avaiable but the workflow itself is not
     return overview;
   }
+
+  public async pingWorkflowService(args: {
+    definitionId: string;
+  }): Promise<boolean | undefined> {
+    const { definitionId } = args;
+    const idUrlMap = await this.dataIndexService.fetchWorkflowServiceUrls();
```
batzionb, line 231: we shouldn't get the serviceUrl like this
it should request it only for current definitionId using 
https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts#L132

batzionb: if fetchWorkflowInfo returns undefined, it should raise an error that the workflow doesn't exist

batzionb:
@LiorSoffer post merge I found some issues with this PR, please create a follow up one

PR #845, opened 2025-05-24
fix(bulk-import): align eslint-rules as per other plugins

its-mitesh-kumar:
## Description

Align ESLint rules as per other plugins 

## Fixes

Fixed : https://issues.redhat.com/browse/RHIDP-7386

## UI after changes

https://github.com/user-attachments/assets/86b4d4a3-a6fe-4455-b45f-fcf0b4d5147f


debsmita1 (approved):
verified the changes locally
/lgtm

PR #882, opened 2025-06-01
fix(bulk-import): removing extra apis call

its-mitesh-kumar:
## Description

This PR focuses on improving the performance and efficiency of the Bulk Import page by reducing unnecessary API calls and improving user experience.

###  Problem
- When switching between the **Organizations** and **Repositories** tabs, API calls for both tabs were being made — even if only one was actively in view.
- Unintended API requests were triggered by generic screen clicks.
- Pagination controls triggered API calls even when the page number or page size didn't change.
- Search input was missing **debounce**, resulting in multiple API requests for each keystroke (e.g., typing “test” sent 4 requests).

### Fixes 
Fixing https://issues.redhat.com/browse/RHIDP-7658

###  Fixes Implemented
1. **Tab Switching Optimization**  
   - Ensured that only the active tab triggers its respective API call.

2. **Reduced Redundant Triggers**  
   - Eliminated unnecessary API calls triggered by unrelated user interactions (e.g., page clicks).

3. **Pagination Optimization**  
   - API calls now only trigger when actual changes occur (e.g., page or page size changes).

4. **Search Input Debounce**  
   - Added `lodash.debounce` to avoid excessive API calls on fast typing.

---
### Recording for Before Changes

https://github.com/user-attachments/assets/2f20751f-1553-4a7b-989e-7e1a56f51c5d

### Recording for After Changes

https://github.com/user-attachments/assets/786f9d49-a0ae-4943-aebe-a3d7e84eaabe

---
###  How to Test
1. Go to the Bulk Import Add repository page.
2. Switch between Organizations and Repositories.
3. Interact with the pagination control without changing values.
4. Type quickly in the search bar and observe the number of requests.

---

### Result
- Lower network traffic.
- Smoother UX for users dealing with large sets of repositories/organizations.
- Increased frontend responsiveness.

### DIFF
--- a/workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositoriesTable.tsx
+++ b/workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositoriesTable.tsx
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import Box from '@mui/material/Box';
 import Paper from '@mui/material/Paper';
 import { useFormikContext } from 'formik';
+import debounce from 'lodash.debounce';
 
 import { useNumberOfApprovalTools } from '../../hooks';
 import {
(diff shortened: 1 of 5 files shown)


workspaces/bulk-import/plugins/bulk-import/src/components/AddRepositories/AddRepositoriesTable.tsx
```diff
@@ -14,11 +14,12 @@
  * limitations under the License.
  */
 
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useMemo, useState } from 'react';
 
 import Box from '@mui/material/Box';
 import Paper from '@mui/material/Paper';
 import { useFormikContext } from 'formik';
+import debounce from 'lodash.debounce';
```
debsmita1, line 22: Could you try using the `useDebounce` hook from `react-use` library ?
We have used this one in our other plugins

its-mitesh-kumar: @debsmita1 Thanks for the review!
Since `lodash.debounce` is already present indirectly via `@babel/helper-define-polyfill-provider`, I preferred using it to avoid adding a new dependency `react-use` just for this functionality — which would slightly increase the bundle size.

Let me know your thoughts on this

debsmita1 (changes requested):

debsmita1 (approved):
/lgtm

PR #883, opened 2025-06-02
SANDBOX-1200: readjust e2e tests sandbox for e2e mode

rsoaresd:
## Description
We need to re-adjust the current e2e tests for the e2e environment.

OCP cluster does not have a valid CA, so in the e2e tests, we need to tell the browser to ignore HTTPS errors. Furthermore, we also need to readjust the test 'Test DevSandbox' to the e2e env.

## How to test
1. You need a OCP cluster 
    - ROSA cluster from ClusterBot will not work since we are not able to modify the OAuth configuration of ROSA clusters created by the ClusterBot.
2. Switch to [this branch](https://github.com/rsoaresd/toolchain-e2e/tree/deploy-sandbox-ui-in-e2e-tests-mode) and make sure the toolchain-e2e directory is located in the same parent folder as rhdh-plugins, like this:
```
<your-repos-folder>/
  ├── toolchain-e2e
  └── rhdh-plugins
```
3. In https://github.com/rsoaresd/rhdh-plugins/tree/readjust_e2e_tests_for_e2e_mode:
- Ensure you are using Node.js version 20
    - to easily manage it, you can run `nvm use 20`
- `cd workspaces/sandbox`
- `make test-e2e-local SSO_USERNAME=${SSO_USERNAME} SSO_PASSWORD=${SSO_PASSWORD}`

## Issue ticket number and link
[SANDBOX-1200](https://issues.redhat.com/browse/SANDBOX-1200)


mfrancisc (commented):
Looks good 👍 

I have few comments only :

- we could document the commands needed for running/stopping/cleaning the e2e tests in the [e2e readme](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/sandbox/packages/app/e2e-tests/README.md)
 - should we move the `make clean-sandbox-ui` to this repo? Since the start command is here.
 
 
 Also not related to this PR, and we could address this as a follow up work, we should consider containerizing the e2e tests so that all the required dependencies are available within the container, thus avoiding installing those locally, this would make the tests reproducible. As for example it seems that at least for firefox is trying to use some dependencies available in my laptop ( outside of the projects ):
 ```
 1) [firefox] › packages/app/e2e-tests/app.test.ts:52:7 › sandbox plugin › Test Homepage ──────────

    Error: browserType.launch: Executable doesn't exist at /Users/fmuntean/Library/Caches/ms-playwright/firefox-1482/firefox/Nightly.app/Contents/MacOS/firefox
    ╔═════════════════════════════════════════════════════════════════════════╗
    ║ Looks like Playwright Test or Playwright was just installed or updated. ║
    ║ Please run the following command to download new browsers:              ║
    ║                                                                         ║
    ║     yarn playwright install                                             ║
    ║                                                                         ║
    ║ <3 Playwright Team                                                      ║
    ╚═════════════════════════════════════════════════════════════════════════╝
 ```  
 So basically having a Dockerfile and building a container with inside the tests and all the required dependencies. I think this would help also with running those in the CI pipeline.

mfrancisc (commented):
Looks good , great job 🏅 

As we briefly discussed, I would move all this to the toolchain-e2e repo. WDYT?

PR #916, opened 2025-06-06
chore(orchestrator): move selected deps among peerDeps

mareklibra:


mareklibra:
still relevant, need rebase

mareklibra:
Not-having peerDependencies is not optimal but keeps us aligned with other plugins unless this approach is changed systematically. So closing without merging.

PR #960, opened 2025-06-13
functional mcp server prototype as backstage router backend plugin

gabemontero:
including both the streamable mcp http protocol and the prior sse protocol (which is what llama-stack supports)

this most likely is a temporary addition until spotify releases their core mcp support, but this prototype will help @johnmcollier @thepetk and Ugo with their upcoming prototyping


gabemontero:
with @johnmcollier 's https://github.com/redhat-ai-dev/community-plugins/pull/1 I think we can officially close this PR.

I'll keep the branch alive for
- facilitating Ugo's initial assignment
- if the SSE problem seen with Cursor to Backstage MCP server ends up being an issue on the backstage side and not the client side

PR #1026, opened 2025-06-23
feat(extensions): introduce GET endpoints for install status

dzemanov:

Introduce GET endpoints `/plugin/:namespace/:name/configuration/status` and `/package/:namespace/:name/configuration/status` that return 
```
{
    installStatus: MarketplacePackageInstallStatus/MarketplacePluginInstallStatus,
    disabled: boolean
}
```

Fixes [RHIDP-7854](https://issues.redhat.com/browse/RHIDP-7854)
Dependent upon: https://github.com/redhat-developer/rhdh/pull/3021 that writes `dynamic-plugins.final.yaml` under `dynamic-plugins-root` with final `disabled` state of dynamic plugins. The less dynamic approach could be to use annotation `extensions.backstage.io/pre-installed` ([example](https://github.com/redhat-developer/rhdh/blob/main/catalog-entities/marketplace/plugins/3scale.yaml#L9)) to figure this out.

#### To do:
- Add tests
- Add how to test this PR


ciiay:
Hi @dzemanov , thanks for the detailed analysis ❤️  This makes a lot of sense.

I think we should definitely keep the `installStatus` enrichment in the processor, since filtering plugins by `installStatus` is part of our design ([Figma](https://www.figma.com/design/E6C9X3OSVzlWkW7ZsbxN7s/Administration?node-id=6479-3805&t=aUBQTZ6xB9EOxGRs-0)).

Also, I agree that your idea of merging the plugin and package processors/providers is a good direction. It could help reduce `catalogApi` calls and make the logic cleaner. The main tradeoff I see is that it adds some complexity around state management and tighter coupling between plugin and package logic. So probably worth pursuing if performance or duplication becomes a concern. @christoph-jerolimov what do you think?

dzemanov:
Thanks @ciiay, I will go ahead then and unify providers and processors. What could also work is to just reread the package/plugin files, compute plugin->package map and recalculate `installStatus` for packages in plugin install status processor. I will try unifying first as it seems to me a nicer approach, although yes, it can add code complexity.

What is the status on Marketplace collections? Do I need to include those since the entity provider i[s disabled](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/marketplace/plugins/catalog-backend-module-marketplace/src/module.ts#L58)? Processor is not though.

ciiay:
> What is the status on Marketplace collections since the entity provider i[s disabled](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/marketplace/plugins/catalog-backend-module-marketplace/src/module.ts#L58)? Processor is not though.

Hi @christoph-jerolimov , what is our plan for collections?

PR #1046, opened 2025-06-25
fix(orchestrator): forward backstage token to sonataflow execute workflow

batzionb:


batzionb:
> Let's add a note to the documentation that the token can be invalidated at any time. We have no guarantee about its lifespan or similar control.
> 
> The token is valid at the time of making the call.

How is it different from github token? I tried to find documentation on the lifespan of backstage token

PR #1050, opened 2025-06-25
fix(orchestrator): changed auth headers to start with X-Authorization

batzionb:
Fixed headers to  X-Authorization-<provider> instead of X-<provider>-header to match documentation and ADR
backport https://github.com/redhat-developer/rhdh-plugins/pull/1051


workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.ts
```diff
@@ -121,6 +121,7 @@ export class SonataFlowService {
       );
     }
 
+    console.log(headers);
```
mareklibra, line 124: leftover

batzionb: fixed thanks

PR #1085, opened 2025-06-30
fix: ensure partition start and end date are created in UTC timezone

clanghout:

I changed the code for the adoption insights partitions to ensure new dates are created in the UTC timezone. (also by changing to Luxon, we can use toIsoDate() instead of slicing the result)

Reason for this was that my local run of backstage for some reason takes timezones into account in this script. That results in the partition to be created for the timerange `FOR VALUES FROM ('2025-05-31 00:00:00+00') TO ('2025-06-30 00:00:00+00')` which fails today (30th of June) because the value is outside of the range, while a new partition is not yet created.

Example from the browser:
```
new Date(2025,5,1)
Date Sun Jun 01 2025 00:00:00 GMT+0200 (Central European Summer Time)

new Date(2025,5,1).toISOString()
"2025-05-31T22:00:00.000Z" 
```

Unfortunately, I could not create a test to verify the bug is fixed. I could not find a way to control the timezone for new Date() in a test.

### DIFF
--- a/workspaces/adoption-insights/.changeset/sad-doodles-bake.md
+++ b/workspaces/adoption-insights/.changeset/sad-doodles-bake.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-adoption-insights-backend': patch
+---
+
+Ensure Dates created for partitioning always are created with the UTC timezone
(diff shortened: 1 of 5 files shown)


clanghout:
The test I added fails with the old `partition.ts` code and running with `TZ=Europe/Amsterdam yarn test`. Running the test the same way, but with the new code the test succeeds.

karthikjeeyar:
Thank you @clanghout for the contribution. The changes looks good to me! 

you may have to run `yarn --cwd workspaces/adoption-insights dedupe` to fix the failing PR check.

PR #1090, opened 2025-07-01
fix(orchestrator):Handle errors on fetch calls by adding try/catch bl…

LiorSoffer:
resolve [FLPATH-2407](https://issues.redhat.com/browse/FLPATH-2407)
image


workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -202,27 +204,6 @@ describe('SonataFlowService', () => {
       );
     });
 
-    it('should propagate fetch thrown error', async () => {
```
batzionb, line 205: what's the reason to remove this?

PR #1093, opened 2025-07-02
fix(adoption-insights): Handle Timezone-aware date grouping and tooltip formatting

karthikjeeyar:
## Fixes:

https://issues.redhat.com/browse/RHDHPAI-924

##  Description:

Our application processes and visualizes timestamped user event data across different timezones. However, inconsistencies were observed when:

Analytics events occurred near midnight UTC or in non-local timezones.
Frontend simulated a different timezone 
Backend used Intl.DateTimeFormat().resolvedOptions().timeZone, which could default to the server’s timezone (e.g., PDT/IST), causing misalignment when interpreting frontend inputs.

This PR fixes the above issue and contains following changes:

- Backend now accepts an explicit `timezone` parameter from the frontend, rather than relying on server defaults.
- Rewrote getDateRange() to use `formatInTimeZone()` from date-fns-tz, ensuring dates are computed relative to the selected timezone.
- Backend Query Alignment: Postgres queries now uses `timezone`, ensuring `created_at` is interpreted correctly relative to the selected user timezone.
- Before returning timestamps to the frontend, we format them using Luxon to respect user selected timezone.
- Updated chart tooltip to add more context about the date and time ranges.
- Fixed Search chart query which wasn't passing `limit` parameter, so backend defaults to only 3 data points which caused some missing data points in the chart.

## How to test

Locally
---

- you can test with default better-sqlite database by running `yarn start`

Cluster:
---

[Recommended]: Cluster mode will use Postgres database, so it is very close to production environment.

- Use this image in the cluster `quay.io/karthik_jk/adoption-insights:pr-1093` to load the adoption insights plugin
    ```
    plugins:
      - package: oci://quay.io/karthik_jk/adoption-insights:pr-1093!red-hat-developer-hub-backstage-plugin-adoption-insights
        disabled: false
      - package: oci://quay.io/karthik_jk/adoption-insights:pr-1093!red-hat-developer-hub-backstage-plugin-adoption-insights-backend
        disabled: false
      - package: oci://quay.io/karthik_jk/adoption-insights:pr-1093!red-hat-developer-hub-backstage-plugin-analytics-module-adoption-insights
        disabled: false
    
    ```
- Events should always be displayed even if it is created in different timezone with proper local timezone interpretation.

_NOTE: Local setup with default better-sqlite3 will not work, this needs POSTGRES database to test_

## Tooltip Screenshots:

Today:
today
----
Last 28 days: 
last_28_days
---
Last week:
last_week
--- 
Last year:
last_year
---


ShiranHi:
Looks great, thank you!

Eswaraiahsapram (approved):
Verified, it works as expected.

Screenshot 2025-07-03 at 11 50 34 AM

/lgtm

PR #1113, opened 2025-07-04
fix(orchestrator): build array from ActiveMultiSelect selector result

mareklibra:
Fixes: FLPATH-2455

The `fetch:response:autocomplete` selector of ActiveMultiSelect is expected to result in an array of strings.

This patch eases this requirement, if just a single string is produced, the array is built automatically from it.


mareklibra:
Related change: https://github.com/rhdhorchestrator/orchestrator-demo/pull/46

batzionb (approved):

PR #1119, opened 2025-07-05
chore(orchestrator): add tests for applySelector

mareklibra:
Small fix and tests for `applySelector` (jsonata)


mareklibra:
Rebased

PR #1140, opened 2025-07-10
feat(orchestrator): add support ui:order in UI schema

mareklibra:
Fixes: FLPATH-2469

The workflow data input schema can newly contain `ui:order` property (https://rjsf-team.github.io/react-jsonschema-form/docs/usage/objects/#specifying-property-order)

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-form-react/src/utils/generateUiSchema.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-form-react/src/utils/generateUiSchema.ts
@@ -21,6 +21,7 @@ import { JsonObject } from '@backstage/types';
 import { UiSchema } from '@rjsf/utils';
 import type { JSONSchema7, JSONSchema7Definition } from 'json-schema';
 import get from 'lodash/get';
+import merge from 'lodash/merge';
 import set from 'lodash/set';
 
 /**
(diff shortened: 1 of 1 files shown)


workspaces/orchestrator/plugins/orchestrator-form-react/src/utils/generateUiSchema.ts
```diff
@@ -110,9 +111,19 @@ function extractUiSchema(mixedSchema: JSONSchema7): UiSchema<JsonObject> {
     } else {
       processLeafSchema(curSchema, path);
     }
+    processOrder(curSchema, path);
     processComposedSchema(curSchema, path);
   };
 
+  const processOrder = (curSchema: JSONSchema7Definition, path: string) => {
```
batzionb, line 118: would be nice to have this in the generateUitSchema unit test, this will also explain this code

batzionb: but not a must

workspaces/orchestrator/plugins/orchestrator-form-react/src/utils/generateUiSchema.ts
```diff
@@ -110,9 +111,19 @@ function extractUiSchema(mixedSchema: JSONSchema7): UiSchema<JsonObject> {
     } else {
       processLeafSchema(curSchema, path);
     }
+    processOrder(curSchema, path);
```
batzionb, line 114: it's not under the schema properties? shouldn't it be there? if it were it would be handled IIUC

batzionb: OK I got it, nvm

PR #1154, opened 2025-07-11
fix(sandbox): update openshift url

mfrancisc:
Update openshift url :
from <cluster>/add/ns/<namespace-dev> to <cluster>/k8s/cluster/projects/<namespace-dev> in Sandbox Dashboard

Jira: https://issues.redhat.com/browse/SANDBOX-1350

### DIFF
--- a/workspaces/sandbox/plugins/sandbox/src/hooks/useProductURLs.ts
+++ b/workspaces/sandbox/plugins/sandbox/src/hooks/useProductURLs.ts
@@ -50,7 +50,7 @@ export const productsURLMapping = (userData: SignupData | undefined) => {
     {
       id: Product.OPENSHIFT_CONSOLE,
       url: isProvisioned
-        ? `${userData?.consoleURL}/add/ns/${userData?.defaultUserNamespace}`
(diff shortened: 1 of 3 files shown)


logonoff (approved):

workspaces/sandbox/plugins/sandbox/src/hooks/useProductURLs.ts
```diff
@@ -50,7 +50,7 @@ export const productsURLMapping = (userData: SignupData | undefined) => {
     {
       id: Product.OPENSHIFT_CONSOLE,
       url: isProvisioned
-        ? `${userData?.consoleURL}/add/ns/${userData?.defaultUserNamespace}`
+        ? `${userData?.consoleURL}/k8s/cluster/projects/${userData?.defaultUserNamespace}`
```
alexeykazakov, line 53: I think we could be a bit smarter here and check if consoleURL already has the trailing `/` and if so then don't add another one. That would help with avoiding `//` in the URL. We can address it in a separate PR though.

alexeykazakov (approved):

PR #1203, opened 2025-07-22
Revert "chore(orchestrator): updated backstage version for overlay repo to publish images compatible with 1.7 (#1184)"

batzionb:
Revert #1184

The change introduced in #1184 was incorrect. The version override workaround was actually needed in the overlays repository, and was handled in [this PR](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/1318)

For transparency, the Backstage version in the codebase should accurately reflect the versions of the installed packages. Therefore, this PR should be reverted.

To properly resolve the issue, a full upgrade of Backstage—including all related dependency updates—is required.


batzionb:
@mareklibra PTAL

PR #1221, opened 2025-07-24
fix(orchestrator):Handle errors on fetch calls by adding try/catch bl…

LiorSoffer:
backport #1090


batzionb:
we will include this fix in 1.7

PR #1263, opened 2025-08-07
fix(orchestrator): filter deleted workflows from workflows table

batzionb:
[FLPATH-2333](https://issues.redhat.com/browse/FLPATH-2333)

Filters from workflows table the deleted workflows that are marked with metadata.status = 'unavailable' by the SonataFlow operator in DataIndex. These are workflows that were marked this way by sonataflow operator when a SonataFlow CR is deleted.
Includes fix of incorrect metadata type.


mareklibra:
LGTM once CI is green

mareklibra (approved):

PR #1277, opened 2025-08-09
feat(quickstart): added support to role-based quickstart

ciiay:

For [RHIDP-8455](https://issues.redhat.com/browse/RHIDP-8455).

This PR adds role-based filtering for Quickstart items, so only steps relevant to the current user’s role (admin, developer, etc.) are shown in the UI. It updates config, data, and UI to support this, improves empty state messaging, adds new icons, and expands test coverage for these changes.

Sections for "developer" role
image

When no content available for current role
image

### DIFF
--- a/workspaces/quickstart/plugins/quickstart/src/components/QuickstartDrawer.tsx
+++ b/workspaces/quickstart/plugins/quickstart/src/components/QuickstartDrawer.tsx
@@ -20,6 +20,7 @@ import { configApiRef, useApiHolder } from '@backstage/core-plugin-api';
 import { Quickstart } from './Quickstart';
 import { useQuickstartDrawerContext } from '../hooks/useQuickstartDrawerContext';
 import { QuickstartItemData } from '../types';
+import { filterQuickstartItemsByRole } from '../utils';
 
 export const QuickstartDrawer = () => {
   const { isDrawerOpen, closeDrawer, drawerWidth } =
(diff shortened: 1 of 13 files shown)


workspaces/quickstart/plugins/quickstart/src/components/QuickstartDrawer.tsx
```diff
@@ -30,6 +31,12 @@ export const QuickstartDrawer = () => {
   const quickstartItems: QuickstartItemData[] = config?.has('app.quickstart')
     ? config.get('app.quickstart')
     : [];
+
+  // This will be dynamically determined based on the logged-in user
+  const userRole = 'developer'; // switch to 'admin', 'developer', or other roles for testing purposes
```
rohitkrai03, line 36: Do you plan to update this later to use roles dynamically?

rohitkrai03 (changes requested):
@ciiay I don't see any logic to fetch user role dynamically and filter based on that. Is that planned for follow up PR or am I missing something?

workspaces/quickstart/plugins/quickstart/src/components/QuickstartDrawer.tsx
rohitkrai03: I would expect it to be completed in the same PR or story. Going ahead with a hardcoded value like that doesn't make sense to me.

PR #1279, opened 2025-08-11
fix(scorecard): add UI with mock data for Scorecard

Eswaraiahsapram:

Fix: https://issues.redhat.com/browse/RHIDP-8522

- Implemented Scorecard UI that consumes mock data, with loading state.
- Structured mock data so it can be easily swapped with the actual backend API later with minimal changes.

Screenshots

Screenshot 2025-08-11 at 11 35 05 AM

### DIFF
--- a/workspaces/scorecard/plugins/scorecard/src/components/ScorecardPage/Scorecard.tsx
+++ b/workspaces/scorecard/plugins/scorecard/src/components/ScorecardPage/Scorecard.tsx
@@ -13,45 +13,53 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { ChartDonut, ChartThemeColor } from '@patternfly/react-charts/victory';
-import WarningAmberIcon from '@mui/icons-material/WarningAmber';
+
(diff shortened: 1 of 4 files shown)


HusneShabbir (changes requested):
Hey @Eswaraiahsapram , noticed that the warning value is currently set using the pattern > \\\d+/, but according to the Figma design, it should represent a range specifically, from the lower limit to the upper limit. Could you please adjust this to reflect the intended behavior as per the design?

Eswaraiahsapram:
> Hey @Eswaraiahsapram , noticed that the warning value is currently set using the pattern > \\d+/, but according to the Figma design, it should represent a range specifically, from the lower limit to the upper limit. Could you please adjust this to reflect the intended behavior as per the design?

Thank you, @HusneShabbir, for reviewing. Yes, we can fix this. However, this is just the UI part of the Scorecard with mock data. As we move forward, thresholds will come from the backend, so we can address this properly in upcoming PRs.

christoph-jerolimov (approved):

PR #1283, opened 2025-08-12
feat(orchestrator): support markdown as workflow result value format

mareklibra:
Screenshot From 2025-08-12 13-33-47


batzionb:
Perhaps the more common use case would be to simply provide a markown text that would fill the whole card instead of using the outputs and links? we can have an output schema that either just gets a markdown text and the message or the result object with all the fields we defined

batzionb:
My suggestion is that the json schema include both options, in json schema oneOf can be used
Question is what is the common use case - putting one sentence? or a paragraph? How would it look with a bulleted paragraph?
We provide an oppinionated UI for showing the results 
IIUC the markdown capability should allow the user to show whatever he wants in that card

PR #1296, opened 2025-08-13
fix(orchestrator):filter last run in wotkflows tab by entity

LiorSoffer:
resolve [FLPATH-2633](https://issues.redhat.com/browse/FLPATH-2633)
plus updating the template to prompt entity name

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts
@@ -431,8 +431,13 @@ export class DataIndexService {
     definitionId: string;
     limit: number;
     offset: number;
+    targetEntity?: string;
   }): Promise<ProcessInstance[]> {
-    const graphQlQuery = `{ ProcessInstances(where: {processId: {equal: "${args.definitionId}" } }, orderBy: {start:DESC}, pagination: {limit: ${args.limit}, offset: ${args.offset}}) { id, processName, state, start, end } }`;
+    const targetEntityWhereCondition = args.targetEntity
+      ? `, variables: {targetEntity: {equal: "${args.targetEntity}" } }`
+      : '';
+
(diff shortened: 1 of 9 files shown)


workspaces/orchestrator/plugins/orchestrator-backend/src/service/DataIndexService.ts
```diff
@@ -431,8 +431,11 @@ export class DataIndexService {
     definitionId: string;
     limit: number;
     offset: number;
+    targetEntity?: string;
   }): Promise<ProcessInstance[]> {
-    const graphQlQuery = `{ ProcessInstances(where: {processId: {equal: "${args.definitionId}" } }, orderBy: {start:DESC}, pagination: {limit: ${args.limit}, offset: ${args.offset}}) { id, processName, state, start, end } }`;
+    const graphQlQuery = args.targetEntity
+      ? `{ ProcessInstances(where: {processId: {equal: "${args.definitionId}" }, variables: {targetEntity: {equal: "${args.targetEntity}" } } }, orderBy: {start:DESC}, pagination: {limit: ${args.limit}, offset: ${args.offset}}) { id, processName, state, start, end } }`
```
batzionb, line 437: can you do something like this instead to avoid duplication


const where = targetEntity
  ? `{ processId: { equal: "${definitionId}" }, variables: { targetEntity: { equal: "${targetEntity}" } } }`
  : `{ processId: { equal: "${definitionId}" } }`;

const fields = `id, processName, state, start, end`;

const graphQlQuery = `
  {
    ProcessInstances(
      where: ${where},
      orderBy: { start: DESC },
      pagination: { limit: ${limit}, offset: ${offset} }
    ) { ${fields} }
  }
`;

LiorSoffer: done (similar way)

batzionb (approved):

PR #1297, opened 2025-08-13
chore(orchestrator):Add readme for workflows tab

LiorSoffer:
resolve [FLPATH-2329](https://issues.redhat.com/browse/FLPATH-2329)

### DIFF
--- a/workspaces/orchestrator/docs/workflows-tab.md
+++ b/workspaces/orchestrator/docs/workflows-tab.md
@@ -1,0 +1,124 @@
+# ⚙️ Workflows tab for Catalog Entities
+
+## 📌 Overview
+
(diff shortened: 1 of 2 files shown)


workspaces/orchestrator/docs/workflows-tab.md
```diff
@@ -0,0 +1,90 @@
...
+> When no workflows are listed, the tab will display workflows inferred from historical runs.
+
+---
+
+### **2️⃣ Configure Dynamic Plugins**
+
+The `entityTabs` and `mountPoint` fields are required for this catalog integration.
+
+```yaml
+dynamicPlugins:
```
batzionb, line 38: we already have this here
https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/docs/dynamic-plugin-installation.md
Do we want to duplicate this?
Maybe add a link

LiorSoffer: I think we should duplicate it,  just in case  :)

PR #1381, opened 2025-08-29
fix(orchestrator): remove workflow diagram (CSP issues)

mareklibra:
Workaround for FLPATH-1327 for the case when there is no fix for https://github.com/apache/incubator-kie-tools/issues/3197 in the RHDH 1.7 timeframe.

This patch should be easily reverted later, when a solution/fix is found.

---
Related to: https://github.com/redhat-developer/rhdh-plugins/pull/1374


mareklibra:
Screenshot From 2025-08-29 09-54-40

PR #1384, opened 2025-09-01
fix(orchestrator): use CodeSnippet instead of kie-tools editor

mareklibra:
Fixes FLPATH-1327 by replacing kie-tools viewer by CodeSnippet.

We are loosing the read-only diagram but
- resolving CSP issues
- significantly reducing the bundle size
- leveraging well-tested and supported component
- avoiding complicated integration of a 3rd party component

Considering the component is shown to administrators only, such simplification should be acceptable.

---
This PR would obsolete
- https://github.com/redhat-developer/rhdh-plugins/pull/1381
- https://github.com/redhat-developer/rhdh-plugins/pull/1374


mareklibra:
Screenshot From 2025-09-01 10-29-41

PR #1506, opened 2025-09-24
feat(translations): add translations backend and download PR

karthikjeeyar:
## Translations backend

Fixes:
https://issues.redhat.com/browse/RHIDP-8892

This PR introduces
1. Added Translations backend to load translations keys from the overrides files provided via app-config.
2. Added Download translation keys to download all the english default messages.
3. Translated the frontend components.

image

french:

image


ciiay (commented):
Tested on the dev app and it works as expected. I noticed that when I switch to another language the downloaded translation file is still named in English. Not sure if that's what we want or should we align the file name with the selected language? After all the download button got translated. 

Also left a small question to the authentication gating.

workspaces/translations/plugins/translations-backend/src/plugin.ts
```diff
@@ -0,0 +1,44 @@
...
+      deps: {
+        http: coreServices.httpRouter,
+        config: coreServices.rootConfig,
+        logger: coreServices.logger,
+      },
+      async init({ http, config, logger }) {
+        http.use(await createRouter({ config, logger }));
+        http.addAuthPolicy({
+          path: '/',
+          allow: 'unauthenticated',
```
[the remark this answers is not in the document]
karthikjeeyar: This API is not for downloading translation keys, this is to get all the overrides json file content that customer has provided + our internal JSON files. At the time of App initialization we will not have the token to pass so this API is ungaurded.

karthikjeeyar:
@debsmita1 Since this is rhdh-plugins repo, and since I only have `translationResource` here in this workspaces I have passed it to showcase the download behaviour, when this is used in RHDH (my next PR) where I will pass all the translationResources (static + dynamic resources in RHDH) so you will get to see all the plugins keys in RHDH.

workspaces/translations/plugins/translations-backend/src/service/TranslationService.ts
```diff
@@ -0,0 +1,315 @@
...
+
+    if (
+      fs.existsSync(targetFolderPath) &&
+      fs.statSync(targetFolderPath).isDirectory()
+    ) {
+      topmostFoundPath = targetFolderPath;
+    }
+
+    return topmostFoundPath;
+  }
```
christoph-jerolimov, line 92: I'm not a fan of this "magic" but its fine for now. I like to discuss it next week.

I would prefer that we configure these paths in the app-config.

karthikjeeyar: Sure, we can improve this next week in a follow up PR.

PR #1523, opened 2025-09-26
fix(scorecard): Introduce unified config and validation

dzemanov:

- Introduces provider validation
- Introduces unified scorecard backend config - useful when we add schedulers to scorecard backend
- Changes jira and github provider ids to match their configurations

#### Fixes
Fixes http://issues.redhat.com/browse/RHDHBUGS-2085


Eswaraiahsapram (approved):
Thanks @dzemanov , tested in local working fine

/lgtm

Screenshot 2025-09-29 at 10 08 00 AM

christoph-jerolimov (changes requested):
Hi, in general this is fine. But can you please update the translations in `plugins/scorecard/src/translations/ref.ts` and all languages files in the same folder as well?

Thank you

christoph-jerolimov (approved):
I'm confused why the provider id should be same as metric id. Maybe these will not be 1:1 in the future, but we can discuss that (if you want) when we continue with this plugin. For now this okay / good to me.

Tested it locally (rebased it on main to see by Jira count and the other plugins):

image

/lgtm
/approve

PR #1527, opened 2025-09-26
RHDHPAI-1019: make ai-integrations/model catalog plugins public

gabemontero:

setting private to false in package.json .... per discussions with @johnmcollier this will be needed when we start publishing the model catalog plugins over in https://github.com/redhat-developer/rhdh-plugin-export-overlays

waiting an seeing if leaving the ai-experience plugins as private == true is allowed 

if not @rohitkrai03 @karthikjeeyar we may have to talk about removals since we are not moving forward with this as I understand it

- [n/a ] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ n/a] Added or Updated documentation
- [ n/a] Tests for new functionality and regression tests for bug fixes
- [ n/a] Screenshots attached (for UI changes)


johnmcollier:
@gabemontero Just re-run `yarn install` and commit changes - the linter will remove the `private: false` entry since it's not needed (no entry is the same as setting `private` to false)

PR #1533, opened 2025-09-27
fix(orchestrator): Simplify Workflow Instance page header

mareklibra:
Simplify Workflow Instance page header based on customer feedback.

Before:

before

---

After:

after


ShiranHi:
> Is this so page will have the ID for copy paste? Maybe putting it in the details is better and improving the header to "Instance page" Or maybe the header can be "Workflow Instance " @ShiranHi WDYT?

I like the "After" solution proposed in this PR. And yes, we can move the ID to be part of the Details card and add a copy icon next to it (see [here](https://www.figma.com/design/XjdTCb4vDQ7L4Jz1TdHKGq/Workflow-Orchestrator?node-id=4896-5438&t=ayNHNWuocxliLEtf-1)).

We also discussed a naming convention update for workflow runs header: we should add sequential numbers to the run names. For example, the first run of the "Move 2Kube workflow" should be displayed as "001-Move 2Kube workflow", the next run will be "002-Move 2Kube workflow", etc...

batzionb:
approved but CI should pass

mareklibra:
Converting to draft, I will add the Run ID among the Details.

PR #1395, opened 2025-09-05
feat(marketplace): integrate plugins-info plugin

ciiay:
## Hey, I just made a Pull Request!

For [RHIDP-8301](https://issues.redhat.com/browse/RHIDP-8301)
This works with [rhdh pr #3366](https://github.com/redhat-developer/rhdh/pull/3366)

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ ] Added or Updated documentation
- [x] Tests for new functionality and regression tests for bug fixes
- [x] Screenshots attached (for UI changes)

Test config:
```
dynamicPlugins:
  rootDirectory: dynamic-plugins-root
  frontend:
    default.main-menu-items:
      menuItems:
        ...
        default.admin:
          title: Administration
          icon: admin
...
    red-hat-developer-hub.backstage-plugin-marketplace:
      appIcons:
        - name: pluginsIcon
          importName: PluginsIcon
      dynamicRoutes:
        - path: /extensions
          importName: DynamicMarketplacePluginRouter
          menuItem:
            icon: pluginsIcon
            text: Extensions
      menuItems:
        extensions:
          parent: default.admin
```

Screen recording of testing on RHDH
Updated on 9/19 (with [rhdh pr #3366](https://github.com/redhat-developer/rhdh/pull/3366)) 

https://github.com/user-attachments/assets/63a3b166-3466-4871-a8b1-eec7687393e5

PR #1558, opened 2025-10-01
feat(extensions): add row actions to installed packages

debsmita1:

Resolves: https://issues.redhat.com/browse/RHIDP-6810

- Addressed the review comments received here https://github.com/redhat-developer/rhdh-plugins/pull/1395
- Fixed the package info drawer to keep the Actions dropdown consistent with the plugin info drawer
- Implemented the row actions
- Handled the RBAC permissions to modify package configuration
- Added translations where ever required
- Added and updated tests

Note: 
- The API to Enable/Disable a package is utilized in the UI, but the functionality needs to be fixed.
- The `Uninstall` action will be worked on in a different PR

Screenshots:

Screenshot 2025-10-01 at 10 27 40 PM

https://github.com/user-attachments/assets/9a1da3c8-c819-4ced-8614-f1dc493db33c


ShiranHi:
Thank you @debsmita1 , looks great!
Could you add a screenshot that shows the tooltip text for the icons?

ciiay (commented):
Hi @debsmita1 , thanks for the PR. I finally got it working with some specific commit. When I use marketplace backend plugin exported from this PR with current `rhidp-8301-remove-dynamic-plugins-info-plugin` branch version RHDH, the marketplace plugin doesn't load any data like this
image
image

Then I tried some old commits, which worked together to test changes in this PR, here's the version I used.
image
image
It seems that the issue is with the backstage version bump in both marketplace backend plugin and the RHDH. After the backstage version bump, the backend stops loading.

Here's the recording for the UI changes in this PR:

https://github.com/user-attachments/assets/432cf362-b45a-4569-b800-401b2399faf2

Things I observed:
1. when hover on Edit button, the highlight effect is not a circle but a oral shape(low priority)
2. clicking on `Cancel` button on package edit page takes the user to the drawer, instead of going directly back to the installed packages table.
3. tooltip for action buttons only shows up when buttons are disabled, and the message seems off, it should be the permission hint instead of the button fuction.
4. `Disable` toggle button not working, will this be handled in another story?

workspaces/marketplace/plugins/marketplace/package.json
```diff
@@ -63,6 +63,7 @@
     "@red-hat-developer-hub/backstage-plugin-marketplace-common": "workspace:^",
     "@scalprum/react-core": "0.9.5",
     "@tanstack/react-query": "^5.60.5",
+    "js-yaml": "^4.1.0",
```
christoph-jerolimov, line 66: I see in the diff that we already use the yaml package, can we just use that?

workspaces/marketplace/plugins/marketplace/src/components/MarketplacePluginContent.tsx
```diff
@@ -255,7 +275,7 @@ const PluginPackageTable = ({ plugin }: { plugin: MarketplacePlugin }) => {
 export const MarketplacePluginContent = ({
   plugin,
 }: {
-  plugin: MarketplacePlugin;
+  plugin: MarketplacePlugin | MarketplacePackage;
```
christoph-jerolimov, line 278: Naaa? Why is this needed? We had a MarketplacePackageContent before? I expect we lose a lot of type save-tyness if we say "this can be a package as well"

ciiay:
Hi @debsmita1 , thanks for the updates. It's almost working perfect now except for the "disable/enable" toggle button which doesn't do anything when I click on it with permission. When I don't have the permission the tooltips are showing two kinds of messages, is this expected?

https://github.com/user-attachments/assets/8cae0c1f-1e31-45ec-824f-6b59df1a211f

PR #1571, opened 2025-10-04
fix(orchestrator): let AuthRequester reference static non-standard auth apis

mareklibra:
Jira: FLPATH-2361

With this patch, the AuthRequester widget can reference statically added non-core auth apis (not well-names like GitHub, not provided by a plugin).

Example, the RHDH OIDC API: https://github.com/redhat-developer/rhdh/blob/main/packages/app/src/api/AuthApiRefs.ts#L18

Workflow data input schema example:

```
"authSetup": {
  "type": "string",
  "ui:widget": "AuthRequester",
  "ui:props": {
  "authTokenDescriptors": [
    {
      "provider": "oidc",
      "customProviderApiId": "internal.auth.oidc",
      "tokenType": "oauth"
    }
  ]}
}
```

PR #1572, opened 2025-10-04
fix(orchestrator): backport: let AuthRequester reference static non-standard auth apis

mareklibra:
Backport: #1571


mareklibra:
Merging before #1571 is properly reviewed, so we can have a build to unblock testing sooner.

PR #1720, opened 2025-11-04
fix(orchestrator): endless onChange() loop and turning ui:allowNewItems from string to boolean type

mareklibra:
Multiple small changes at once:
- fix for an infinite onChange() loop by `ActiveDropdown()`.
- use generic `readOnly` instead of proprietary `ui:readonly` - https://json-schema.org/understanding-json-schema/keywords#readonly
- the `ui:allowNewItems` property became boolean (instead of string)

---
Example of use: https://github.com/rhdhorchestrator/orchestrator-demo/pull/72


gciavarrini (approved):
lgtm

PR #1741, opened 2025-11-07
RHIDP-9653: Add scoped resolution for refractor/prismjs

alizard0:
It adds package.json resolution for refractor/prismjs


rohitkrai03 (approved):
/lgtm
/approve

PR #1830, opened 2025-11-21
chore(e2e): quickstart e2e with translations and accessibility tests

teknaS47:

Added e2e test for Quick start plugin with translations and accessibility tests. [RHIDP-9440](https://issues.redhat.com/browse/RHIDP-9440)

### DIFF
--- a/workspaces/global-header/packages/app/e2e-tests/globalHeader.test.ts
+++ b/workspaces/global-header/packages/app/e2e-tests/globalHeader.test.ts
@@ -38,6 +38,7 @@ test.beforeAll(async ({ browser }) => {
     () => globalThis.navigator.language,
   );
   await page.goto('/');
+  await page.waitForTimeout(2000);
   await page.getByRole('button', { name: 'Enter' }).click();
 
   // Extract base language code (e.g., "en" from "en-US")


workspaces/quickstart/packages/app/e2e-tests/utils/translations.ts
```diff
@@ -0,0 +1,58 @@
...
+    const lastIndex = path.length - 1;
+    path.reduce((acc, currentPath, i) => {
+      (acc as any)[currentPath] =
+        lastIndex === i
+          ? (messages as any)[key]
+          : (acc as any)[currentPath] || {};
+      return (acc as any)[currentPath];
+    }, res);
+    return res;
+  }, {} as any);
```
jrichter1: if you're trying to keep it type-safe, something like this would make more sense
```typescript
function transform(messages: typeof quickstartTranslationDe.messages) {
  const result = Object.keys(messages).reduce((res, key) => {
    const path = key.split('.');
    const lastIndex = path.length - 1;
    path.reduce((acc: Record<string, string|{}>, currentPath, i) => {
      acc[currentPath] =
        lastIndex === i ? (messages as Record<string, string>)[key] : acc[currentPath] || {};
      return acc[currentPath];
    }, res);
    return res;
  }, {});

  return result as QuickstartMessages;
}
```

workspaces/global-header/packages/app/e2e-tests/globalHeader.test.ts
```diff
@@ -148,6 +147,7 @@ test('Verify Search functionality and results', async () => {
   const expectedUrl = /\/example-grpc-api/;
 
   await search.fill(searchQuery);
+  await page.waitForTimeout(5000);
```
jrichter1: This looks like a leftover, can you remove it?

PR #1958, opened 2025-12-15
fix: Validation errors incorrectly shown on wrong step when navigating back

upendhran:

Fixes a bug where validation errors from other steps would incorrectly appear on the current step when using widgets with `validate:url` in multi-step workflow forms.

#### Problem
When navigating between steps in a multiple step workflow:

1. User is on step 2 and triggers validation
2. User clicks "Back" to return to step 1
3. Makes a change on step 1 and proceeds forward
4. step 1 incorrectly shows validation errors about step 2

This issue occurs specifically with workflows that use widgets with `validate:url`, which triggers the `getExtraErrors` callback. This callback validates all fields across all steps and returns an error object.

#### Root Cause

The `getExtraErrors` callback validates all fields across all steps and returns a nested error object by step. When filtering errors for the current step, the fallback logic incorrectly set the entire error object when the current step had no errors.

#### Fix
- When current step has no errors, set `extraErrors` to `undefined` instead of the full error object
- Update navigation logic to only evaluate the current step's errors before proceeding

#### Screenshots

Screenshot 2025-12-16 at 12 11 44 PM

In the first step, the user types `test` app registration name and hits next.

Screenshot 2025-12-16 at 12 11 55 PM

In the second step, the user is not entering the namespace name, and goes back.

Screenshot 2025-12-16 at 12 12 09 PM

From the first step, the user is trying to go to the second step again, and sees the validation errors.

### DIFF
--- a/workspaces/orchestrator/.changeset/fix-extra-errors-wrong-step.md
+++ b/workspaces/orchestrator/.changeset/fix-extra-errors-wrong-step.md
@@ -7,5 +7,6 @@ Fix validation errors incorrectly shown on wrong step when navigating back.
 When using widgets with `validate:url`, the `getExtraErrors` callback validates all fields across all steps and returns a nested error object. The previous logic had full error object when the current step had no errors, causing validation errors from other steps to appear on the wrong step.
 
 This fix:
+
 - Sets `extraErrors` to `undefined` when current step has no errors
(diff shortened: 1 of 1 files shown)


upendhran:
@lokanandaprabhu Fixed the CI issue. Thanks

PR #1967, opened 2025-12-17
chore: resolve CVE-2025-57319 by upgrading pino in all workspaces

alizard0:
It fixes CVE-2025-57319 by upgrading (resolution) pino to v10.1.0 in all workspaces.


christoph-jerolimov:
@alizard0 you need to run yarn install in all updated workspaces

kim-tsao (commented):
The CVE has been [withdrawn](https://github.com/advisories/GHSA-ffrw-9mx8-89p8) and the update is also a major version bump which could cause issues.  I think we should close this PR

PR #1985, opened 2025-12-24
feat(orchestrator): add fetch:error:ignoreUnready and fetch:response:default options for form widgets

lokanandaprabhu:

Fixes:
https://issues.redhat.com/browse/RHDHBUGS-2454
https://issues.redhat.com/browse/RHDHBUGS-2453

This PR introduces two new ui:props options for the orchestrator form widgets to improve user experience when working with dynamic forms that depend on fetch operations.

--------------------------

**Feature 1**: fetch:error:ignoreUnready

**Problem**: When using widgets with fetch:retrigger dependencies, the initial fetch often fails because dependent fields don't have values yet. This results in HTTP errors (e.g., 404, 400) being displayed to users during initial load, even though this is expected behavior.

**Solution**: Added fetch:error:ignoreUnready option that suppresses fetch error display until all fetch:retrigger dependencies have non-empty values.

**Behavior**:

- When true, errors are hidden while dependencies are empty
- Once all dependencies have values, errors are shown normally
- Supported by: ActiveTextInput, ActiveDropdown, ActiveMultiSelect, SchemaUpdater

--------------------------

**Feature 2**: fetch:response:default

**Problem**: Widgets previously required fetch:response:value for default values, meaning a fetch must succeed for any default to be applied. 

This was problematic when:

- You want a static fallback default when fetch fails or hasn't completed
- You want to pre-populate a field without requiring a fetch
- The fetch depends on fields that aren't filled yet

**Solution**: Added fetch:response:default option for static default values applied immediately on form initialization.

**Key Implementation Details**:

- Defaults are extracted from schema and applied at form initialization level (in OrchestratorForm), not inside widgets
- This ensures controlled components work correctly and values persist across wizard step navigation
- Static defaults act as fallback when fetch fails, hasn't completed, or returns empty
- Fetched values only override defaults when non-empty
- Loading spinners are skipped when static defaults exist for instant display

**Supported Types**:
ActiveTextInput: string (e.g., "create")
ActiveDropdown: string (e.g., "default-profile")
ActiveMultiSelect: string[] (e.g., ["tag1", "tag2"])

--------------------------

Orchestrator - fetch error skip and resp default

--------------------------

### DIFF
--- a/workspaces/orchestrator/.changeset/fetch-error-skip-and-response-default.md
+++ b/workspaces/orchestrator/.changeset/fetch-error-skip-and-response-default.md
@@ -3,13 +3,13 @@
 '@red-hat-developer-hub/backstage-plugin-orchestrator-form-widgets': minor
 ---
 
-Add fetch:error:skip and fetch:response:default options for form widgets
+Add fetch:error:ignoreUnready and fetch:response:default options for form widgets
 
-**Feature 1: fetch:error:skip**
+**Feature 1: fetch:error:ignoreUnready**
 
 When using widgets with `fetch:retrigger` dependencies, the initial fetch often fails because dependent fields don't have values yet. This results in HTTP errors being displayed during initial load.
 
-- Add `fetch:error:skip` option to suppress fetch error display until all `fetch:retrigger` dependencies have non-empty values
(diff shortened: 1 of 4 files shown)


mareklibra:
The `fetch:error:skip` name can be confusing, what about `fetch:error:ignoreUnready`?

The point is to make clear that the error is skipped only at the beginning of the flow, not later when either data or server issues occur.

PR #2034, opened 2026-01-12
Fix Color of Arrow Down / Up

asmasarw:
Fixed Arrows Colors
We were Rendering the opposite colors

image


PreetiW (approved):
good catch 👍🏻

PR #2035, opened 2026-01-12
fix(scorecard): Scorecard aggregated metrics UI integration

Eswaraiahsapram:

Fixes: 
1. https://issues.redhat.com/browse/RHIDP-11157
2. https://issues.redhat.com/browse/RHIDP-10985

### Description

- Integrated Backend Aggregated metrics API with Frontend
- In the existing entity scorecard component, we started using `recharts` pie chart
- Added Missing permission for Aggregated metrics

### Screen Recording 

https://github.com/user-attachments/assets/7653f30a-835d-4d0c-a11e-c832263bdca9

### DIFF
--- a/workspaces/scorecard/packages/app/e2e-tests/pages/CatalogPage.ts
+++ b/workspaces/scorecard/packages/app/e2e-tests/pages/CatalogPage.ts
@@ -40,6 +40,6 @@ export class CatalogPage {
     await page.getByRole('link', { name: 'Settings' }).click();
     await page.getByRole('button', { name: 'English' }).click();
     await page.getByRole('option', { name: locale }).click();
-    await page.locator('a').filter({ hasText: 'Home' }).click();
+    await page.locator('a').filter({ hasText: 'Catalog' }).click();
   }
 }
--- a/workspaces/scorecard/packages/app/e2e-tests/utils/rbacSetup.ts
+++ b/workspaces/scorecard/packages/app/e2e-tests/utils/rbacSetup.ts
@@ -57,7 +57,7 @@ export async function setupRBAC(page: Page) {
   await page.getByTestId('nextButton-2').click();
   await page.getByRole('button', { name: 'Create' }).click();
 
-  await page.locator('a').filter({ hasText: 'Home' }).click();
+  await page.locator('a').filter({ hasText: 'Catalog' }).click();
   await expect(page.getByRole('button', { name: 'Create' })).toBeVisible();
 }
 async function selectCheckbox(page: Page, label: string) {


workspaces/scorecard/plugins/scorecard/src/components/Scorecard/CustomLegend.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  if (
+    !thresholds ||
+    thresholds?.definition?.rules?.length === 0 ||
+    thresholds?.definition?.rules === undefined
+  ) {
+    return (
+      <StyledLegend>
+        <StyledLegendItem>
+          <StyledLegendColorBox color={theme.palette.grey['400']} /> --
```
imykhno: ??

workspaces/scorecard/plugins/scorecard/src/components/Scorecard/Scorecard.tsx
```diff
@@ -80,225 +64,188 @@ const Scorecard = ({
...
-                      sx={{
-                        display: 'flex',
-                        alignItems: 'center',
-                        gap: 1,
-                      }}
-                    >
-                      <Box
+                return (
+                  <g transform={`translate(${cx}, ${cy})`}>
+                    <foreignObject x={-12} y={-28} width={24} height={24}>
```
imykhno: ??

workspaces/scorecard/plugins/scorecard/src/components/ScorecardHomepageSection/ResponsivePieChart.tsx
```diff
@@ -0,0 +1,84 @@
...
+  Tooltip,
+  TooltipProps,
+} from 'recharts';
+import { PieData } from '../../utils/utils';
+
+interface ResponsivePieChartProps {
+  pieData: PieData[];
+  isMissingPermission?: boolean;
+  LabelContent?: any;
+  legendContent: (props: any) => React.ReactNode;
```
imykhno: ?

workspaces/scorecard/plugins/scorecard/src/components/ScorecardHomepageSection/__tests__/ScorecardHomepageSection.test.tsx
```diff
@@ -17,160 +17,159 @@
...
+    title: 'GitHub open PRs',
+    description: 'Open PR count',
+    type: 'number',
+    history: true,
+  },
+  result: {
+    total: 8,
+    values: [{ name: 'success', count: 8 }],
+    timestamp: '2024-01-01T00:00:00Z',
+  },
```
imykhno: Value is correct?

dzemanov (approved):
I tested the PR with postgres and sqlite.

https://github.com/user-attachments/assets/905efdb9-d3e5-4bf1-a2e8-58c08dcf2487


I wasn't able to reproduce No entities state, looks like we are returning [] from backend.
I was able to get Missing permission when all permissions missing or only for some scorecards with conditional permission, like here: 
Screenshot 2026-01-14 at 15 25 10

Translations worked fine for me.
Screenshot 2026-01-14 at 14 59 12
Screenshot 2026-01-14 at 14 52 59

When there is error in thresholds or in fetching metric, backend doesn't return the entity's result. When all entities have errors, backend returns [] and it leaves these empty spaces:
Screenshot 2026-01-14 at 15 18 07

Maybe it would be good to return response also for 0 entities from backend, at least with metric metadata so it can be shown or the no entities state can be shown. If all entities have errors, it could help frontend if we return response with "status": "error".

workspaces/scorecard/plugins/scorecard/src/components/ScorecardHomepageSection/ResponsivePieChart.tsx
```diff
@@ -0,0 +1,84 @@
...
+  tooltipContent: (props: TooltipProps<number, string>) => React.ReactNode;
+}
```
imykhno: I noticed that `any` was used in this interface, should we define some type instead?

imykhno:
I’ve tested the same scenarios locally as Dominika and observed the same results. However, when a user does not have access to the catalogue, the error shown in the screenshot below is displayed. Is this the expected behavior?

Screenshot 2026-01-14 at 16 53 17

imykhno:
One more point regarding **SonarQube**. I think it would be great if we could resolve or at least minimize the issues highlighted in the report.

christoph-jerolimov:
> > I’ve tested the same scenarios locally as Dominika and observed the same results. However, when a user does not have access to the catalogue, the error shown in the screenshot below is displayed. Is this the expected behavior?
> 
> My understanding is that the interface should dynamically hide the card for any user who lacks authorized access to the related catalog entities.

Hi @ShiranHi, the homepage doesn't support this. We have a feature for the homepage to add RBAC checks to it. We will have that with the next release.

dzemanov (approved):

PR #2049, opened 2026-01-12
chores: fixes CVE-2025-64756 by upgrading glob dependency on lightspeed workspace

alizard0:
Upgrades glob dependency for lightspeed workspace


christoph-jerolimov (approved):
We have tests and they passes, so we should be able to merge. If this break, we need to improve the tests. :)

PR #2097, opened 2026-01-16
chore(e2e): add Italian (it) and Japanese (ja) locale tests

HusneShabbir:
## Summary
Add Italian (`it`) and Japanese (`ja`) locale configurations to Playwright e2e tests across all workspaces that support i18n testing.

Which issue(s) does this PR fix
Fixes : [RHIDP-11392](https://issues.redhat.com/browse/RHIDP-11392)

## Changes
Updated `playwright.config.ts` in 9 workspaces:
- adoption-insights
- bulk-import
- extensions
- global-floating-action-button
- global-header
- homepage
- lightspeed
- quickstart
- scorecard

Each workspace now runs e2e tests in 4 locales: `en`, `fr`, `it`, `ja`.

### DIFF
--- a/workspaces/adoption-insights/packages/app/e2e-tests/insights.test.ts
+++ b/workspaces/adoption-insights/packages/app/e2e-tests/insights.test.ts
@@ -248,21 +248,19 @@ test.describe(() => {
   test('New data shows in searches', async () => {
     const panel = getPanel(
       page,
-      new RegExp(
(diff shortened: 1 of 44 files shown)


jrichter1:
Given the label for Japanese is in Japanese, you're going to have trouble matching it with the locale string. We might have to use a bit more sophisticated system if we keep adding languages that use different alphabets.

PR #2099, opened 2026-01-16
fix(lightspeed): Preserve file attachments when changing display mode

its-mitesh-kumar:
## Summary
Fixes an issue where attached files were lost when changing the display mode (overlay → docked → window), while the text input was correctly retained.

## Fixes
- Fixed https://issues.redhat.com/browse/RHDHBUGS-2491
## Problem
When the user attaches a file and then changes the display mode:
- The text input was preserved ✓
- The file attachment was lost ✗

This happened because:
- `draftMessage` was stored in `LightspeedDrawerContext` (persists across mode changes)
- `fileContents` was stored in `FileAttachmentContext` inside `LightspeedChatContainer` (remounts on mode change)

## Solution
Added `draftFileContents` and `setDraftFileContents` to `LightspeedDrawerContext`, following the same pattern as `draftMessage`. The `FileAttachmentContext` now:
1. Initializes `fileContents` from `draftFileContents` on mount
2. Syncs changes back to `draftFileContents` via useEffect

## UI before changes

https://github.com/user-attachments/assets/815ae1c4-a1ee-48e8-b6b4-a9a78157ff84

## UI after changes 
https://github.com/user-attachments/assets/745b6cf6-3d20-4c75-bea9-9dbc271ebbf3

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed/src/components/__tests__/LightspeedPage.test.tsx
+++ b/workspaces/lightspeed/plugins/lightspeed/src/components/__tests__/LightspeedPage.test.tsx
@@ -34,6 +34,13 @@ jest.mock('@backstage/plugin-permission-react', () => ({
   RequirePermission: jest.fn(),
 }));
 
+jest.mock('../../hooks/useLightspeedDrawerContext', () => ({
+  useLightspeedDrawerContext: () => ({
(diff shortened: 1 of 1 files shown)


rohitkrai03 (approved):
Code looks good to me.

PR #2112, opened 2026-01-20
[Konflux] prepare first release

rrosatti:
## Description

In this PR we're adding a `changeset` and "reseting" Konflux package versions to `"0.0.0"` in preparation for the first release ☕ 


workspaces/konflux/package.json
```diff
@@ -1,7 +1,6 @@
 {
   "name": "@internal/konflux",
-  "version": "0.1.0",
-  "private": true,
```
rrosatti: curious question: is it correct to remove the `"private": true` from `workspaces/konflux/package.json` (above change)? if not, I will undo it :)

PR #2116, opened 2026-01-20
FLPATH-2322 | When Apply Recommendations button is disabled there is …

asmasarw:
FLPATH-2322 | When Apply Recommendations button is disabled there is no tooltip explaining why

image

image

image


PreetiW (approved):
LGTM 🎉

PR #2182, opened 2026-01-27
feat(lightspeed): Add Escape key to cycle through display modes

its-mitesh-kumar:
## Description
- Adds Escape key functionality to cycle through chatbot display modes
- Pressing Escape transitions: Fullscreen → Docked → Overlay → Close (FAB only)
- Settings dropdown closes on first Escape; display mode cycles on subsequent presses
- Works correctly across all display modes without breaking existing functionality

## UI after changes

https://github.com/user-attachments/assets/b916703b-b6ac-466b-a016-546d85a25c55

## Fixes
- Fixed https://issues.redhat.com/browse/RHDHBUGS-2490


HusneShabbir (commented):
@aprilma419 what’s your take on the above observation?

debsmita1:
I agree with @karthikjeeyar. Let’s avoid over-complicating the Esc key behavior. Since display mode selection is now persisted, cycling through display modes on Esc would unnecessarily complicate that logic. Instead, pressing Esc should first close any open modal or the Lightspeed settings dropdown; pressing it again should close the chatbot

cc @aprilma419

PR #2188, opened 2026-01-28
chore(x2a): Add GitHub auth provider

mareklibra:
Fixes: FLPATH-3085

Depends on (needs rebase):
- [x] #2178 

Adding configuration to enable GitHub auth provider in our DEV environment.

To use it, make sure to set following environment variables:
```
export AUTH_GITHUB_CLIENT_ID=....
export AUTH_GITHUB_CLIENT_SECRET=...
yarn dev
```

There was attempt to add RBAC but this will need more work since `@backstage/plugin-permission-backend` fails to load.
This will be fixed in a follow-up, some debugging insight is included.


workspaces/x2a/entities/example-rbac-policy.csv
```diff
@@ -0,0 +1,7 @@
+p, role:default/x2aAdmin, x2a.admin, read, allow
```
mareklibra: So far just an example.

The dev instance will be configured in a follow-up, failing in the mocked dev Backstage instances used so far.

elai-shalev (approved):

PR #2250, opened 2026-02-04
chore(lightspeed): Removed resolutions 

alizard0:
It removed the resolutions added by https://github.com/redhat-developer/rhdh-plugins/pull/1151/


karthikjeeyar (commented):
CI fails with Typescript errors, to fix it you have to remove `/alpha` from this backend/src/index.ts - https://github.com/redhat-developer/rhdh-plugins/blob/2fc6542c88621ed8e05834145465e19074207559/workspaces/lightspeed/packages/backend/src/index.ts#L23-L24

PR #2266, opened 2026-02-06
feat(x2a): Enable backend application for collectartifacts

eloycoto:
This is the initial implementation on the collectArtifacts, there are a few things missing(Marked as TODO), but I think that it'll be good to have a bootstrap already, and improved in different steps like security and events on job finished.


workspaces/x2a/plugins/x2a-common/client/src/schema/openapi/generated/models/Telemetry.model.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  summary: string;
+  /**
+   * Phase name (init, analyze, migrate, publish)
+   */
+  phase: string;
+  /**
+   * When the phase started (ISO 8601 format)
+   */
+  startedAt: Date;
```
mareklibra: This can be retrieved from the jobs table. From the UI perspective, the job is not finished until its status in the DB table says so. And that's the time when we can mark the finishedAt as well.

mareklibra (commented):
Let's discuss the use-case for this from the RHDH user perspective.

The inner subflows metrics (means per agent) sound interesting especially for us but what the end-user can do about them?

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
```diff
@@ -0,0 +1,351 @@
...
+        value: JSON.stringify(artifacts.telemetry),
+      });
+    }
+
+    // Serialize external links
+    if (artifacts.externalLinks) {
+      for (const [name, url] of Object.entries(artifacts.externalLinks)) {
+        result.push({
+          id: randomUUID(),
+          type: 'externalLink',
```
mareklibra: I would rather stick to more specific types via `ArtifactType` modeled even in the OpenApi spec.
The tool handles them explicitly, we have more-or-less finite set of those.

workspaces/x2a/plugins/x2a-backend/src/schema/openapi/generated/router.ts
```diff
@@ -550,6 +550,110 @@ export const spec = {
...
+                  "error": {
+                    "type": "string",
+                    "description": "Error message if status is Error"
+                  },
+                  "jobId": {
+                    "type": "string",
+                    "description": "UUID of the completed job"
+                  },
+                  "artifacts": {
+                    "$ref": "#/components/schemas/PhaseArtifact"
```
mareklibra: There is already `Artifact` component

workspaces/x2a/plugins/x2a-backend/src/router.ts
```diff
@@ -701,19 +703,33 @@ export async function createRouter({
...
+    async (req: express.Request, res: express.Response) => {
+      const endpoint = 'POST /projects/:projectId/collectArtifacts';
+      const { projectId } = req.params;
+      const { moduleId, phase } = req.query;
+
+      logger.info(
+        `${endpoint} request received: projectId=${projectId}, moduleId=${moduleId}, phase=${phase}`,
+      );
+
+      const handler = new CollectArtifactsHandler({
```
mareklibra: Let's be careful when rebasing here - this contract is different.

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
```diff
@@ -0,0 +1,209 @@
...
+      }
+    },
+  );
+}
+
+function validateRequest(
+  requestBody: unknown,
+  queryParams: { moduleId?: string; phase: MigrationPhase },
+): CollectArtifactsRequestBody {
+  if (queryParams.phase === 'init' && queryParams.moduleId !== undefined) {
```
mareklibra: This is overlapping with the test in the post-handler above

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
```diff
@@ -0,0 +1,209 @@
...
+          );
+        }
+
+        if (job.projectId !== projectId) {
+          throw new NotFoundError(
+            `Job ${validatedRequest.jobId} does not belong to project ${projectId}`,
+          );
+        }
+
+        if (job.phase !== phase) {
```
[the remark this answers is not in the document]
eloycoto: Thinking that the x2a job is a bash script, do you want to take the risk that the user request phase is not the same as what It should?

mareklibra (approved):

PR #2279, opened 2026-02-09
fix(orchestrator-form-widgets): show spinner immediately on ActiveText retrigger

lokanandaprabhu:
#### Fixes:

https://issues.redhat.com/browse/RHDHBUGS-2594

#### Problem
ActiveText widgets with `fetch:retrigger` (and no `fetch:url`) briefly show stale `ui:text` during the debounce window. The spinner only appears once evaluation starts, which is visible in production latency.

#### Solution

- Start loading immediately when retrigger dependencies change (before debounce).
- Keep spinner visible through fetch/evaluation completion.
- Clear widget values on retrigger when `fetch:clearOnRetrigger` is enabled to avoid stale state during refetch.

#### Changes

- useFetch: mark loading as soon as retrigger deps change and are satisfied.
- useFetchAndEvaluate: mark loading immediately on retrigger change even without fetch.
- Add `fetch:clearOnRetrigger` support for `ActiveTextInput`, `ActiveDropdown`, and `ActiveMultiSelect`.

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-form-widgets/src/utils/useClearOnRetrigger.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-form-widgets/src/utils/useClearOnRetrigger.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { useEffect, useRef } from 'react';
+import { useLayoutEffect, useRef } from 'react';
 import isEqual from 'lodash/isEqual';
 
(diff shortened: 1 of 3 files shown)


lokanandaprabhu:
Orchestrator - ActiveDropdown spinner
Orchestrator-active text spinner

mareklibra:
Thanks, it was tricky issue. It's good that this behavior is left to be optional, there are cases where the old flow makes sense.

PR #2304, opened 2026-02-10
feat(x2a): real K8s jobs for analyze and migrate

elai-shalev:
##  PR Summary

Implements working **analyze** and **migrate** K8s job phases, fixes job secret naming, adds automatic secret cleanup via ownerReferences, removes the callback mechanism (replaced with a TODO for a `report` command), and adds self-healing job status reconciliation against K8s.

- **Analyze phase**: copy `migration-plan.md` into the source dir before running, use correct output filename (`migration-plan-{module_name}.md` instead of `module_migration-plan.md`)
- **Migrate phase**: correct output path (`ansible/` instead of `migrated_ansible/`), copy migration plan into source dir, add `git pull --rebase` before push
- **Job secret naming**: include phase in secret name (`x2a-job-secret-{phase}-{jobId}`) to avoid collisions between concurrent jobs of different phases
- **OwnerReference on job secrets**: after K8s Job creation, set ownerReference on the job secret so it is garbage-collected automatically when the Job is deleted
- **Remove callback mechanism**: remove curl-based `collectArtifacts` callback from job script and error handler (replaced with TODO for `report` command)
- **Make repo auth tokens optional**: `sourceRepoAuth` and `targetRepoAuth` are no longer required in the run request body — they fall back to config values
- **Auto-reconcile job status**: when the `JobAlreadyRunning` guard finds active jobs in the DB, check K8s for each one; if K8s reports completion, update the DB and allow new jobs to proceed (self-healing for stale DB records left by the removed callback)

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/src/services/JobResourceBuilder.test.ts
+++ b/workspaces/x2a/plugins/x2a-backend/src/services/JobResourceBuilder.test.ts
@@ -828,7 +828,7 @@ describe('JobResourceBuilder', () => {
 
         const container = job.spec?.template.spec?.containers![0];
         // Script handles unknown phases with exit 1
-        expect(container!.args![0]).toContain('ERROR: Unknown phase');
+        expect(container!.args![0]).toContain('Unknown phase');
       });
     });
   });


elai-shalev:
@eloycoto @mareklibra PTAL
I've tested all the migration phases/jobs (except publish)
you can see the output here: 
https://github.com/elai-shalev/chef-examples/tree/main/1425eb20-91b7-4be2-96ac-8c49f9adee5d.TMP

elai-shalev:
> It LGTM.
> 
> Let's consider it in the context of #2266 .
> 
> Do you prefer to get this and 2266 in sync before merging?

Honestly, until the report command is implemented, I wouldn't change the job service to use collectArtifact. Even if 2266 is merged before, I would not refactor this PR

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -135,39 +137,50 @@ case "${PHASE}" in
...
     cp -v "${SOURCE_BASE}"/*.yaml "${OUTPUT_DIR}/" 2>/dev/null || true
 
     echo ""
     echo "=== Output directory contents ==="
     ls -la "${OUTPUT_DIR}/"
 
-    if [ ! -f "${OUTPUT_DIR}/module_migration-plan.md" ]; then
-      echo "ERROR: module_migration-plan.md not created"
+    if [ ! -f "${OUTPUT_DIR}/migration-plan-${MODULE_NAME_SANITIZED}.md" ]; then
+      echo "ERROR: migration-plan-${MODULE_NAME_SANITIZED}.md not created"
```
eloycoto: the same as before, this echo and error should be a report_error function.

PR #2325, opened 2026-02-13
feat(x2a): enhancements on x2a job script

eloycoto:
- Add project metadata on init phase.
- Fix some paths on the analyze phase to get the right migration path.


workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -213,12 +216,13 @@ case "${PHASE}" in
 
     # Copy migration-plan.md from target repo to source dir
     # The x2a tool does os.chdir(source_dir) and reads migration-plan.md from there
-    if [ ! -f "${PROJECT_PATH}/migration-plan.md" ]; then
+    if [ ! -f "${SOURCE_BASE}/${PROJECT_DIR}/migration-plan.md" ]; then
       ERROR_MESSAGE="migration-plan.md not found in ${PROJECT_PATH}/ - init phase must be run first"
```
elai-shalev: the error is still referencing the old path

PR #2326, opened 2026-02-13
feat(x2a): collectArtifacts with empty artifacts when it's an error

eloycoto:
I spotted this when debugging #2325, where the artifacts are required and it's not moving the state.


elai-shalev (approved):
lgtm

PR #2491, opened 2026-03-09
fix(x2a): alignment and wrapping of module and status cells

mareklibra:
Fix alignment and wrapping of module and status with icons.

Before:
Screenshot From 2026-03-09 09-38-29

---

After:
Screenshot From 2026-03-09 09-40-14


workspaces/x2a/plugins/x2a/src/components/ProjectStatusCell.tsx
```diff
@@ -17,7 +17,7 @@ import { useState } from 'react';
 import { ProjectStatus } from '@red-hat-developer-hub/backstage-plugin-x2a-common';
 
 import { PieChart, PieValueType } from '@mui/x-charts';
-import { Grid, makeStyles, Tooltip, Chip } from '@material-ui/core';
+import { Box, Grid, makeStyles, Tooltip, Chip } from '@material-ui/core';
```
elai-shalev: no need for the Grid import anymore I guess

PR #2498, opened 2026-03-09
feat(x2a): add the run-next bulk action

mareklibra:
Fixes: FLPATH-3409

Bulk run project action added. Filters eligible projects based on privileges.
The user can run next phase on
- all eligible projects
- per project, triggered from both the project list and project details page

A confirmation dialog describing scope of the action is shown prior running the action.

Screenshot From 2026-03-09 15-55-57

---

Screenshot From 2026-03-09 15-56-47

### DIFF
--- a/workspaces/x2a/plugins/x2a/src/components/tools/areEligibleModulesToRun.test.ts
+++ b/workspaces/x2a/plugins/x2a/src/components/tools/areEligibleModulesToRun.test.ts
@@ -1,0 +1,152 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


mareklibra:
> What if we add checkboxes to mark which modules we want to include in the batch?
currently he design has 1 or all, but we can add a way to pick and chose which modules to "batch run"

It would mean that the user needs to click per module to include it in the batch. So a very similar task as clicking on the `run next` action in the module role which we already have.

We might flip the view - list & select all modules by default and let the user to "uncheck" some of them.

I feel the UI would get overwhelmed by such "views" on the same functionality. 

Such feature would make sense for larger amount of modules. This is recently not expected, let's improve in case of change.

mareklibra:
> Should the button be greyed out if no modules have been discovered yet? 

Thanks. The action will be disabled if there is no eligible module to run it for.
It's not the case of the global action - it would be too costly to calculate.

PR #2510, opened 2026-03-11
feat(x2a): bitbucket git repos support

mareklibra:
Fixes: FLPATH-3389

The application authentication plus source and target repos newly support the [Bitbucket](https://bitbucket.org).

App authentication configuration works by following https://backstage.io/docs/auth/bitbucket/provider/.

OAuth token scopes and other requirements are documented in the README, there are specifics especially regarding username and resolver.

Tested on a private Bitbucket project (so requiring authenticated access), both `git push` and `git pull`.

Works well against cloud-based Bitbucket.
In case there is a requirement for a Bitbucket server (self-hosted, custom domain), additional work is needed, especially around SCN provider detection (so far naive implementation via parsing URL).

The source and target repos can be freely combined or reused.

---

Test matrix:

Screenshot From 2026-03-12 09-08-09


mareklibra:
Rebased

PR #2538, opened 2026-03-13
feat(x2): cancel phase action

mareklibra:
Fixes: FLPATH-3218

Additions:
- Added new `Canceled` state for jobs.
- New endpoint: `/projects/:projectId/modules/:moduleId/cancel`
  which stores current logs, updates the DB state, and deletes the Kubernetes job resource.
- Users can now access the cancel action from the projects table and phase detail card.

Fixes:
- Module status rendering: aligned with common icon+text chip format.
- PhasesCard tabs: view now persists after clicking action buttons.

The cancel action is relevant for analyze, migrate, and publish phases only.
The init phase cannot be canceled, as projects can be easily deleted and recreated. There is also no action to retrigger the init yet.

There is no confirmation dialog before canceling since reruns are simple and reliable. A dialog could be annoying and lead to delays (e.g., more LLM tokens spent on unintended runs).

Based on (needs rebase):
- [x] #2510 

---
Shown when running only:

Screenshot From 2026-03-13 10-19-47

---
Running and so eligible to `Cancel`:

Screenshot From 2026-03-13 11-09-19

---
When a phase is cancelled:

Screenshot From 2026-03-13 11-10-18


elai-shalev (commented):
Tested new functionality, lgtm!

PR #2565, opened 2026-03-17
feat(x2a): MCP server tools

eloycoto:
This enables a way to check the current status of the project using MCP tools, and can be used inside any project that needs to have AI tooling on top of it.

Because this uses OAuth2, I need to implement a quick page, because `plugin-auth` is not that easy to run within the current auth implementation.

To test, I use MCP Inspector, like this:

```
DANGEROUSLY_OMIT_AUTH=true npx @modelcontextprotocol/inspector
```

The auth env var is important, because without it, it messes up the OAuth2 token that we set[0].

I can see room for improvements where:
  - The auth could maybe be moved to https://www.npmjs.com/package/@backstage/plugin-auth
  - The router is highly coupled; maybe we need to have an HTTP router and services that can be consumed by MCP and by the router.
  - The MCP actions could be expanded (follow-up PR):
      - Run module
      - Get project - Get module

So the idea is just to get the basic terms of the MCP server in place, and finish all the implementation correctly by the end of the quarter.

[0] https://github.com/modelcontextprotocol/inspector/issues/633

### DIFF
--- a/workspaces/x2a/yarn.lock
+++ b/workspaces/x2a/yarn.lock
@@ -3777,7 +3777,7 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@backstage/plugin-permission-node@npm:^0.10.10":
+"@backstage/plugin-permission-node@npm:^0.10.10, @backstage/plugin-permission-node@npm:^0.10.3, @backstage/plugin-permission-node@npm:^0.10.6, @backstage/plugin-permission-node@npm:^0.10.7, @backstage/plugin-permission-node@npm:^0.10.9":
   version: 0.10.10
   resolution: "@backstage/plugin-permission-node@npm:0.10.10"
   dependencies:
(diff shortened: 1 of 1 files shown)


workspaces/x2a/packages/app/src/components/oauth2/ConsentPage.tsx
```diff
@@ -0,0 +1,249 @@
...
+          <ResponseErrorPanel
+            title="Authorization Error"
+            error={new Error(state.error)}
+          />
+        )}
+
+        {state.status === 'completed' && (
+          <Typography variant="h5" align="center" style={{ marginTop: 32 }}>
+            {state.action === 'approve'
+              ? `You have successfully authorized the application to access your ${appTitle} account. Redirecting...`
```
mareklibra: All strings must be internationalized

workspaces/x2a/plugins/x2a-backend/src/mcp/projects.ts
```diff
@@ -0,0 +1,89 @@
...
+          totalCount: z.number().describe('Total number of projects'),
+          projects: z
+            .string()
+            .describe('JSON-serialized array of project objects'),
+        }),
+    },
+    action: async ({ input, credentials, logger }) => {
+      logger.info('Running x2a:project:list MCP action');
+      const response = await listProjects(
+        input,
```
mareklibra: This is good. TypeScript can help here with maintenance in the future.

In addition, we will need unit tests covering
- functionality of the action
- guarding the contract between our router/services and the MCP tools.

This is simple to achieve at this scale but a headache when the tool list grows.

workspaces/x2a/plugins/x2a-backend/src/router/listProjects.ts
```diff
@@ -0,0 +1,68 @@
...
+import type { ProjectsGet } from '../schema/openapi';
+
+export async function listProjects(
+  query: ProjectsGet['query'],
+  deps: Pick<RouterDeps, 'permissionsSvc' | 'catalog' | 'x2aDatabase'>,
+  credentials: BackstageCredentials<BackstageUserPrincipal>,
+): Promise<ProjectsGet['response']> {
+  const { permissionsSvc, catalog, x2aDatabase } = deps;
+
+  const [userResult, adminViewResult] = await Promise.all([
```
mareklibra: This is reinventing the wheel, let's use the `useEnforceX2APermissions` which is well tested and consistent already. If not, we will eventually implement the same logic from scratch and end up in a similar solution.

workspaces/x2a/plugins/x2a-backend/src/router/projects.ts
```diff
@@ -90,24 +85,11 @@ export function registerProjectRoutes(
...
-    // list projects
-    const credentials = await httpAuth.credentials(req, { allow: ['user'] });
-    const userRef = getUserRef(credentials);
-    const groupsOfUser = await getGroupsOfUser(userRef, {
-      catalog,
-      credentials,
-    });
-
-    const { projects, totalCount } = await x2aDatabase.listProjects(query, {
+    const response = await listProjects(
```
mareklibra: Instead of introducing yet another layer (we have a lot of them already), the MCP tools can act as API clients. This will simplify the flow, testing and maintenance. The overhead of calls is very small.

workspaces/x2a/app-config.yaml
```diff
@@ -72,6 +78,9 @@ integrations:
     - host: bitbucket.org
 
 auth:
+  experimentalDynamicClientRegistration:
+    # enable the feature
+    enabled: true
```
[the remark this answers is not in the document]
elai-shalev: oops sorry. I meant "Auto enabled". as in do we want this enabled by default

PR #2594, opened 2026-03-24
feat(x2a): Initial API changes for devspaces

eloycoto:
To enable Openshift dev spaces we need a way to:

Trigger: Post the devspace.
URL: Get the devspace for a project
Delete: Delete the devspace when the user finished.

This is just the placeholder for the API, because I expect hard opinions on this topic, so just the placeholder for the API, and a followup of 6 to 10 PR step by step, from the deployment to the bussiness logic in backstage.

Second PR for FLPATH-3457
Related PR: https://github.com/x2ansible/x2a-convertor/pull/156


elai-shalev (commented):
overall looks good. assuming its a contract api pr for now

elai-shalev (approved):

mareklibra (approved):
It would make more sense to merge this along particular endpoint implementation to avoid orphaned code in the tree. But since there is a clear path to do it asap, it's ok.

PR #2614, opened 2026-03-26
FLPATH-3236 | [Bug] OpenShift Cost Management page accessible without RBAC authorization

asmasarw:
[FLPATH-3236 ](https://redhat.atlassian.net/browse/FLPATH-3236)| [Bug] OpenShift Cost Management page accessible without RBAC authorization

User always can see OpenShift Table Data regardless of what permission he has.

Main issue is that the API returns: 413 Payload Too Large

So the Solution is to call the API in chunks and then to see what is the DICISION - ALLOW or DENY.

UPDATED SCREENSHOT after fixing API Call with No-Permissions Granted.

image


PreetiW (changes requested):
@asmasarw please rebase the PR and left a comment for you. 
Thank you!

PR #2693, opened 2026-04-02
feat(app-defaults): add app-auth and app-integration

gashcrumb:
This change adds an 'app-auth' plugin which provides the sign-in page and related authentiction module API factories to an app using the new frontend system.  This change also adds a second 'app-integration' plugin which provides a default ScmAuth configuration.  In both cases these plugins bring code migrated from the old frontend system setup in the RHDH repository.  This change also adds a change to cleanup some noisy non-outcome altering output on the console when running the tests, along with some related updates to the eslint config so that the lint run by the pre-commit hook still works.  Based on review feedback, this change also includes an improved error message should the app be misconfigured such that the sign-in page would not render properly

Fixes [RHIDP-12359](https://redhat.atlassian.net/browse/RHIDP-12359)

Assisted-By: Cursor Desktop

Exporting these modules as dynamic plugins and starting RHDH backend's `start:next` script along with the relevant configuration that enables these modules provides a sign-in page that uses the correct authentication method as well as the ability to sign in using a secondary auth provider to access source control, for example:

image

image

When starting the workspace in development mode using `yarn start` the guest authentication method can be used to sign-in.


gashcrumb:
not sure the best version to use in the package.json files for these, maybe I should have just kept them at 0.0.1.

rohitkrai03 (approved):
/lgtm

workspaces/app-defaults/app-config.yaml
```diff
@@ -72,6 +72,8 @@ techdocs:
     type: 'local' # Alternatives - 'googleGcs' or 'awsS3'. Read documentation for using alternatives.
 
 auth:
+  # Required for RHDH app-auth SignInPage (guest in dev) and local e2e
+  environment: development
```
alizard0, line 76: just a question: shouldn't environment be outside auth? can we have multiple environments defined in the same file?

gashcrumb: This is the typical location for this variable, for example [here](https://github.com/redhat-developer/rhdh/blob/main/app-config.example.yaml#L31) from RHDH.  I believe yes, under here multiple environments can be defined per auth provider, however this sets the effective environment.  Also this is for the local development config for these plugins, ensuring the development app can be started properly.

workspaces/app-defaults/plugins/app-integrations/src/mergeScmAuthFromDeps.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    const configApi = {
+      getOptionalConfigArray: jest.fn().mockReturnValue(undefined),
+    } as unknown as ConfigApi;
+
+    mergeScmAuthFromDeps({
+      github: oauthStub,
+      gitlab: oauthStub,
+      azure: oauthStub,
+      bitbucket: oauthStub,
+      configApi,
```
alizard0, line 39: I assume integrations.* is configApi?

workspaces/app-defaults/plugins/app-integrations/src/mergeScmAuthFromDeps.test.ts
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    // Two GitHub hosts + default singletons for gitlab, azure, bitbucket.
```
alizard0, line 77: 👍🏽

workspaces/app-defaults/plugins/app-integrations/src/mergeScmAuthFromDeps.test.ts
```diff
@@ -0,0 +1,81 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import type { ConfigApi } from '@backstage/core-plugin-api';
+import { ScmAuth } from '@backstage/integration-react';
+
+import { mergeScmAuthFromDeps } from './mergeScmAuthFromDeps';
+
+describe('mergeScmAuthFromDeps', () => {
```
alizard0, line 22: Just a comment, non-blocking at all. We could increase the test base to cover all the providers. Currently, we only test github. I know it will be the same for the rest of the providers 😄

workspaces/app-defaults/plugins/app-integrations/src/mergeScmAuthFromDeps.test.ts
gashcrumb: yep, normally `configApi` is a service a plugin would import in the API to access the config, this test mocks that.

workspaces/app-defaults/plugins/app-auth/package.json
```diff
@@ -0,0 +1,80 @@
...
+    "prepack": "backstage-cli package prepack",
+    "postpack": "backstage-cli package postpack"
+  },
+  "dependencies": {
+    "@backstage/core-app-api": "^1.19.6",
+    "@backstage/core-components": "^0.18.8",
+    "@backstage/core-plugin-api": "^1.12.4",
+    "@backstage/frontend-plugin-api": "^0.15.1",
+    "@backstage/plugin-app": "^0.4.2",
+    "@mui/material": "5.18.0"
```
alizard0, line 58: Does it need to be pinned?

workspaces/app-defaults/plugins/app-integrations/src/mergeScmAuthFromDeps.test.ts
gashcrumb: yep, for sure.  Worth discussing the best approach here down the line, as this could be done somewhat here but also some of this kind of testing could make more sense when there's an entry in the overlay repo, allowing for tests to be done against actual systems as needed.  At this level probably the tests should use mocks rather than real services.

workspaces/app-defaults/plugins/app-auth/src/appAuthModule.tsx
```diff
@@ -0,0 +1,122 @@
...
+        OAuth2.create({
+          discoveryApi,
+          oauthRequestApi: oauthRequestApi,
+          provider: {
+            id: 'auth0',
+            title: 'Auth0',
+            icon: () => null,
+          },
+          defaultScopes: ['openid', 'email', 'profile'],
+          environment: configApi.getOptionalString('auth.environment'),
```
alizard0, line 77: So we can't have auth providers using different environments?

alizard0 (approved):
lgtm

workspaces/app-defaults/plugins/app-auth/src/appAuthModule.tsx
gashcrumb: they can be configured, but afaik there can only be one effective enabled environment.  I should mention that this is based on the behavior of the existing [sign-in page from the old frontend system](https://github.com/redhat-developer/rhdh/blob/main/packages/app/src/components/SignInPage/SignInPage.tsx#L158) that the plugin brings in.

workspaces/app-defaults/plugins/app-auth/package.json
gashcrumb: mui/material yes, the other dependencies I don't think so.  The backstage-cli would override these pinned versions to use ^ on the next backstage version bump anyways.

PR #2745, opened 2026-04-10
fix(x2a): git using authN credentials better

eloycoto:
Just to avoid to have the token in the URL and to be leaked by any way, if we do this way the token is not going  to be in the url.
Doc: https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlbaseinsteadOf


elai-shalev (commented):
looks good

elai-shalev (approved):

eloycoto:
@elai-shalev bitbucket is working: 

image

PR #2749, opened 2026-04-13
feat(orchestrator): Run as Event for Kafka-configured workflows

lokanandaprabhu:
### Story:

https://redhat.atlassian.net/browse/RHIDP-9144

### Summary:

Users can trigger workflows as an event from the RHDH orchestrator UI when orchestrator.kafka is present in app config. The flow mirrors Run, but the execute request includes isEvent: true in the input payload so the platform/backend can treat it as an event-style trigger (e.g. CloudEvent → Kafka).

When the execute API returns an instance id of kafkaEvent (async / delayed start), the UI does not open a single instance page. It navigates to the workflow runs view and shows an informational message that an event was sent and the run will appear in the list once it starts.

https://github.com/user-attachments/assets/5d90c7e6-a611-4a8c-87be-a45a7ee5682c

https://github.com/user-attachments/assets/a243c357-32f1-440b-8da2-c448e6275c69

https://github.com/user-attachments/assets/2715d566-ad56-48d2-a1c9-20f7f01ad835

Screenshot 2026-04-13 at 12 17 26 PM

### How to test

**- Config**

  - In app config, add an orchestrator.kafka block (clientId, brokers, optional logLevel) so the frontend gate is on.
  - Restart the app so config is picked up.
  - Run as Event visibility

**- Open a workflow’s Execute page.**

  - Confirm Run as Event appears next to Run on the review step (multi-step) or equivalent for single-step + Missing schema flow.
  - Remove or comment out orchestrator.kafka, restart, and confirm Run as Event is hidden.

**- Payload**

  - Use browser devtools (network tab) on execute.
  - Click Run as Event and confirm the execute request body includes isEvent: true inside inputData (alongside your form fields).

**- kafkaEvent redirect and message (needs backend returning that id, or a temporary test shim):**

  - After execute, if the API returns id: kafkaEvent, the app should land on the workflow runs list (or entity-scoped runs view) with the event triggered info alert.
  - Dismissing the alert should remove eventTriggered from the URL.

**- Normal run unchanged**

  - Run (without event) should still navigate to the instance page when a real instance id is returned.

### DIFF
--- a/workspaces/orchestrator/.changeset/run-as-event-kafka.md
+++ b/workspaces/orchestrator/.changeset/run-as-event-kafka.md
@@ -1,0 +1,8 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+'@red-hat-developer-hub/backstage-plugin-orchestrator-common': patch
+'@red-hat-developer-hub/backstage-plugin-orchestrator-form-api': patch
+'@red-hat-developer-hub/backstage-plugin-orchestrator-form-react': patch
+---
+
+Add Run as Event when `orchestrator.kafka` is configured: send `isEvent` with execute input, redirect to workflow runs with a notice when the response id is `kafkaEvent`.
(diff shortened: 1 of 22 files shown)


lholmquist (approved):
applied this to my PR #2512 locally and everything worked as expected

/lgtm

lholmquist (approved):
/lgtm

PR #2764, opened 2026-04-14
chore(x2a): project name and short id used in target repo

elai-shalev:
instead of a long UUID representing a project in the target repo, now a combination of a sanitzed project name + UUID[0:6] chars will be used. This will enhance user experience

related to https://redhat.atlassian.net/browse/FLPATH-4020

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/src/services/types.ts
+++ b/workspaces/x2a/plugins/x2a-backend/src/services/types.ts
@@ -47,7 +47,7 @@ export interface JobCreateParams {
   projectId: string;
   projectName: string;
   /**
-   * Project abbreviation - used for directory naming in target repository
+   * Project abbreviation
    */
   projectAbbrev: string;
   phase: MigrationPhase;
--- a/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
+++ b/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
@@ -235,9 +235,7 @@ git_clone_repos
 # Define paths
 TARGET_BASE="/workspace/target"
 SOURCE_BASE="/workspace/source"
-SANITIZED_NAME=$(echo "${PROJECT_NAME}" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9-]/-/g' | sed 's/--*/-/g' | sed 's/^-//;s/-$//')
-SHORT_UUID="${PROJECT_ID:0:6}"
-PROJECT_DIR="${SANITIZED_NAME}-${SHORT_UUID}"
+# PROJECT_DIR is pre-computed by the backend (sanitized name + short UUID)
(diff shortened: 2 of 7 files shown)


mareklibra:
By having that, we can completely remove the project abbreviation.

elai-shalev:
> By having that, we can completely remove the project abbreviation.

Maybe we should? anyway that removal would be in a seperate pr

mareklibra:
>>    By having that, we can completely remove the project abbreviation.

> Maybe we should? anyway that removal would be in a seperate pr

The abbreviation and changes in this PR share the use-case. There is no additional use for it when we merge this.
It's ok to do the removal in a follow-up.

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -235,7 +235,9 @@ git_clone_repos
 # Define paths
 TARGET_BASE="/workspace/target"
 SOURCE_BASE="/workspace/source"
-PROJECT_DIR="${PROJECT_ID}.${PROJECT_ABBREV}"
+SANITIZED_NAME=$(echo "${PROJECT_NAME}" | tr '[:upper:]' '[:lower:]' | sed 's/[^a-z0-9-]/-/g' | sed 's/--*/-/g' | sed 's/^-//;s/-$//')
```
mareklibra: The `SANITIZED_NAME` can become very long, what about limiting it to 64 characters or similar? Taking first and last part of the project name? Or truncate at least.

Using a default for empty project name? This should not happen thanks to our other checks, but... If this happens, the name will start with a dash which should be avoided in any case (like project name starts with "-n ; echo I am malicious".

mareklibra (commented):
Can you add a changeset?

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
elai-shalev: good points, let me apply some guards

workspaces/x2a/plugins/x2a-backend/src/services/types.ts
```diff
@@ -47,7 +47,7 @@ export interface JobCreateParams {
   projectId: string;
```
eloycoto, line 47: What about if we create a ValueObject called Project:

https://khalilstemmler.com/articles/typescript-value-object/

And from here, we can say:

Project.name () -> self.projectName+self.abbre+self.id.slice(0, 5)?
Project.id() -> self.id
Project.base_name: self.projectName+sefl.abbr


And we have all in there, so we can skip the bash changes? and we can do all sanitization here?

elai-shalev:
@eloycoto @mareklibra Thank you for the review
I've implemented Eloy's suggestion - it does seem much cleaner and maintainable. 
also added changeset

mareklibra (approved):

eloycoto (approved):

PR #2766, opened 2026-04-14
Add API Proxy | Add All Tabs to reflect API

asmasarw:
## Overview

Introduces the **DCM (Dynamic Configuration Management) plugin** for Red Hat Developer Hub — a Backstage plugin that provides a self-service UI for managing infrastructure provisioning workflows through the DCM API Gateway.

---

### Providers management
Register, view, and manage service providers — the infrastructure backends that fulfil provisioning requests. Each provider entry includes its endpoint, service type, schema version, supported operations, and live health status.

### Policy management
Create and manage placement policies that control how resources are allocated. Policies can be scoped globally or per-user and toggled enabled/disabled. A priority field controls evaluation order when multiple policies apply.

### Catalog Items
Define reusable service blueprints (catalog items) backed by a service type. Each catalog item carries a configurable field schema that controls what parameters end users fill in when provisioning an instance.

### Catalog Item Instances
Provision instances from catalog items by filling in the required field configuration. Instances represent active provisioning requests and can be deleted when no longer needed.

### Resources
View and manage placement resources — the concrete infrastructure objects created by the Policy Engine. Resources can be created from a catalog item instance, deleted, or rehydrated (re-evaluated against current policies).

---

## Backend capabilities

- **Secure proxy** — all API calls from the browser are routed through the DCM backend, which injects a short-lived SSO bearer token (obtained via client-credentials flow from Red Hat SSO) before forwarding to the configured `dcm.apiGatewayUrl`. Tokens are cached server-side and refreshed automatically before expiry.
- **Permission integration** — access to the plugin is controlled via Backstage's permission framework. A `dcm.plugin` read permission gates both the UI and the `/token` endpoint, so users without the required permission cannot interact with the DCM API.
- **Access endpoint** — exposes a `/access` endpoint the frontend uses to determine whether the current user has permission, enabling the UI to show appropriate messaging rather than silent failures.

---

## Configuration

```yaml
dcm:
  apiGatewayUrl: https://your-dcm-gateway.example.com
  clientId: your-client-id
  clientSecret: your-client-secret
  # optional — defaults to https://sso.redhat.com
  ssoBaseUrl: https://sso.redhat.com
  
  
 ---

## Tests and Coverage
- **Tests** — added tests to cover every tab content (Each API Endpoint)

### CODE REVIEW

**Comment by asmasarw** [comment]
> ### Code Review by Qodo
> `🐞 Bugs (3)`   `📘 Rule violations (0)`   `📎 Requirement gaps (0)`
> 🐞\ `≡ Correctness (1)` `⛨ Security (1)` `⚙ Maintainability (1)`
> Grey Divider
> 
> 
> Action required
> 
> 1. Proxy skips permission check `🐞` `⛨`
> 
> > Description
> > 
> > createDcmProxy forwards requests for any authenticated caller but never runs the DCM permission
> > authorization used elsewhere (authorize() via permissions.authorize), so users without DCM
> > permissions can still reach upstream DCM APIs through /proxy. This is an authorization bypass
> > relative to /token which explicitly denies unauthorized callers with 403.
> 
> > Code
> > 
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[R31-75]](https://github.com/redhat-developer/rhdh-plugins/pull/2766/files#diff-af44fcebd2874e9d3da44a84c6255826ab5580e9495228d2bf7f3dfd5accad46R31-R75)`
> > ```diff
> > +export function createDcmProxy(options: RouterOptions) {
> > +  return async (req: Request, res: Response): Promise<void> => {
> > +    const { logger, config } = options;
> > +
> > +    const apiGatewayUrl = config.getOptionalString('dcm.apiGatewayUrl');
> > +    if (!apiGatewayUrl) {
> > +      logger.error(
> > +        'dcm.apiGatewayUrl is not configured — cannot proxy DCM API requests.',
> > +      );
> > +      res
> > +        .status(503)
> > +        .json({ error: 'DCM API gateway is not configured on the server.' });
> > +      return;
> > +    }
> > +
> > +    // req.params[0] is the captured wildcard after /proxy/
> > +    const wildcardPath = (req.params as Record<string, string>)[0] ?? '';
> > +
> > +    const targetUrl = new URL(
> > +      `${API_BASE_PATH}/${wildcardPath}`,
> > +      apiGatewayUrl,
> > +    );
> > +
> > +    // Forward all query parameters from the original request
> > +    const incomingParams = new URLSearchParams(
> > +      req.query as Record<string, string>,
> > +    );
> > +    incomingParams.forEach((value, key) => {
> > +      targetUrl.searchParams.set(key, value);
> > +    });
> > +
> > +    logger.debug(
> > +      `DCM proxy: ${req.method} ${req.path} → ${targetUrl.toString()}`,
> > +    );
> > +
> > +    let tokenResult;
> > +    try {
> > +      tokenResult = await getTokenFromApi(options);
> > +    } catch (err) {
> > +      logger.error(`DCM proxy: failed to obtain access token — ${err}`);
> > +      res
> > +        .status(502)
> > +        .json({ error: 'Failed to obtain upstream access token.' });
> > +      return;
> > +    }
> > ```
> 
> > Evidence
> > 
> > The proxy handler obtains an upstream token and forwards the request without calling authorize();
> > meanwhile, the /token endpoint explicitly calls authorize() (which uses permissions.authorize over
> > dcmPluginPermissions) and returns 403 when denied. The backend plugin’s httpRouter auth policy for
> > /token and /proxy is the same (user-cookie), demonstrating that the proxy still needs an explicit
> > permission check if DCM permissions are intended to gate access.
> > 
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[31-75]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts/#L31-L75)`
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/token.ts[23-35]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-backend/src/routes/token.ts/#L23-L35)`
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/access.ts[22-34]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-backend/src/routes/access.ts/#L22-L34)`
> > `[workspaces/dcm/plugins/dcm-backend/src/plugin.ts[48-63]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-backend/src/plugin.ts/#L48-L63)`
> 
> > Agent prompt
> > 
> > ```
> > The issue below was found during a code review. Follow the provided context and guidance below and implement a solution
> > 
> > ## Issue description
> > The new DCM backend proxy (`/proxy/*`) forwards upstream DCM API requests for any authenticated user, but it does not enforce the DCM permission check used by `/token`.
> > 
> > ## Issue Context
> > `/token` calls `authorize(req, options)` and returns `403` when the user lacks DCM permissions. The proxy should apply the same authorization before minting an upstream token and forwarding requests.
> > 
> > ## Fix Focus Areas
> > - workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[31-75]
> > - workspaces/dcm/plugins/dcm-backend/src/routes/token.ts[23-35]
> > - workspaces/dcm/plugins/dcm-backend/src/routes/access.ts[22-34]
> > 
> > ## Implementation notes
> > - Import and call `authorize(req, options)` early in the proxy handler.
> > - If denied, return `403 { error: 'Forbidden' }` (consistent with `/token`) and do not call `getTokenFromApi` or `fetch` upstream.
> > - Consider adding/adjusting tests in `proxy.test.ts` to verify a denied authorization returns 403.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ⓘ Copy this prompt and use it to remediate the issue with your preferred AI generation tools`
> 
> 
> Remediation recommended
> 
> 2. Query params lose duplicates `🐞` `≡`
> 
> > Description
> > 
> > The proxy converts req.query to URLSearchParams and uses searchParams.set(), which overwrites any
> > duplicate query keys and can change semantics for multi-valued parameters. This can cause upstream
> > requests like ?a=1&a=2 to arrive as a single value instead of two.
> 
> > Code
> > 
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[R54-60]](https://github.com/redhat-developer/rhdh-plugins/pull/2766/files#diff-af44fcebd2874e9d3da44a84c6255826ab5580e9495228d2bf7f3dfd5accad46R54-R60)`
> > ```diff
> > +    // Forward all query parameters from the original request
> > +    const incomingParams = new URLSearchParams(
> > +      req.query as Record<string, string>,
> > +    );
> > +    incomingParams.forEach((value, key) => {
> > +      targetUrl.searchParams.set(key, value);
> > +    });
> > ```
> 
> > Evidence
> > 
> > The implementation uses URLSearchParams and then calls targetUrl.searchParams.set(key, value) for
> > each param. URLSearchParams.set replaces existing values for the same key, so repeated query params
> > cannot be preserved.
> > 
> > `[workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[54-60]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts/#L54-L60)`
> 
> > Agent prompt
> > 
> > ```
> > The issue below was found during a code review. Follow the provided context and guidance below and implement a solution
> > 
> > ## Issue description
> > The proxy currently forwards query parameters by iterating `URLSearchParams` and calling `targetUrl.searchParams.set(key, value)`, which overwrites repeated keys.
> > 
> > ## Issue Context
> > Some clients/servers rely on repeated query keys (e.g. `?label=a&label=b`). The proxy should preserve these instead of collapsing them.
> > 
> > ## Fix Focus Areas
> > - workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts[54-60]
> > 
> > ## Implementation notes
> > - Iterate `req.query` entries directly.
> > - If a value is an array, `append` each element.
> > - If a value is a single scalar, `set` it (or `append` consistently).
> > - Prefer `append` over `set` when forwarding from an already-parsed structure to preserve duplication.
> > - Add a focused unit test for duplicate query keys (e.g. `/proxy/providers?a=1&a=2`).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ⓘ Copy this prompt and use it to remediate the issue with your preferred AI generation tools`
> 
> 
> Advisory comments
> 
> 3. Headers merge overridden `🐞` `⚙`
> 
> > Description
> > 
> > DcmBaseClient.fetch constructs merged headers but then spreads ...init after, which will overwrite
> > the merged headers object whenever init.headers is provided. This makes the header merging
> > fragile and can silently drop defaults in future call sites.
> 
> > Code
> > 
> > `[workspaces/dcm/plugins/dcm-common/src/clients/DcmBaseClient.ts[R42-48]](https://github.com/redhat-developer/rhdh-plugins/pull/2766/files#diff-7eb4fcb3de3534ac3ab61a566ae1e3068a84f04f0ec229d6882ac418aef12c29R42-R48)`
> > ```diff
> > +  protected async fetch<T>(path: string, init?: RequestInit): Promise<T> {
> > +    const baseUrl = await this.discoveryApi.getBaseUrl(PLUGIN_ID);
> > +    const url = `${baseUrl}/proxy/${path}`;
> > +    const response = await this.fetchApi.fetch(url, {
> > +      headers: { 'Content-Type': 'application/json', ...init?.headers },
> > +      ...init,
> > +    });
> > ```
> 
> > Evidence
> > 
> > The RequestInit object passed to fetch is built with headers: { ... } and then ...init. If
> > init contains a headers property, it replaces the previously built merged headers object,
> > negating the merge behavior.
> > 
> > `[workspaces/dcm/plugins/dcm-common/src/clients/DcmBaseClient.ts[42-48]](https://github.com/redhat-developer/rhdh-plugins/blob/76f9380f00cdf9d7b7c14e50afd69d38347b0fb1/workspaces/dcm/plugins/dcm-common/src/clients/DcmBaseClient.ts/#L42-L48)`
> 
> > Agent prompt
> > 
> > ```
> > The issue below was found during a code review. Follow the provided context and guidance below and implement a solution
> > 
> > ## Issue description
> > `DcmBaseClient.fetch` attempts to merge default headers with `init.headers`, but it spreads `...init` after specifying `headers`, which can overwrite the merged header object.
> > 
> > ## Issue Context
> > This is easy to miss and can cause subtle bugs if future callers provide `init.headers` (e.g., adding `Accept`) and unintentionally drop defaults.
> > 
> > ## Fix Focus Areas
> > - workspaces/dcm/plugins/dcm-common/src/clients/DcmBaseClient.ts[42-48]
> > 
> > ## Implementation notes
> > - Spread `...init` first, then define `headers` last, e.g.:
> >  - `const response = await fetch(url, { ...init, headers: { 'Content-Type': 'application/json', ...(init?.headers ?? {}) } });`
> > - Keep the existing override semantics (caller headers should win).
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > `ⓘ Copy this prompt and use it to remediate the issue with your preferred AI generation tools`
> 
> Grey Divider
> 
> ⓘ The new review experience is currently in Beta. [Learn more](https://docs.qodo.ai/qodo-documentation/code-review)
> Grey Divider
> 
> [Qodo Logo](https://www.qodo.ai)

**RBAC is NOT Implemented YET.
WILL BE ADDED SOON**

File: workspaces/dcm/.changeset/slim-lobsters-refactor.md [hunk 1]
```diff
@@ -0,0 +1,39 @@
+---
+'@red-hat-developer-hub/backstage-plugin-dcm': minor
+---
+
+Refactor DCM frontend plugin for reusability, maintainability, and test coverage.
```
mareklibra, line 7: This will be part of the CHANGELOG.md

It's not common to have such a detailed list of internal changes, the logs tends to be focused on the user of the plugin rather than its developers. Usually PRs or commit messages are better fit for such level of details.

Just noting. Leaving that up to you consideration.

workspaces/dcm/plugins/dcm/src/components/DcmCrudTabLayout.tsx
```diff
@@ -0,0 +1,199 @@
...
+ *
+ * Renders either:
+ *   - A `<Progress>` spinner while loading
+ *   - An error alert with a Retry button when `loadError` is set
+ *   - An empty state illustration + primary action when `items` is empty
+ *   - A toolbar row with the primary action button + a searchable
+ *     `<InfoCard>` containing a paginated `<Table>` when items are present
+ *
+ * @example
+ * <DcmCrudTabLayout
```
mareklibra: I am not sure if this example is of much help

workspaces/dcm/plugins/dcm/src/components/DcmFormDialogActions.tsx
```diff
@@ -0,0 +1,86 @@
...
+   */
+  disabled?: boolean;
+}>;
+
+/**
+ * Standardised Save / Cancel action row for DCM form dialogs.
+ *
+ * Pass this as the `actions` prop to `DcmFormDialog`:
+ * @example
+ * <DcmFormDialog
```
mareklibra: likewise

If needed, I would rather focus on describing the properties themselves, so the caller can better decide.

workspaces/dcm/plugins/dcm/src/components/DcmFormDialogActions.tsx
```diff
@@ -0,0 +1,86 @@
...
+ *     />
+ *   }
+ * >
+ *   ...
+ * </DcmFormDialog>
+ */
+export function DcmFormDialogActions({
+  onSubmit,
+  onCancel,
+  submitLabel = 'Save',
```
mareklibra: Such defaults will be harder to internationalize

workspaces/dcm/plugins/dcm/src/components/TruncatedText.tsx
```diff
@@ -0,0 +1,70 @@
...
+ * can grow unbounded (names, endpoints, descriptions, paths, IDs, …).
+ */
+export function TruncatedText({
+  text,
+  maxWidth = 180,
+  variant = 'body2',
+  color,
+  bold = false,
+  fallback = (
+    <Typography variant="caption" color="textSecondary">
```
mareklibra: Can you move such complex default to the place where it is actually used?

workspaces/dcm/plugins/dcm/src/pages/catalog-items/CatalogItemsTabContent.tsx
```diff
@@ -0,0 +1,387 @@
...
+      },
+      {
+        title: 'Fields',
+        field: 'spec.fields',
+        sorting: false,
+        render: item => {
+          const count = item.spec?.fields?.length ?? 0;
+          return (
+            <Typography variant="body2" color="textSecondary">
+              {count} {count === 1 ? 'field' : 'fields'}
```
mareklibra, line 184: The amount of complex text is growing. 

Mind introducing i18n asap, it will safe you a lot of work (like this one) and troubles later.

workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts
```diff
@@ -0,0 +1,120 @@
...
+      targetUrl.searchParams.set(key, value);
+    });
+
+    logger.debug(
+      `DCM proxy: ${req.method} ${req.path} → ${targetUrl.toString()}`,
+    );
+
+    let tokenResult;
+    try {
+      tokenResult = await getTokenFromApi(options);
```
mareklibra, line 68: I might be missing something but how is `dcm.plugin` permission enforced here?

It looks like any authenticated Backstage user can get this token.

workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts
```diff
@@ -0,0 +1,120 @@
...
+    let upstreamResponse: globalThis.Response;
+    try {
+      upstreamResponse = await fetch(targetUrl.toString(), {
+        method: req.method,
+        headers: requestHeaders,
+        // Only attach body for methods that support it
+        body:
+          ['POST', 'PUT', 'PATCH'].includes(req.method) &&
+          req.headers['content-length'] !== '0'
+            ? JSON.stringify(req.body)
```
mareklibra, line 96: Is json ALWAYS accepted by the target API?

workspaces/dcm/.changeset/slim-lobsters-refactor.md
```diff
@@ -0,0 +1,39 @@
+---
+'@red-hat-developer-hub/backstage-plugin-dcm': minor
```
mareklibra, line 2: All changed plugins should be listed here

mareklibra (changes requested):
Left some comments.

Some of the Sonar findings are relevant.

Mind introducing i18n asap - it will safe you from headaches later. And help even now...

workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts
asmasarw: Because there are no Authentication or Authorization for this Plugin.
It will be implemented later on per planning.

workspaces/dcm/plugins/dcm-backend/src/routes/proxy.ts
asmasarw: Yes

workspaces/dcm/.changeset/slim-lobsters-refactor.md
asmasarw: Since there was a big diff (refactor) between two commits - I prefer to have them here just in case.

mareklibra (approved):

PR #2768, opened 2026-04-14
fix(x2a): bitbucket RepoUrlPicker wrapper

eloycoto:
This is a wrapper around RepoUrlPicker that it's a hotfix for the following issue which gives enough context:

https://github.com/backstage/backstage/issues/33887

Fix FLPATH-3887


workspaces/x2a/plugins/x2a/src/scaffolder/RepoUrlPickerWithBitbucketFix.tsx
```diff
@@ -0,0 +1,143 @@
...
+        const integration = target.byHost(host);
+        if (!integration || !BITBUCKET_TYPES.has(integration.type)) {
+          return integration;
+        }
+
+        return new Proxy(integration, {
+          get(innerTarget, innerProp, innerReceiver) {
+            if (innerProp === 'type') {
+              return 'bitbucket';
+            }
```
elai-shalev, line 93: we assume this part will change / be removed once the bug is resolved, correct? as they may decouple bitbucketCloud and bitbucketServer

elai-shalev (approved):
Looks good considering it's a hack. I would add some tests for the RepoUrlPickerWithBitbucketFix component. Those would likely break once backstage version is updated (if it includes fix) and would signify we need to fix here.

workspaces/x2a/plugins/x2a/src/scaffolder/RepoUrlPickerWithBitbucketFix.tsx
eloycoto: yuup

workspaces/x2a/plugins/scaffolder-backend-module-x2a/templates/conversion-project-template.yaml
```diff
@@ -177,7 +177,7 @@ spec:
                     description: |
                       The Owner should be your SCM (Source Code Management) username. The repository name should be a name that already exists in your SCM account. It will be populated by converted Ansible sources and intermediary artifacts.
                     type: string
-                    ui:field: RepoUrlPicker
+                    ui:field: X2ARepoUrlPicker
```
mareklibra, line 188: A brief comment that this is a workaround for BitBucker will be helpful in the future.

mareklibra:
This is probably the fastest way to get a fix and the only one for RHDH 1.9.
Anyway, this will most probably break in RHDH 1.10 - can you create a jira to track the need for retest/replacements?

Tests for the `wrapIntegrationsApi` will be helpful.

mareklibra (approved):

PR #2802, opened 2026-04-16
fix(orchestrator): makes the bodyParser limit configurable.

lholmquist:

A customer was having an issue where the workflows content length was over the 100kb limit.  This adds a configurable value to make the content lenght larger

fixes: https://redhat.atlassian.net/browse/RHDHSUPP-351

### DIFF
--- a/workspaces/orchestrator/app-config.yaml
+++ b/workspaces/orchestrator/app-config.yaml
@@ -158,6 +158,8 @@ catalog:
 dynamicPlugins:
   frontend: {}
 orchestrator:
+  # Uncomment to set the content length limit for the requests. Defaults to 102400 bytes (100kb)
+  # contentLengthLimit: 10mb
   # Uncomment and configure to use the log viewer
   # workflowLogProvider:
   #   loki:
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
@@ -202,9 +202,12 @@ export async function createBackendRouter(
   const permissionsIntegrationRouter = createPermissionIntegrationRouter({
     permissions: orchestratorPermissions,
   });
-  router.use(express.json());
+  const contentLengthLimit = config.getOptionalString(
+    'orchestrator.contentLengthLimit',
+  );
+  router.use(express.json({ limit: contentLengthLimit }));
   router.use(permissionsIntegrationRouter);
-  router.use('/workflows', express.text());
+  router.use('/workflows', express.text({ limit: contentLengthLimit }));
   router.get('/health', (_, response) => {
     logger.info('PONG!');
     response.json({ status: 'ok' });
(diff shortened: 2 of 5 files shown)


workspaces/orchestrator/app-config.yaml
```diff
@@ -158,6 +158,8 @@ catalog:
 dynamicPlugins:
   frontend: {}
 orchestrator:
+  # Uncomment to set the content length limit for the requests. Defaults to 102400 bytes (100kb)
+  # contentLengthLimit: 10mb
```
PatAKnight, line 162: Don't forget to add the new config value here: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/plugins/orchestrator-common/config.d.ts

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -201,9 +201,12 @@ export async function createBackendRouter(
   const permissionsIntegrationRouter = createPermissionIntegrationRouter({
     permissions: orchestratorPermissions,
   });
-  router.use(express.json());
+  const contentLengthLimit = config.getOptionalString(
+    'orchestrator.contentLengthLimit',
+  );
+  router.use(express.json({ limit: contentLengthLimit }));
```
PatAKnight, line 216:
```suggestion
  router.use(express.json());
```

It looks like the QODO review is complaining about having the limit added globally as well as the `/workflows` endpoint. I agree with the suggestion to drop the limit globally and just have the one on the endpoints that best need the larger limits.

PatAKnight (changes requested):
I would add some docs to this as well. I don't know where the current product docs live for the orchestrator plugin but having a small blurb here can give us something to point to for the docs team to ingest.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
lholmquist: So i decided to ask cursor about that `/workflows` endpoint and it doesn't look like it is actually being used.  In fact, if you try to go to it, it will return a 404.  The endpoint to actually interact with workflows is `/v2/workflows/...`. i should probably remove the limit config from that `/workflows` endpoint and it should be kept on the global one

workspaces/orchestrator/app-config.yaml
lholmquist: added

lholmquist:
@PatAKnight I added a comment on the router, which i think we can point the docs to at, https://github.com/redhat-developer/rhdh-plugins/pull/2802/changes#diff-b34e43c026bf84e8d06cc07eceed2806b08b96206b3c033d493ec8a7b1ce3ceeR208-R215

Also see my comment about the `/workflows` route,  https://github.com/redhat-developer/rhdh-plugins/pull/2802#discussion_r3100929357

PatAKnight (approved):
LGTM

PR #2826, opened 2026-04-20
RHINENG-25551 fix(cost-management): surface SSO auth errors instead of silent failures

PreetiW:
### Description

[RHINENG-25551](https://redhat.atlassian.net/browse/RHINENG-25551)  [FLPATH-3237](https://redhat.atlassian.net/browse/FLPATH-3237)
fix(cost-management): surface SSO auth errors instead of silent failures

When the plugin is configured with invalid hybrid cloud service account credentials, the UI previously showed a generic "Bad Gateway" or empty table instead of an actionable error message.

- tokenUtil: extract SSO error details (error_description/error) from the response body instead of throwing bare HTTP status text
- secureProxy: return 502 with credential-specific message when SSO authentication fails, distinguishing it from other proxy errors
- OptimizationsClient & CostManagementSlimClient: read the error field from JSON response bodies on non-OK responses so the frontend displays the backend's descriptive message

Made-with: Cursor

### Screenshots 

image

image


asmasarw (approved):
LGTM

PR #2830, opened 2026-04-20
Create Deployment Scripts for Both Dynamic-Plugin and Backstage-Standalone

asmasarw:
# Introduced new Image Creation Options

**We have 2 Types of Images:**

-  Dynamic Plugin
-  Standalone Backstage Application

## Dynamic Plugin

**What it builds:** A tiny OCI artifact (~few MB) containing only the compiled plugin JavaScript files.

**How it works:**

Exports the frontend (plugins/dcm) and backend (plugins/dcm-backend) as dynamic plugins
Packages them into a single OCI image tagged quay.io/dcm-project/dcm-ui:VERSION
tgz additionally extracts the .tgz tarballs from that image via skopeo
Who consumes it: An already-running Red Hat Developer Hub instance. RHDH loads the plugin at startup by pulling the OCI artifact and mounting the plugin files into its own runtime. You don't run this image directly — RHDH runs it.

**Use case:** Your production RHDH cluster wants to add the DCM plugin without rebuilding the whole Backstage app.

## Standalone Backstage Application

**What it builds:** A full, self-contained Docker image (~1GB+) with the entire Backstage application baked in — Node.js runtime, all dependencies, compiled frontend, and compiled backend.

**How it works:**

Runs yarn tsc + yarn workspace app build + yarn workspace backend build inside a build container
Copies only the production artifacts into a lean runtime image
Tags it quay.io/dcm-project/dcm-ui:VERSION
Who consumes it: You run it directly with docker run or podman-compose up. It starts a complete Backstage server on port 7007 that serves both the API and the frontend — no RHDH required.

**Use case:** Your compose.yaml setup — a standalone Backstage instance specifically for the DCM plugin.

---
# How to Generate Image

### To Generate Dynamic Plugin
`VERSION=<semver> ./scripts/generate-image.sh [oci|tgz|push]`

### To Generate Standalone Backstage Application
`VERSION=<semver> ./scripts/generate-image.sh backstage-image [--push]`

### Script Description (Help)
`./scripts/generate-image.sh --help`


mareklibra (approved):
This is good for (dev-) testing but eventually you will need OCI builds by https://github.com/redhat-developer/rhdh-plugin-export-overlays .

For the NPM packages for vanilla Backstage deployment, the rhdh-plugins' repo CI with its changesets and Version package PRs is the best option.

PR #2834, opened 2026-04-20
fix(orchestrator-form): Review step crash when form data is empty

lokanandaprabhu:
#### Problem

On workflow execute, moving to the Review step could throw when formData was effectively empty—for example input schemas with only display-only ActiveText (no onChange, so nothing is submitted) or when every visible value is filtered out for review (e.g. `ui:hidden`). `generateReviewTableData` could return undefined, and `NestedReviewTable` called Object.entries on that value.

#### Solution

`generateReviewTableData`: always return a root payload object; use {} when the processed root is missing.
`NestedReviewTable`: treat missing data as {} so the table never receives undefined.

---------------

-----BEFORE-----

Screenshot 2026-04-20 at 5 05 50 PM

---------------

----AFTER-----

Screenshot 2026-04-20 at 3 51 49 PM

---------------

Use below schema for testing

```
{
  "$id": "classpath:/schemas/test-activetext-only-review.input-schema.json",
  "title": "IT Cluster Platform Access Request Input Schema",
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "infoMessage": {
      "type": "string",
      "ui:widget": "ActiveText",
      "ui:props": {
        "ui:text": "User **$${{identityApi.displayName}}** will be added to the IT Cluster Platform users rover group and thus granted permissions to use the IT Cluster Platform plugin.\n\n**Note:** Group synchronization within IT Developer Hub can take up to an hour. If you experience permission issues after the workflow completes, please allow time for synchronization to complete."
      }
    }
  }
}
```

### DIFF
--- a/workspaces/orchestrator/.changeset/fix-review-empty-formdata.md
+++ b/workspaces/orchestrator/.changeset/fix-review-empty-formdata.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-form-react': patch
+---
+
+fix: avoid Review step crash when form data is empty (for example display-only `ActiveText` with no submitted values). `generateReviewTableData` returns `{}` instead of undefined, and `NestedReviewTable` handles missing data.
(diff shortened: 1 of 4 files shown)


karthikjeeyar (approved):
/approve
/lgtm

PR #2840, opened 2026-04-20
fix(theme): avoid sidebar background color bleeding into page content

karthikjeeyar:

https://redhat.atlassian.net/browse/RHDHBUGS-2981

The PR containss following chagns:

1. Align the navigation sidebar with merged `palette.navigation` and `rhdh.general` colors, including submenu rows and selected/active `BackstageSidebarItem` states. 
2. Removes a theme override that set  `sidebarBackgroundColor` in a way that applied the sidebar background color to main page inset.
3. Introduce a new token `pageInsetBackgroundColor`, so the page inset background can be controlled by config (defaults to appBarBackgroundColor).

## Screenshots:

**Before:**

main-background-inherits-sidbar-color

---

**After:**

sidebar-bg

### DIFF
--- a/workspaces/theme/.changeset/nasty-bears-relate.md
+++ b/workspaces/theme/.changeset/nasty-bears-relate.md
@@ -2,4 +2,4 @@
 '@red-hat-developer-hub/backstage-plugin-theme': patch
 ---
 
-Fix main content incorrectly picking up the sidebar background. Set the page shell to `appBarBackgroundColor` so the page inset matches the global header; keep main content on `mainSectionBackgroundColor`.
+Align the navigation sidebar with merged `palette.navigation` and `rhdh.general` colors, including submenu rows and selected/active `BackstageSidebarItem` states. Add `rhdh.general.pageInsetBackgroundColor` so the page inset shell can use its own color (defaults match the previous app bar fill; falls back to `appBarBackgroundColor` when unset). Main content area remains on `mainSectionBackgroundColor`.
--- a/workspaces/theme/.changeset/nasty-bears-relate.md
+++ b/workspaces/theme/.changeset/nasty-bears-relate.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+Fix main content incorrectly picking up the sidebar background. Set the page shell to `appBarBackgroundColor` so the page inset matches the global header; keep main content on `mainSectionBackgroundColor`.
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -22,6 +22,7 @@ import { type CSSObject } from '@mui/material/styles';
 
 import { ThemeConfig, ThemeConfigOptions, RHDHThemePalette } from '../types';
 import { redHatFontFaces, redHatFonts } from '../fonts';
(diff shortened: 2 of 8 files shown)


workspaces/theme/plugins/theme/src/utils/createComponents.ts
```diff
@@ -728,7 +728,6 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         root: {
           // Controls the page inset as in PF6 -- only in desktop view
           '@media (min-width: 600px)': {
-            backgroundColor: general.sidebarBackgroundColor,
```
logonoff, line 731: the background color must be set, looks like the global header and the "page inset" background colour is mismatched now. either make it the sidebar background color, the global header background color, or add a new token

karthikjeeyar: Thanks @logonoff for the review.

Setting the background color to `appBarBackgroundColor` matches it with global-header, does this works?

image

logonoff: I'm good with that

karthikjeeyar: Updated it.

workspaces/theme/.changeset/nasty-bears-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+remove misplaced `sidebarBackgroundColor` so main content no longer inherits the sidebar background
```
logonoff: remember to update the changeset

logonoff (approved):

karthikjeeyar:
@christoph-jerolimov @logonoff The user is trying to set the sidebar color using the below config, I see we have two tokens `palette.navigation.backgroundColor` , `rhdh.general.sidebarBackgroundColor` for setting the background color but the one from palette is not used at the moment.

```
       palette:
          navigation:
            background: "#222427"  # This is ignored at the moment.
            color: "#345820"
            selectedColor: "#ffffff"
            navItem:
              hoverBackground: "#333333"
```

Although text, selected colors are picked from `palette.navigation`  but background is always ignored. Since we have multiple tokens available to set the background color and there are chances where some of the users using one over the other and expecting it work.


My proposal is, can we make a utility function that detects if the user has set the background color using `palette` or `rhdh.general` by comparing the final user value against the default theme values, this will allows us to provide predictable outcome using different tokens and we can also safely backport the issue.

Precendence order:

1. If `navigation.background` is set, it will be picked up
2. if `rhdh.sidebarBackgroundColor`, it will be picked up
3. if both are configured, then `rhdh`config wins.

karthikjeeyar:
@logonoff I have updated the PR above changes and introduced a new token for pageInsetBackgroundColor.

**Default RHDH without any customizations:**

image

---

**Customized sidebar background color states:**

```
  branding:
   theme:
    light:
        mode: "light"
        palette:
          navigation:
            background: "#222427"
            color: "#fffff"
            selectedColor: "#ffffff"
            submenu:
              background: "#222427"
            navItem:
              hoverBackground: "#333333"
```

    
image

logonoff (approved):
Haven't tested but conceptually I think this would be a good solution

lokanandaprabhu (approved):
/lgtm

PR #2847, opened 2026-04-21
Skip SSO Authorization

asmasarw:
The backend proxy now skips the SSO token exchange when `clientId` or
`clientSecret` are absent, forwarding requests to the API gateway without
an Authorization header instead of failing with "Failed to obtain upstream
access token."

### DIFF
--- a/workspaces/dcm/plugins/dcm-backend/src/util/tokenUtil.ts
+++ b/workspaces/dcm/plugins/dcm-backend/src/util/tokenUtil.ts
@@ -51,12 +51,20 @@ export const getTokenFromApi = async (
     return cachedToken;
   }
 
+  const clientId = config.getOptionalString('dcm.clientId') ?? '';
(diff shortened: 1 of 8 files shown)


workspaces/dcm/plugins/dcm-backend/src/util/tokenUtil.ts
```diff
@@ -51,12 +51,20 @@ export const getTokenFromApi = async (
     return cachedToken;
   }
 
+  const clientId = config.getOptionalString('dcm.clientId') ?? '';
```
mareklibra, line 54: Please make sure that these are set as `optional`  in the config.d.ts.

mareklibra (commented):
Minor comment about config.d.ts : https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/dcm/plugins/dcm-backend/config.d.ts#L45

mareklibra (approved):

PR #2849, opened 2026-04-21
chores(x2a): split x2a-node services.ts

mareklibra:
Refactoring only for better maintenance, no functional change.


elai-shalev (approved):
lgtm

PR #2855, opened 2026-04-21
feat(homepage): add permission-aware default cards backend

christoph-jerolimov:

WIP


workspaces/homepage/plugins/homepage-backend/src/defaultWidgets/buildUserContext.ts
```diff
@@ -0,0 +1,73 @@
...
+    (userEntity?.relations ?? [])
+      .filter(relation => relation.type === RELATION_MEMBER_OF)
+      .map(relation => relation.targetRef),
+  );
+
+  const permissionDecisions = new Map<string, PermissionDecision>();
+  if (referencedPermissions.size > 0) {
+    const names = [...referencedPermissions];
+    const requests = names.map(name => ({
+      permission: createPermission({ name, attributes: {} }),
```
karthikjeeyar, line 61: In homepage-common, `homepageDefaultWidgetsReadPermission` is defined as a resource permission. This means it expects a resourceRef (like a widget ID) to evaluate conditional policies correctly.

Currently, the homepage backend still checks this via `createPermission` with empty attributes, which is just a basic global name-level check. This won't trigger any logic that relies on specific card IDs.

Should we be adding a second `permissions.authorize` call and pass the resourceRef for each card to support these conditional policies? Or is `if.permissions` meant to stay high-level? I

It’s worth clarifying the intended model in the documentation

workspaces/homepage/plugins/homepage-common/src/permissions.ts
```diff
@@ -0,0 +1,48 @@
...
+import {
+  createPermission,
+  ResourcePermission,
+} from '@backstage/plugin-permission-common';
+
+/**
+ * @public
+ */
+export const RESOURCE_TYPE_HOMEPAGE_DEFAULT_WIDGET = 'homepage-default-card';
+
```
karthikjeeyar: nit:

maybe we should align this with a common naming ? "card" vs "widget" - Everywhere else we have widget in the naming.

karthikjeeyar (commented):
This is a strong step forward 🎉  This feature turns  "everyone sees the same static tiles" to "you see what you are meant to see", I have some small comments, but nothing takes away the overall approach.

workspaces/homepage/plugins/homepage-common/src/permissions.ts
christoph-jerolimov: renamed

```
/**
 * @public
 */
export const RESOURCE_TYPE_HOMEPAGE_DEFAULT_WIDGET = 'homepage-default-widget';
```

In general, I've renamed (hopefully) all new types from cards to widgets.

workspaces/homepage/plugins/homepage-backend/src/permissions/rules.ts
```diff
@@ -0,0 +1,62 @@
...
+
+export type HomepageDefaultWidgetFilter = {
+  key: string;
+  values: Array<string> | undefined;
+};
+
+type HasCardIdParams = { cardIds?: string[] | undefined };
+
+const hasCardId = createPermissionRule({
+  name: 'HAS_CARD_ID',
```
karthikjeeyar: Resource permission is not enforced yet in this new implementation, should we have a followup ticket to handle conditional policies?

Also can we rename this to widget id as well ?
```suggestion
  name: 'HAS_WIDGET_ID',
```

workspaces/homepage/app-config.yaml
```diff
@@ -81,6 +81,52 @@ app:
...
+      ref: VisitCard
+      layout:
+        xl: { w: 12, h: 5 }
+        lg: { w: 12, h: 5 }
+        md: { w: 12, h: 6 }
+        sm: { w: 12, h: 7 }
+        xs: { w: 12, h: 9 }
+        xxs: { w: 12, h: 14 }
+    - id: onboarding
+      ref: Onboarding
```
karthikjeeyar: This should match with the mountpoint ID 'rhdh-onboarding-section`

karthikjeeyar (commented):
Changes looks good to me. 


1. Conditional RBAC policy isn't handled. can we have followup ticket for this?
2. Good to have a small blurb in plugin documentation to capture the new implementation `homepage.defaultWidgets`
3. some test cases are failing.

karthikjeeyar (approved):

karthikjeeyar (approved):
We can address these in followup issues:

- Conditional RBAC policy isn't handled. can we have followup ticket for this?
- Good to have a small blurb in plugin documentation to capture the new implementation homepage.defaultWidgets


/lgtm

PR #2861, opened 2026-04-22
RHIDP-13056: separating notebooks and lightspeed vector_store

JslYoon:


workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
```diff
@@ -65,10 +65,7 @@ export async function createNotebooksRouter(
   const notebooksRouter = Router();
   notebooksRouter.use(express.json());
 
-  const lightSpeedPort =
-    config.getOptionalNumber('lightspeed.servicePort') ??
-    DEFAULT_LIGHTSPEED_SERVICE_PORT;
```
Jdubrick, line 70: We need this AFAIK, if the port changes youre going to fallback to the default `8080`

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/openapi.yaml
```diff
@@ -1,1147 +0,0 @@
-openapi: 3.0.3
```
Jdubrick, line 1: Did this get moved elsewhere? Or was it not supposed to be present?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/constant.ts
```diff
@@ -45,7 +46,6 @@ Make no mistakes.
 /**
  * HTTP and networking constants
  */
-export const LIGHTSPEED_SERVICE_HOST = '0.0.0.0'; // Lightspeed core service host
```
Jdubrick, line 48: I think we keep this as `0.0.0.0` instead of `localhost`. wdyt @karthikjeeyar @yangcao77

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -107,6 +108,12 @@ export async function createRouter(
     DEFAULT_LIGHTSPEED_SERVICE_PORT;
   const system_prompt = config.getOptionalString('lightspeed.systemPrompt');
 
+  const vectorStoresOperator = new VectorStoresOperator(
+    `http://0.0.0.0:${port}`,
+    logger,
+  );
```
Jdubrick: Is this a separate service?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -589,6 +596,17 @@ export async function createRouter(
...
+
+        // get the vector store id for the rhdh-product-docs vector store
+        if (lightspeed_vector_store_id === '') {
+          const vectorStores = await vectorStoresOperator.vectorStores.list();
+          lightspeed_vector_store_id =
+            vectorStores.data.find((v: any) =>
+              v.name.startsWith('rhdh-product-docs'),
+            )?.id || '';
+        }
+        request.body.vector_store_ids = ['asdf'];
```
Jdubrick: `asdf`? is this a leftover from testing?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -605,7 +623,7 @@ export async function createRouter(
           settingsStore,
           user_id,
         );
-
+        console.log('requestBodyasdf', requestBody);
```
Jdubrick: Lets remove console.logs

workspaces/lightspeed/app-config.yaml
```diff
@@ -19,13 +19,13 @@ organization:
 # Disable AI Notebooks feature by default
 lightspeed:
   notebooks:
-    enabled: false
+    enabled: true
```
Jdubrick: notebooks should be disabled by default

workspaces/lightspeed/app-config.yaml
```diff
@@ -19,13 +19,13 @@ organization:
...
     queryDefaults:
-      model: redhataillama-31-8b-instruct
+      model: redhataillama-31-8b-instruct ## move these to run.yaml, provider id needs to match
       provider_id: vllm
     sessionDefaults:
       provider_id: notebooks
-      embedding_model: ${LLAMA_STACK_EMBEDDING_MODEL}
+      embedding_model: sentence-transformers/all-mpnet-base-v2
       embedding_dimension: 768
```
Jdubrick: Do we need this here? Can we leverage the embedding model that is attached with Lightspeed?

workspaces/lightspeed/app-config.yaml
```diff
@@ -19,13 +19,13 @@ organization:
 # Disable AI Notebooks feature by default
 lightspeed:
   notebooks:
-    enabled: false
+    enabled: true
     queryDefaults:
-      model: redhataillama-31-8b-instruct
+      model: redhataillama-31-8b-instruct ## move these to run.yaml, provider id needs to match
       provider_id: vllm
```
Jdubrick: Will there be a follow up PR for this? If so I would favour an env var to be set here rather than a hardcoded value

Jdubrick:
@JslYoon can we add a changeset too for a patch bump please

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
JslYoon: added!

workspaces/lightspeed/plugins/lightspeed-backend/src/service/constant.ts
JslYoon: the lightspeed-configs run showed localhost, but would it be 0.0.0.0 on the actual config?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
JslYoon: Yep, will take it out

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
JslYoon: sorry, forgot to take out test code

workspaces/lightspeed/app-config.yaml
JslYoon: yep, will use that

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
JslYoon: added!

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/openapi.yaml
JslYoon: the documentation was outdated, not supposed to be present

workspaces/lightspeed/app-config.yaml
JslYoon: that comment should not be there, my bad

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
JslYoon: this is to handle llama stack vector_stores, that is proxied by lightspeed.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/constant.ts
Jdubrick: `0.0.0.0` and `localhost` would work the same in our case because there is no ingress or service setup for LCORE to leave the Pod, but `0.0.0.0` binds to available addresses so if we ever added a service it'd be accessible via that whereas `localhost` would not

I more-so suggested it for consistency, since we are using `0.0.0.0` in multiple other spots in this plugin, also, if we ever want to move LCORE out of the RHDH Pod and have as its own deployment, it saves us a code change. 

I will leave this as an optional change, up to you.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -107,6 +108,12 @@ export async function createRouter(
     DEFAULT_LIGHTSPEED_SERVICE_PORT;
   const system_prompt = config.getOptionalString('lightspeed.systemPrompt');
 
+  const vectorStoresOperator = VectorStoresOperator.getInstance(
+    `http://0.0.0.0:${port}`,
```
Jdubrick, line 112: First param is supposed to be the LCORE instance but we hardcode `0.0.0.0` here, should try and make this consistent with that new default constant

workspaces/lightspeed/app-config.yaml
Jdubrick: My comment was about the hardcoded values here

Jdubrick (changes requested):

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
JslYoon: This is consistent with how it's set up in router.ts, I can change the file to make it all consistent though

workspaces/lightspeed/app-config.yaml
```diff
@@ -23,10 +23,6 @@ lightspeed:
     queryDefaults:
       model: redhataillama-31-8b-instruct
       provider_id: vllm
```
Jdubrick: These should be env vars it assumes everyone running the plugin has this model setup at this provider?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/sessions/sessionService.ts
```diff
@@ -43,25 +41,11 @@ export class SessionService {
...
-      config?.getNumber(
-        'lightspeed.notebooks.sessionDefaults.embedding_dimension',
-      ),
-      'lightspeed.notebooks.sessionDefaults.embedding_dimension',
-    );
+    // Use optional config with default fallback
+    this.providerId =
+      config?.getOptionalString(
+        'lightspeed.notebooks.sessionDefaults.provider_id',
+      ) ?? 'notebooks';
```
Jdubrick: In the app config you got rid of `sessionDefaults` entirely, do we need to check for the optional string then or just set it

workspaces/lightspeed/app-config.yaml
JslYoon: these should be env vars, I'll make the changes

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/sessions/sessionService.ts
JslYoon: It's honestly optional, in case users modify the notebooks llama stack config they would need to match it.

JslYoon: deleted for now

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/documents/documentService.ts
```diff
@@ -102,23 +101,16 @@ export class DocumentService {
    * Upload a file to the Files API
    * @param content - File content as string
    * @param title - File title/name
-   * @param fileType - Optional file type for MIME type detection
    * @returns File ID from the Files API
    * @throws Error if upload fails
    */
```
Jdubrick: Can we remove one of these? It's duplicated

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -589,6 +596,17 @@ export async function createRouter(
...
+
+        // get the vector store id for the rhdh-product-docs vector store
+        if (lightspeed_vector_store_id === '') {
+          const vectorStores = await vectorStoresOperator.vectorStores.list();
+          lightspeed_vector_store_id =
+            vectorStores.data.find((v: any) =>
+              v.name.startsWith('rhdh-product-docs'),
+            )?.id || '';
+        }
+        request.body.vector_store_ids = [lightspeed_vector_store_id];
```
Jdubrick: Curious what happens if it can't find the vector store, it looks like we will send the `vector_store_ids` as an array with 1 element of an empty string? I think the addition of this `vector_store_ids` body should be conditional on the vector store being found? wdyt.

JslYoon: Made the changes

workspaces/lightspeed/plugins/lightspeed-backend/README.md
```diff
@@ -110,15 +110,8 @@ lightspeed:
...
-      model: llama3.1-8b-instruct # Model to use for answering queries
-      provider_id: ollama # AI provider for the query model
-
-    # Required: Session defaults for creating vector stores
-    # All three fields are required when Notebooks is enabled
-    sessionDefaults:
-      provider_id: notebooks # Vector store provider ID (must match Llama Stack config)
-      embedding_model: sentence-transformers/all-mpnet-base-v2 # Model for generating embeddings
-      embedding_dimension: 768 # Embedding vector dimension (must match model output)
+      model: ${NOTEBOOKS_QUERY_MODEL} # Model to use for answering queries. Must map to a model inabled in your Lightspeed Stack run.yaml
```
Jdubrick:
```suggestion
      model: ${NOTEBOOKS_QUERY_MODEL} # Model to use for answering queries. Must map to a model available through the provider set in $NOTEBOOKS_QUERY_PROVIDER_ID
```

workspaces/lightspeed/plugins/lightspeed-backend/README.md
```diff
@@ -110,15 +110,8 @@ lightspeed:
...
+      provider_id: ${NOTEBOOKS_QUERY_PROVIDER_ID} # AI provider for the query model. Must map to a provier inabled in your Lightspeed Stack run.yaml
```
Jdubrick:
```suggestion
      provider_id: ${NOTEBOOKS_QUERY_PROVIDER_ID} # AI provider for the query model. Must map to a provider enabled in your Lightspeed config.yaml
```

workspaces/lightspeed/plugins/lightspeed-backend/README.md
```diff
@@ -143,11 +136,7 @@ lightspeed:
...
 
-**Session Defaults** _(required when enabled)_:
-
-- **`sessionDefaults.provider_id`** _(required)_: Vector store provider identifier. Must match a provider configured in your Llama Stack instance (e.g., `notebooks`, `chromadb`). This determines where document embeddings are stored.
-- **`sessionDefaults.embedding_model`** _(required)_: The embedding model to use for converting documents to vectors (e.g., `sentence-transformers/all-mpnet-base-v2`). Must be available in Llama Stack.
-- **`sessionDefaults.embedding_dimension`** _(required)_: Dimension of the embedding vectors produced by the embedding model. Must match the model's output dimension (commonly `768`, `384`, or `1536`).
+> **Important**: The `model` and `provider_id` values must map to a provider and model that are actually enabled in your Lightspeed Stack run.yaml configuration. If the provider or model is not available in Lightspeed Stack, queries will fail. For example, if `openai` enabled in Lightspeed via ENABLE_OPENAI, then model must be available (model=gpt-4o-mini).
```
Jdubrick:
```suggestion
> **Important**: The `model` and `provider_id` values must map to a provider and model that are actually enabled in your Lightspeed config.yaml configuration. If the provider or model is not available in Lightspeed, queries will fail. For example, if `openai` enabled in Lightspeed via ENABLE_OPENAI, then model must be available, e.g (model=gpt-4o-mini).
```

Jdubrick (approved):
/lgtm

PR #2873, opened 2026-04-22
[Konflux] Fix Latest Releases not shown when applications field is omitted or uses wildcard patterns.

rrosatti:
### Description

In this PR we're fixing a bug where the "Latest Releases" card (shown in Overview tab) was not showing any release when the `applications` field was omitted or used wildcard patterns.

Fixes https://redhat.atlassian.net/browse/KFLUXUI-1230

### How to test it?

- create a backstage entity/component in which the `konflux-ci.dev/clusters` either have an `applications` field that uses wildcard patterns or ommit the `applications` field. Example:
```yaml
kind: Component
apiVersion: backstage.io/v1alpha1
metadata:
  name: component-1
  description: Component 1
  title: Component 1
  annotations:
    konflux-ci.dev/overview: 'true'
    konflux-ci.dev/konflux: 'true'
    konflux-ci.dev/ci: 'true'
    konflux-ci.dev/clusters: |
      - cluster: <your-cluster>
        namespace: <your-namespace>
spec:
  type: service
  lifecycle: experimental
  owner: user:guest
``` 
- navigate to `Overview` tab
- notice that now the `Konflux Latest Releases` UI component should show the latest release given your config
- :coffee:

### Visual references

Before:

https://github.com/user-attachments/assets/956b0407-865d-480a-8bcc-19659baa8974

After:

https://github.com/user-attachments/assets/29073189-f010-4a20-bcf4-3017da75e9fc

### DIFF
--- a/workspaces/konflux/plugins/konflux/src/components/List/LatestReleasesList/utils.ts
+++ b/workspaces/konflux/plugins/konflux/src/components/List/LatestReleasesList/utils.ts
@@ -16,6 +16,7 @@
 
 import {
   getApplicationFromResource,
+  matchesApplicationPattern,
   ReleaseResource,
   SubcomponentClusterConfig,
 } from '@red-hat-developer-hub/backstage-plugin-konflux-common';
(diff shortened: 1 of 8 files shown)


workspaces/konflux/plugins/konflux/src/components/List/LatestReleasesList/utils.ts
milantaky, line 70: nit: I think this check isn't necessary since the length is checked on `filteredReleases` and the sort function doesn't remove any items :)

milantaky (approved):
Left a small nit, otherwise releases show as expected with no applications, pattern, and exact name:)

workspaces/konflux/plugins/konflux/src/components/List/LatestReleasesList/utils.ts
rrosatti: good catch! in theory it does not need, but I did that check just to "shut up Typescript" :D

PR #2896, opened 2026-04-23
fix(orchestrator): workflow start state might also be in the start.stateName variable when compiled and cloudevent input data should be one level deep

lholmquist:

While i was testing the Emitting Cloud Events to Trigger workflows on an Openshift cluster,  using "compiled" workflows instead of running them in dev mode on my laptop, I ran into this error: "Error executing workflow with id lock-flow, No States that match the start state"

I discovered that the "start" parameter might also have another parameter in it that needs to be taken into account.

For example,  The previous code assumed that when parsing the source for the "start" param it would look like this:

```
id: lock-flow
name: Kafka and Cloud events demo
...
start: listenToLock
```
code example from here: https://github.com/masayag/poc-kafka-logic-operator/blob/main/callback-flow/src/main/resources/lock.sw.yaml

When i ran the "compiled" version of this workflow,  using the manifests that are generated when building the workflow, the source looks more like this:

```
start:
      stateName: listenToLock
```
code example from here: https://github.com/masayag/poc-kafka-logic-operator/blob/main/manifests/02-sonataflow_lock-flow.yaml

This code change checks for that "stateName" param and uses it if available, it not then falls back to just the "start" param

The other thing that I came across was when a workflow has input data from a form sent to it.  The inputData object will look like this:

```
{
  workflowData: {
    input1: Data,
    input2: moreData
  }
}
```
and the data section of the cloud event would look something like this: `data: { workflowData: {} }` which will error the workflow when the kafka message sends the data back.  The input data that is added to the data section of the cloudevent should just be the workflowData level and not the nested object.  I added that fix here too

### DIFF
--- a/workspaces/orchestrator/.changeset/fast-pens-explode.md
+++ b/workspaces/orchestrator/.changeset/fast-pens-explode.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-backend': patch
+---
+
+fix: inputdata added to the cloudevent should be one level deep and not be inputData.workflowData
--- a/workspaces/orchestrator/.changeset/fast-pens-explode.md
+++ b/workspaces/orchestrator/.changeset/fast-pens-explode.md
@@ -2,4 +2,4 @@
 '@red-hat-developer-hub/backstage-plugin-orchestrator-backend': patch
 ---
 
-fix: inputdata added to the cloudevent should be one level deep and not be inputData.workflowData
+fix: input data added to the cloudevent should be one level deep and not be inputData.workflowData
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
@@ -277,7 +277,6 @@ describe('SonataFlowService', () => {
       expect(result?.id).toBe('12345');
     });
 
-
     it('should error on a bad connection', async () => {
       const kafkaServiceOptionsMock: OrchestratorKafkaServiceOptions = {
         clientId: 'kafkaClientId',
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
@@ -223,6 +223,54 @@ describe('SonataFlowService', () => {
       expect(result?.id).toEqual('12345');
     });
 
+    it('should pass data from workflowdata payload into the clouevent data', async () => {
+      const kafkaServiceOptionsMock: OrchestratorKafkaServiceOptions = {
+        clientId: 'kafkaClientId',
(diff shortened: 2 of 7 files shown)


workspaces/orchestrator/.changeset/fast-pens-explode.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-backend': patch
+---
+
+fix: inputdata added to the cloudevent should be one level deep and not be inputData.workflowData
```
AndrienkoAleksandr: Small typo:

```suggestion
fix: input data added to the cloudevent should be one level deep and not be inputData.workflowData
```

workspaces/orchestrator/plugins/orchestrator-backend/src/service/SonataFlowService.test.ts
```diff
@@ -223,6 +223,54 @@ describe('SonataFlowService', () => {
...
+              paramter1: '12345',
+              isEvent: true,
+            },
+          },
+        });
+      expect(spy).toHaveBeenCalled();
+      expect(result).toBeDefined();
+      expect(result?.id).toBeDefined();
+      expect(result?.id).toEqual('12345');
+    });
```
AndrienkoAleksandr, line 278: Sorry, but placing assertions inside mocks is considered bad practice. It would be better to:
```suggestion
    it('should pass data from workflowdata payload into the clouevent data', async () => {
      const kafkaServiceOptionsMock: OrchestratorKafkaServiceOptions = {
        clientId: 'kafkaClientId',
        brokers: ['localhost:9091'],
      };
      const sonataFlowServiceWithKafka = new SonataFlowService(
        dataIndexServiceMock,
        loggerMock,
        kafkaServiceOptionsMock,
      );

      const sendMock = jest.fn();

      jest
        .spyOn(
          sonataFlowServiceWithKafka.getOrchestratorKafkaImpl() as any,
          'producer',
        )
        .mockImplementation(() => ({
          connect: jest.fn(),
          send: sendMock,
          disconnect: jest.fn(),
        }));

      const result =
        await sonataFlowServiceWithKafka.executeWorkflowAsCloudEvent({
          definitionId,
          workflowSource: 'workflowSource',
          workflowEventType: 'workflowEventType',
          contextAttribute: 'lockid',
          inputData: {
            workflowdata: {
              paramter1: '12345',
              isEvent: true,
            },
          },
        });

      expect(sendMock).toHaveBeenCalled();

      const { messages } = sendMock.mock.calls[0][0];
      const parsed = JSON.parse(messages[0].value);

      expect(parsed.data).toEqual({
        paramter1: '12345',
        lockid: '12345',
      });

      expect(parsed.isEvent).toBeUndefined();

      expect(result).toBeDefined();
      expect(result?.id).toBe('12345');
    });

```

lholmquist: thanks for the suggestion!

lholmquist:
@AndrienkoAleksandr thanks for the suggestions,  updated

AndrienkoAleksandr (approved):

PR #2897, opened 2026-04-23
feat(extensions): enforce collision policy for duplicate entity identities

hopehadfield:

- Add collision handling for extensions catalog entities in the backend provider by keying on normalized kind/namespace/name.
- Keep the first definition when duplicates are equivalent, and fail fast when duplicates with the same identity have conflicting content.
- Add targeted unit tests covering equivalent duplicates, conflicting duplicates, and same-name entities across different namespaces.

This change is being made as a result of [RHIDP-12376](https://redhat.atlassian.net/browse/RHIDP-12376). Multi-catalog ingestion can introduce duplicate entity identities; this change makes behaviour deterministic and prevents silent shadowing of conflicting definitions.

Manually validated runtime behavior in three scenarios:

1. equivalent duplicate -> warning + continue
2. conflicting duplicate -> provider error
3. same name with different namespace -> both accepted

## How to Test

Create a directory `/tmp/extensions-test`, with subdirectories `primary/catalog-entities/plugins` and `extra/community/catalog-entities/plugins`

For **case A** (equivalent duplicate), create an identical plugin yaml file on each path, for example:
`/tmp/extensions-collision-test/primary/catalog-entities/plugins/github.yaml` and `/tmp/extensions-collision-test/extra/community/catalog-entities/plugins/github.yaml` both containing

```
apiVersion: extensions.backstage.io/v1alpha1
kind: Plugin
metadata:
  name: github
spec:
  owner: redhat
 ```
 
 After running `yarn start-backend` you should see in the logs that the duplicate entity was skipped, with the first definition being kept. The backend continues successfully.

For **case B** (conflicting duplicate), update the `spec.owner` in `/tmp/extensions-collision-test/extra/community/catalog-entities/plugins/github.yaml` to `community` and restart the backend. You should see that the provider fails with a conflicting entities message surfaced in the logs.

### DIFF
--- a/workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
+++ b/workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
@@ -13,11 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import { SchedulerServiceTaskRunner } from '@backstage/backend-plugin-api';
+import {
+  LoggerService,
+  SchedulerServiceTaskRunner,
+} from '@backstage/backend-plugin-api';
 import {
   ANNOTATION_LOCATION,
   ANNOTATION_ORIGIN_LOCATION,
   Entity,
+  stringifyEntityRef,
 } from '@backstage/catalog-model';
 import {
   EntityProvider,
@@ -39,26 +43,23 @@ export abstract class BaseEntityProvider<T extends Entity>
   private connection?: EntityProviderConnection;
   private taskRunner: SchedulerServiceTaskRunner;
   private config?: Config;
+  private logger?: LoggerService;
 
   private static readonly EXTENSIONS_DIRECTORY = '/extensions';
   private static readonly DEPRECATED_MARKETPLACE_DIRECTORY = '/marketplace';
 
-  constructor(taskRunner: SchedulerServiceTaskRunner, config?: Config) {
+  constructor(
+    taskRunner: SchedulerServiceTaskRunner,
+    config?: Config,
+    logger?: LoggerService,
+  ) {
     this.taskRunner = taskRunner;
     this.config = config;
+    this.logger = logger;
   }
 
   abstract getProviderName(): string;
   abstract getKind(): string;
-
-  private getEntityIdentity(entity: Entity): string {
-    const namespace = entity.metadata.namespace ?? 'default';
-    return [
-      entity.kind.toLocaleLowerCase('en-US'),
-      namespace.toLocaleLowerCase('en-US'),
-      entity.metadata.name.toLocaleLowerCase('en-US'),
-    ].join('/');
-  }
 
   private addProviderAnnotations(entity: T): T {
     return {
@@ -79,7 +80,7 @@ export abstract class BaseEntityProvider<T extends Entity>
       return [];
     }
 
-    const entitiesByIdentity = new Map<
+    const entitiesByEntityRef = new Map<
       string,
       { entity: T; filePath: string }
     >();
@@ -89,10 +90,14 @@ export abstract class BaseEntityProvider<T extends Entity>
         continue;
       }
 
-      const identity = this.getEntityIdentity(fileData.content);
-      const existing = entitiesByIdentity.get(identity);
+      const identity = stringifyEntityRef({
+        kind: fileData.content.kind,
+        namespace: fileData.content.metadata.namespace ?? 'default',
+        name: fileData.content.metadata.name,
+      }).toLocaleLowerCase('en-US');
+      const existing = entitiesByEntityRef.get(identity);
       if (!existing) {
-        entitiesByIdentity.set(identity, {
+        entitiesByEntityRef.set(identity, {
           entity: fileData.content,
           filePath: fileData.filePath,
         });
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by debsmita1** [comment]
@hopehadfield Could you add some examples to test your PR ?

**Comment by hopehadfield** [comment]
@debsmita1 I included a "How to Test" section in the PR, please let me know if you have any questions

File: workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts [hunk 1]
```diff
@@ -50,23 +51,69 @@ export abstract class BaseEntityProvider<T extends Entity>
   abstract getProviderName(): string;
   abstract getKind(): string;
 
+  private getEntityIdentity(entity: Entity): string {
```
dzemanov: Can we use  `stringifyEntityRef` from '@backstage/catalog-model'  ?

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
```diff
@@ -50,23 +51,69 @@ export abstract class BaseEntityProvider<T extends Entity>
...
-          ...file.content.metadata,
-          annotations: {
-            ...file.content.metadata.annotations,
-            [ANNOTATION_LOCATION]: `file:${this.getProviderName()}`,
-            [ANNOTATION_ORIGIN_LOCATION]: `file:${this.getProviderName()}`,
-          },
-        },
-      }));
+
+    const entitiesByIdentity = new Map<
```
dzemanov: Minor: might use `entitiesByEntityRef`

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
```diff
@@ -50,23 +51,69 @@ export abstract class BaseEntityProvider<T extends Entity>
...
+        console.warn(
+          `Skipping duplicate Extensions entity '${identity}' from '${fileData.filePath}'. Keeping first definition from '${existing.filePath}'.`,
+        );
+        continue;
+      }
+
+      throw new Error(
+        `Conflicting Extensions entities detected for '${identity}' in '${existing.filePath}' and '${fileData.filePath}'.`,
+      );
+    }
```
dzemanov: Would it make sense to do what Catalog [1](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/database/DefaultProcessingDatabase.ts#L373), [2](https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/database/DefaultProviderDatabase.ts#L184-L193) does in case there is conflicting entity: log warning via `logger.warn` and skip ingesting this entity, all other entities will be ingested instead of error?

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
```diff
@@ -50,23 +51,69 @@ export abstract class BaseEntityProvider<T extends Entity>
...
+      if (!existing) {
+        entitiesByIdentity.set(identity, {
+          entity: fileData.content,
+          filePath: fileData.filePath,
+        });
+        continue;
+      }
+
+      if (isDeepStrictEqual(existing.entity, fileData.content)) {
+        console.warn(
```
dzemanov: Can you use `logger.warn`?

dzemanov (changes requested):
THank you @hopehadfield for these changes, I left some small comments.

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
hopehadfield: Do you want me to also migrate the existing `console.warn`/`error` calls in this provider in the same PR?

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
hopehadfield: My thought here was to avoid silent failures, but ultimately I think you're right that it's probably best still to ingest all the other entities. Less damaging to having a missing entity with an explanation in the logs. I'll update this accordingly

hopehadfield:
Thanks for the review @dzemanov! I've addressed your comments, please let me know if there's anything else required before merging

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
```diff
@@ -38,35 +43,83 @@ export abstract class BaseEntityProvider<T extends Entity>
...
+  private addProviderAnnotations(entity: T): T {
+    return {
+      ...entity,
+      metadata: {
+        ...entity.metadata,
+        annotations: {
+          ...entity.metadata.annotations,
+          [ANNOTATION_LOCATION]: `file:${this.getProviderName()}`,
+          [ANNOTATION_ORIGIN_LOCATION]: `file:${this.getProviderName()}`,
+        },
```
dzemanov, line 73: Should this be filename instead?

dzemanov: Oh, I see it being used the same way [here](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts#L170), so it's correct.

dzemanov (changes requested):
Tested that this works fine, thank you @hopehadfield!

Tested with:
- duplicate entities
- conflicting entities

setup:
```
extensions:
  directory: /tmp/extensions-test
```
I see this optional configuration is not in `config.d.ts`, would be good to mention it there 😅. I can create a small following PR or we can add it in this PR.

It would be good to fix this [test entity](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/extensions/examples/plugins/certified-plugin-2-by-vendor-a.yaml#L6) invalid name before merging this PR (I noticed it now thanks to catalog warn log).

If you have time, there are some sonar issues flagged that would be good to fix.

Otherwise, good to go!

workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
dzemanov: Would be good to update it now if you have time

hopehadfield:
Addressed your additional comments @dzemanov, thanks! I just have one question on this note:

> I see this optional configuration is not in config.d.ts

I might be missing which config key you mean. In this PR I didn’t add a new plugin config option (the only existing key here is `extensions.directory`, already in `config.d.ts`).  If I'm mistaken and there is something I should be adding, I'm happy to append it to this PR!

dzemanov (approved):
Thank you @hopehadfield, works great!
I was looking into incorrect config.d.ts 🙈

debsmita1 (approved):

PR #2913, opened 2026-04-24
feat(scorecard): File level checks - all providers

djanickova:

This is the second version of the new filecheck scorecard. The implementation is now provider agnostic and works for all SCMs. It contains mostly the same functionality as https://github.com/redhat-developer/rhdh-plugins/pull/2751, but the logic was moved from the `github` module into a new `filecheck` module.

This PR introduces a new type of scorecard: file level checks ([RHDHPLAN-397](https://redhat.atlassian.net/browse/RHDHPLAN-397)). This adds ability for Scorecard to verify if a list of required files (one or more) is present (such as a license, CODEOWNERS, dockerfile, .gitignore, etc). The paths to these files are configurable.

Example of these scorecards on the entity page:


djanickova:
Tested this by having both Github and Gitlab components in my catalog, scorecards showed up correctly for both of these.

Example for a Gitlab component (I added one more filecheck locally to app-config compared to what is there by default):
Screenshot 2026-04-24 at 11 10 44

Example of an aggregated card (with custom title) detail when I have 2 entities:
Screenshot 2026-04-24 at 16 22 35

Most of the code (especially for scorecard and scorecard-backend packages) remains unchanged compared to https://github.com/redhat-developer/rhdh-plugins/pull/2751.

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
```diff
@@ -0,0 +1,80 @@
...
+
+    try {
+      const tree = await this.urlReader.readTree(target, {
+        etag: cached?.etag,
+        filter: filePath => pathsSet.has(filePath),
+      });
+
+      const files = await tree.files();
+      foundPaths = new Set(files.map(f => f.path));
+      this.cache.set(cacheKey, { etag: tree.etag, foundPaths });
```
christoph-jerolimov: I don't say its wrong, but someone should test this with a big repo. It's probably better to make 3 `readUrl` calls here instead of one big `readTree`. In the GH implementation it looks like it downloads the complete repo for that. Also if its just downloaded to a tmp folder, it's probably saver (security concerns with cloned git repos into RHDH) to just download the README, LICENSE, and what else file instead of the complete repo.

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/metricProviders/FilecheckConfig.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const INVALID_PATH_CHARS = /[\n\r"\\]/;
+
+export function validateFilePath(id: string, path: string): void {
+  if (INVALID_PATH_CHARS.test(path)) {
+    throw new Error(
+      `Invalid file path for '${id}': path must not contain newlines, quotes, or backslashes`,
+    );
+  }
+  if (path.startsWith('/') || path.startsWith('./')) {
```
christoph-jerolimov: should we forbid also `../` ? I don't see a real risk, just to be even more "save" here?

```suggestion
  if (path.startsWith('/') || path.startsWith('./') || path.startsWith('../')) {
```

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/metricProviders/FilecheckConfig.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  const files: FilecheckEntry[] = filesConfig.map(fileConfig => {
+    const keys = fileConfig.keys();
+    if (keys.length !== 1) {
+      throw new Error(
+        'Each file config entry must have exactly one key-value pair',
+      );
+    }
+    const id = keys[0];
```
christoph-jerolimov: Just as an idea: Why not using then an object?

```yaml
    filecheck:
      files:
        license: 'LICENSE'
        codeowners: 'CODEOWNERS'
```

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-filecheck",
+  "version": "0.1.0",
```
christoph-jerolimov: I think you need to switch to 0.0 here to get a 0.1.0 with your minor changeset:

```suggestion
  "version": "0.0.0",
```

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/README.md
```diff
@@ -0,0 +1,126 @@
...
+Define which files to check under `scorecard.plugins.filecheck.files` in your `app-config.yaml`. Each entry is a single key–value pair where the key becomes the metric identifier suffix and the value is the relative file path inside the repository:
+
+```yaml
+# app-config.yaml
+scorecard:
+  plugins:
+    filecheck:
+      files:
+        - readme: README.md
+        - license: LICENSE
```
christoph-jerolimov: LICENSE, LICENSE.txt, ...

It's fine, but really expect that this will come back that each metric should support multiple filenames.

The time might not be enough anymore, but I would recommend:

```yaml
scorecard:
  plugins:
    filecheck:
      readme:
        files: README.md
      license:
        files:
          - LICENSE
          - LICENSE.txt
```

That would also allow you later to add other configurations for a metric. Like `caseInsensitivity: true` or whatever...

workspaces/scorecard/plugins/scorecard/src/utils/translationUtils.ts
```diff
@@ -0,0 +1,54 @@
...
+  if (dotIndex !== -1) {
+    const base = metricId.slice(0, dotIndex);
+    const name = metricId.slice(dotIndex + 1);
+    const templateKey = `metric.${base}.${field}`;
+    const templateTranslated = t(templateKey as any, { name });
+    if (templateTranslated !== templateKey) return templateTranslated;
+  }
+
+  return fallback ?? translated;
+}
```
christoph-jerolimov, line 54: cc @Eswaraiahsapram @its-mitesh-kumar

christoph-jerolimov (commented):
Hi @djanickova, great work 👍 

I didn't tested this PR myself but wanted drop some thoughts for you, and @dzemanov, @imykhno, @PatAKnight because I'm off next week. You can pick or ignore based on your internal or scorecard meeting discussions on Monday:

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
```diff
@@ -0,0 +1,80 @@
...
+    } catch (error: any) {
+      if (cached && error.name === 'NotModifiedError') {
```
PatAKnight:
```suggestion
      if (cached && error instanceof NotModifiedError) {
```

and add:

```typescript
import { NotModifiedError } from '@backstage/errors'; 
```

Since we already have `@backstage/errors` in the packstage.json

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/metricProviders/FilecheckConfig.ts
```diff
@@ -0,0 +1,93 @@
...
+/**
+ * Parses the filecheck configuration from the root Backstage config.
+ * Returns undefined if no files are configured.
+ */
+export function parseFilecheckConfig(
+  config: Config,
+): FilecheckConfig | undefined {
+  const filesConfig = config.getOptionalConfigArray(
+    'scorecard.plugins.filecheck.files',
+  );
```
PatAKnight: I think we need to handle the potential for duplicate keys,  so something like:

```
- readme: README.md
- readme: readme.md
```

would collide silently and with the last matching key winning.

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
```diff
@@ -0,0 +1,80 @@
...
+  async checkFiles(
+    entity: Entity,
+    filePaths: string[],
+  ): Promise<Map<string, boolean>> {
+    const { target } = getEntitySourceLocation(entity);
+
+    const sortedPaths = [...filePaths].sort((a, b) => a.localeCompare(b));
+    const cacheKey = `${target}\0${sortedPaths.join('\0')}`;
+    const pathsSet = new Set(filePaths);
+    const cached = this.cache.get(cacheKey);
```
PatAKnight: It doesn't seem like we remove entries from the cache and don't seem to have a limit anywhere. This will cause it to grow without bound.

workspaces/scorecard/plugins/scorecard-backend/src/scheduler/tasks/PullMetricsByProviderTask.ts
```diff
@@ -141,6 +143,83 @@ export class PullMetricsByProviderTask implements SchedulerTask {
...
+
+              if (enabledMetricIds.length === 0) {
+                return undefined;
+              }
+
+              try {
+                const resultsMap = await provider.calculateMetrics(entity);
+
+                return enabledMetricIds.map(metricId => {
+                  const value = resultsMap.get(metricId)!;
```
PatAKnight: While a small chance, I think having a check here might be worth it. Something like

```typescript
if (!resultsMap.has(metricId)) {
    return [
      {
        catalog_entity_ref: entityRef,
        metric_id: metricId,
        value: undefined,
        timestamp: new Date(),
        error_message: `calculateMetrics() did not return an entry for metric '${metricId}'`,
        entity_kind: entityKind,
        entity_namespace: entityNamespace,
        entity_owner: entityOwner,
      } as DbMetricValueCreate,
    ];
  }
```

This way you could do `const value = resultsMap.get(metricId);` We also protect against malformed scorecard backend modules that might come from outside of our contributions.

PatAKnight (commented):
Thank you for this @djanickova!! Some suggestions

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/README.md
djanickova: I agree that it will be handy to have the option to define multiple paths per file. However, I would keep the changes now at minimum considering we do not have much time left

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/metricProviders/FilecheckConfig.ts
djanickova: Thank you for the suggestion, I think an object would be better and cleaner. I refactored the code to use objects instead

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
djanickova: Thank you for noticing this 🙈 I updated the implementation to use CacheService and pass TTL when calling `set()`. I configured the TTL to be 1 hour (based on how bulk-import does it), but maybe some other value/approach would be more reasonable, wdyt?

Eswaraiahsapram (approved):
Thanks, @djanickova. Reviewed the frontend changes and tested locally. Looks good to me 🎉 

https://github.com/user-attachments/assets/ca6ae7d7-31d9-48c8-b045-0aef215faa4c

workspaces/scorecard/plugins/scorecard/src/utils/translationUtils.ts
Eswaraiahsapram: This one looks good to me, @christoph-jerolimov. Resolving it.

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
PatAKnight: I think 1 hour is fine for now and we can always go back and make it configurable as a future enhancement, if it is needed.

PatAKnight (approved):
LGTM

workspaces/scorecard/plugins/scorecard-backend-module-filecheck/src/clients/FilecheckClient.ts
djanickova: We have discussed this and came to the conclusion to leave it as it is for now. Both readTree and readUrl have disadvantages and with readTree, there is a smaller chance of hitting rate limits. We can also use etags and caching to further optimize readTree. We can always come back to this if we find out that the current approach is not good :)

PR #2927, opened 2026-04-27
feat(x2a): log streaming and copy&paste without row numbers

mareklibra:
Fixes: FLPATH-4087

Stream logs in real-time as backend jobs produce them.

Exclude row numbers from copy selection.

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/src/services/KubeService.ts
+++ b/workspaces/x2a/plugins/x2a-backend/src/services/KubeService.ts
@@ -19,15 +19,20 @@ import {
   createServiceFactory,
   LoggerService,
 } from '@backstage/backend-plugin-api';
+import { Log } from '@kubernetes/client-node';
 import type {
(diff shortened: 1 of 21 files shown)


workspaces/x2a/plugins/x2a-backend/src/services/KubeService.ts
```diff
@@ -397,11 +431,35 @@ export class KubeService implements KubeServiceApi {
...
+        const pass = new PassThrough();
+        const k8sLog = new Log(this.#kubeConfig);
+        const logAbort: AbortController = await k8sLog.log(
+          this.#namespace,
+          podName,
+          containerName,
+          pass,
+          { follow: true },
+        );
+        pass.on('close', () => {
```
elai-shalev, line 452: does this need a timeout?

elai-shalev (approved):
Looks good, if this is properly tested then it's a great addition

PR #2931, opened 2026-04-27
feat(orchestrator): use conditional policies for orchestrator

AndrienkoAleksandr:
## What does this pull request do

Fixes: https://redhat.atlassian.net/browse/RHDHBUGS-2820

This pull request replaces dynamic RBAC basic permissions with conditional policies to improve performance and RBAC UI support. The RBAC UI supports both basic "static" permissions and conditional policies, but does not support "dynamic" basic permissions(`orchestrator.workflow.<workflowId>` or `orchestrator.workflow.use.<workflowId>`). As a result, users cannot create such permissions through the UI, nor can they view or manage them within the interface. Conversely, the UI provides excellent support for conditional policies.

### !Deprecation notes

We leave dynamic basic permission support just for one release, after that we will drop them/

### Migration to conditional policies

We do not provide any auto-migration functionality. Users must manually migrate permissions. All dropped dynamic permissions lose their allow effect after next release. The pull request includes a migration document that covers permissions defined using:

- Permission policy files
- Permissions created through the REST API

This migration guide does not provide information on migrating permissions created through the UI, because the UI has never provided a way to create such dynamic permissions and does not display them even when they are created via the REST API. Therefore, we do not expect users to have created such permissions through the UI.

### How to test it locally:

Use guest user for testing purpose.

# Test conditional policies yaml file.

Provide rbac configuration:

```
permission:
  enabled: true
  rbac:
    policies-csv-file: ../../docs/rbac-policy.csv
    conditionalPoliciesFile: ../../docs/conditional-policies.yaml
    policyFileReload: true
    pluginsWithPermission:
      - orchestrator
    admin:
      users:
        - name: user:default/mareklibra

```

yarn start, go to orchestrator ui - should be dispaly three workflows from ../../docs/rbac-policy.csv. Also you have opportunity execute two of them.

# Test deprecated dynamic permission

Provide configuration:

```
permission:
  enabled: true
  rbac:
    policies-csv-file: ../../docs/rbac-policy.csv
    # disable conditional policies file
    # conditionalPoliciesFile: ../../docs/conditional-policies.yaml
    policyFileReload: true
    pluginsWithPermission:
      - orchestrator
    admin:
      users:
        - name: user:default/mareklibra
```

Uncomment in the ../../docs/rbac-policy.csv permissions:

```
p, role:default/workflowUser, orchestrator.workflow.yamlgreet, read, allow
p, role:default/workflowUser, orchestrator.workflow.wait-or-error, read, allow
p, role:default/workflowUser, orchestrator.workflow.use.yamlgreet, update, allow
p, role:default/workflowUser, orchestrator.workflow.use.wait-or-error, update, allow

```

yarn start, go to the orchestrator ui - should be visible two workflows with ids: yamlgreet and wait-or-error and you are be able to execute them.

# Test general basic permission: all workflows are visible and executable

Uncomment general basic permissions:

```
#p, role:default/workflowUser, orchestrator.workflow.yamlgreet, read, allow
#p, role:default/workflowUser, orchestrator.workflow.wait-or-error, read, allow
#p, role:default/workflowUser, orchestrator.workflow.use.yamlgreet, update, allow
#p, role:default/workflowUser, orchestrator.workflow.use.wait-or-error, update, allow

p, role:default/workflowUser, orchestrator.workflow.use, update, allow
p, role:default/workflowUser, orchestrator.workflow, read, allow
```

yarn start, go to the orchestrator ui - should be visible all workflowsand you are be able to execute all of them. 

Notice: If you are mac OS user, uncomment sonataflow image:

```
container: registry.redhat.io/openshift-serverless-1/logic-swf-devmode-rhel8:1.36.0
```


workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
```diff
@@ -0,0 +1,728 @@
...
+Or manually edit the file and remove lines like:
+
+```csv
+p, role:default/user, orchestrator.workflow.yamlgreet, read, allow
+p, role:default/user, orchestrator.workflow.use.checkout, update, allow
+```
+
+#### Remove policies via RBAC UI
+
+RBAC UI don't allow users create or delete deprecated permissions.
```
lholmquist:
```suggestion
RBAC UI doesn't allow users create or delete deprecated permissions.
```

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
```diff
@@ -0,0 +1,728 @@
...
+
+#### Remove policies via REST API
+
+If you created policies via REST API, then fetch list all policies to find the ones to delete
```
lholmquist:
```suggestion
If you created policies via REST API, then fetch the list of all policies to find the ones to delete
```

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
```diff
@@ -0,0 +1,728 @@
...
+     -H "Authorization: Bearer $TOKEN" \
+     -v
+```
+
+### Step 5: Verify Migration
+
+#### Test that new permissions work
+
+1. **Verify conditional policies were created:**
+   - **YAML file**: Check your `conditional-policies.yaml` file contains the correct entries
```
lholmquist:
```suggestion
   - **YAML file**: Check that your `conditional-policies.yaml` file contains the correct entries
```

workspaces/orchestrator/packages/app/package.json
lholmquist, line 1: This is the same for the other package.json in the packages directory,  should these deps be downgraded.  i know this PR has been open for a while, so i'm assuming that is why it looks that way?

workspaces/orchestrator/packages/backend/package.json
lholmquist, line 1: This is the same for the other package.json in the packages directory,  should these deps be downgraded.  i know this PR has been open for a while, so i'm assuming that is why it looks that way?

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
AndrienkoAleksandr: Fixed in 3612d9509 — applied your grammar suggestions.

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
AndrienkoAleksandr: Fixed in 3612d9509 — applied your grammar suggestions.

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
AndrienkoAleksandr: Fixed in 3612d9509 — applied your grammar suggestions.

workspaces/orchestrator/packages/app/package.json
AndrienkoAleksandr: Good question — these are intentional, not accidental drift. This PR migrates the orchestrator workspace to the NFS app/backend packages and bumps Backstage/RBAC deps so conditional policies can be tested end-to-end (RBAC UI, permission metadata, etc.). Downgrading would undo that test setup. Happy to split the NFS migration into a separate PR if you prefer a smaller diff focused only on conditional policies.

PatAKnight: I believe what @lholmquist is mentioning is that the `@backstage` deps have all been downgraded in this PR. Example: `"@backstage/cli": "^0.36.2",` is now `"@backstage/cli": "^0.36.0",`. But it might be nice to move out all the auth/rbac related additions to their own PR to make it more focused on the conditional policies.

AndrienkoAleksandr: @PatAKnight you are right, I fixed dependencies.

AndrienkoAleksandr: > @PatAKnight: But it might be nice to move out all the auth/rbac related additions to their own PR to make it more focused on the conditional policies. 

ok, I will fix it.

AndrienkoAleksandr: Done.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -121,39 +174,60 @@ const filterAuthorizedWorkflowIds = async (
...
+      matches({ workflowId: id }, filters),
+    );
+    remainingIds = workflowIds.filter(id => !conditionallyAllowed.includes(id));
+  }
 
-  return workflowIds.filter(
-    (_, idx) => decisions[idx].result === AuthorizeResult.ALLOW,
-  );
+  // @deprecated Remove this legacy fallback block in next release
+  if (remainingIds.length > 0) {
```
PatAKnight, line 224: If I am understanding this correctly, then even after going through the process of migrating, there is still a chance to have slow performance because we still check using the legacy fallback anyways. 

I don't know if there is an easy way to solve this but it would be nice if we could figure out a way to handle the case where platform engineers have migrated so that we can prevent the slowness for them.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
```diff
@@ -76,30 +84,75 @@ interface RouterApi {
...
+// @deprecated Remove in next release — legacy dynamic permission fallback
+const legacyAuthorize = async (
+  request: HttpRequest,
+  specificPermission: BasicPermission,
+  permissionsSvc: PermissionsService,
+  httpAuth: HttpAuthService,
+  logger: LoggerService,
+): Promise<boolean> => {
+  const credentials = await httpAuth.credentials(request);
+  const [decision] = await permissionsSvc.authorize(
```
PatAKnight, line 131: This use to be batched. I wonder if the change would add any additional slowness to the performance now that it has been changed.

workspaces/orchestrator/.changeset/fresh-swans-complain.md
```diff
@@ -1,6 +1,10 @@
...
+'@red-hat-developer-hub/backstage-plugin-orchestrator-backend': minor
+'@red-hat-developer-hub/backstage-plugin-orchestrator-common': minor
+'@red-hat-developer-hub/backstage-plugin-orchestrator': minor
+'@red-hat-developer-hub/backstage-plugin-scaffolder-backend-module-orchestrator': patch
 ---
 
-The updated overview of a workflow will have the availability property if the isAvailable prop is false
+Add conditional RBAC policy support for orchestrator workflows using the `IS_ALLOWED_WORKFLOW_ID` rule. Dynamic workflow-specific permissions (`orchestrator.workflow.<workflowId>` and `orchestrator.workflow.use.<workflowId>`) are deprecated and will be removed in the next release.
+
+**BREAKING CHANGE:** Migrate from deprecated dynamic permissions to conditional policies. See `docs/MIGRATION-CONDITIONAL-POLICIES.md`.
```
PatAKnight:
```suggestion
Migrate from deprecated dynamic permissions to conditional policies. See `docs/MIGRATION-CONDITIONAL-POLICIES.md`.
```

I think that it should be fine to drop the `**BREAKING CHANGE**` now that we are deprecating instead of outright removing.

workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/WorkflowsTable.tsx
```diff
@@ -64,47 +58,21 @@ export interface WorkflowsTableProps {
...
-    ),
-  );
+  // With RBAC conditional policies, the backend filters workflows by permission
+  // If a workflow is returned in the list, the user has permission to view it
+  // For "use" permission, we optimistically allow - the backend will deny if not permitted
   return {
-    allowed: items.map((_, idx) => generic.allowed || specific.allowed[idx]),
+    allowed: items.map(() => true),
   };
 };
```
PatAKnight: Why the removal of `usePermission` here? With the removal across the board, I would think that this can lead to some poor UI/UX. Because even if a user can read the workflow, they might not be able to use the workflow. The RunButton would gate this but without the `usePermission` hooks, users are still able to attempt it even if they can't access it and nothing will happen.

workspaces/orchestrator/package.json
```diff
@@ -48,11 +48,13 @@
     "@jest/environment-jsdom-abstract": "^30.3.0",
     "@types/jest": "^30.0.0",
     "@types/jsdom": "^27.0.0",
+    "@types/supertest": "^7.2.0",
     "jest": "^30.3.0",
     "jsdom": "^27.1.0",
     "knip": "^5.40.0",
     "node-gyp": "^9.0.0",
     "prettier": "3.8.4",
+    "supertest": "^7.2.2",
```
PatAKnight: Do we need to include `supertest` here if we already have it in `orchestrator-backend/package.json` as a devDependency?

workspaces/orchestrator/plugins/orchestrator-backend/src/service/permission-rules.ts
```diff
@@ -0,0 +1,103 @@
...
+  createPermissionRule,
+} from '@backstage/plugin-permission-node';
+
+import { z } from 'zod/v3';
+
+import { ORCHESTRATOR_WORKFLOW_RESOURCE_TYPE } from '@red-hat-developer-hub/backstage-plugin-orchestrator-common';
+
+import type { OrchestratorService } from './OrchestratorService';
+
+let orchestratorService: OrchestratorService | undefined;
```
PatAKnight: I don't know how I feel about this global bind. 

I wonder why something like `initPublicServices` couldn't have been added to `plugin.ts`? Then we could have had something like:

```typescript
permissionsRegistry.addResourceType({
  resourceRef: orchestratorWorkflowResourceRef,
  getResources: refs =>
    Promise.all(
      refs.map(ref =>
        publicServices.orchestratorService.fetchWorkflowOverview({
          definitionId: ref,
        }),
      ),
    ),
  permissions: orchestratorPermissions,
  rules: orchestratorPermissionRules,
});
```

To handle the `getResources` better, mirroring what other plugins do today. WDYT?

workspaces/orchestrator/docs/Permissions.md
```diff
@@ -1,67 +1,222 @@
...
+
+### Option 2: Create Conditional Policy with help of RBAC UI
+
+Create conditional policies via RBAC UI:
+
+1. Navigate to your Backstage RBAC Administration page
+2. Select the role you want to configure (e.g., `role:default/developer`) from the list
+3. Click role name link
+4. Take a look permissions table and click "Edit" button
+5. Expand list plugins and select "Orchestator" plugin
```
PatAKnight:
```suggestion
5. Expand list plugins and select "Orchestrator" plugin
```

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
AndrienkoAleksandr: Good point on the legacy fallback performance.

If migration to conditional policies is complete and the policies cover the same workflows as the old dynamic permissions, the fallback should not run. Flow is: one authorizeConditional call, then local matches() against IS_ALLOWED_WORKFLOW_ID. Workflows that pass go into conditionallyAllowed; remainingIds stays empty, so the legacy block is skipped.

The slow path only happens when there are workflows in remainingIds — e.g. partial migration, conditional policy doesn’t cover all needed workflows, or the user still relies on deprecated orchestrator.workflow.<id> entries in CSV. Then we do N per-workflow legacyAuthorize() calls for those remaining IDs.

After a full migration with correct conditional policies, you shouldn’t hit that path. If you do (e.g. misconfigured policy), you’d still pay for N deny checks even when deprecated permissions are already removed — wasted round-trips, not extra access.

I think the only way to force-disable fallback early would be a config flag (e.g. orchestrator.permissions.disableLegacyFallback). I’m not keen on that: it adds surface area for something we plan to remove in the next release anyway when we drop deprecated dynamic permissions and the legacy blocks in router.ts / the UI hooks.

Happy to document this in the migration guide if that helps — let me know.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/permission-rules.ts
AndrienkoAleksandr: Ok, I handled that, but to achieve that I changed more code than I wanted to.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
AndrienkoAleksandr: Done, I reverted batching.

workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/WorkflowsTable.tsx
AndrienkoAleksandr: I turned back permission logic.

AndrienkoAleksandr:
@PatAKnight, I hope I have handled your code review feedback.

workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts
PatAKnight: Yeah, I believe that I am fine with just documenting this in the migration guide. I don't really want to introduce a new config value since it will be removed in the next release. And I don't believe that there is an easy way to fix this problem at all until we remove the dynamic permissions.

AndrienkoAleksandr: Done, I updated migration guide.

lholmquist:
@AndrienkoAleksandr looks like there are some merge conflicts now.  But overall i think this is looking good and the new functionality works using the testing information provided

AndrienkoAleksandr:
@lholmquist, thanks, I rebased pr.

lholmquist (commented):
@AndrienkoAleksandr everything worked for me when i tested using what you added in the description.    

LGTM

Can we get @PatAKnight to give an approval also, since he is a little more familiar with the RBAC specifics?

workspaces/orchestrator/.changeset/fair-colts-deny.md
```diff
@@ -0,0 +1,10 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-backend': minor
+'@red-hat-developer-hub/backstage-plugin-orchestrator-common': minor
+'@red-hat-developer-hub/backstage-plugin-orchestrator': minor
+'@red-hat-developer-hub/backstage-plugin-scaffolder-backend-module-orchestrator': patch
```
PatAKnight:
```suggestion
```

workspaces/orchestrator/docs/MIGRATION-CONDITIONAL-POLICIES.md
```diff
@@ -0,0 +1,481 @@
...
+
+### Performance note on legacy fallback
+
+During this deprecation release, the backend checks conditional policies first. If the conditional policy covers the requested workflow(s), access is resolved immediately — no legacy fallback runs.
+
+Conditional policies are faster than deprecated dynamic permissions. With conditional policies, the backend makes a single `authorizeConditional` call and then filters workflows locally using `matches()` — regardless of how many workflows exist. With deprecated dynamic permissions, the backend makes a separate `authorize` call per workflow, so latency grows linearly with the number of workflows.
+
+The legacy fallback only runs for workflows **not covered** by conditional policies. For each uncovered workflow, an individual `authorize` call is made to check the deprecated dynamic permission. This means:
+
+- **After a complete migration** (conditional policies cover all workflows your users need): the legacy fallback is never triggered. Performance is the same as if dynamic permissions were already removed.
```
PatAKnight: I don't think this is always true. Migration improves the happy path, but legacy can still run when the conditional allow-list doesn't cover the full catalog i.e. unmatched workflow IDs still go through the legacy fallback.

workspaces/orchestrator/plugins/orchestrator-form-widgets/package.json
```diff
@@ -94,7 +94,7 @@
   },
   "devDependencies": {
     "@backstage/cli": "^0.36.3",
-    "@backstage/core-app-api": "^1.20.2",
+    "@backstage/core-app-api": "^1.20.3",
```
PatAKnight, line 97: I could be wrong, but I think you might need to include the `orchestrator-form-widgets` in the changeset now that a dependency was updated.

PatAKnight (commented):
Everything is looking pretty good. Just one last minor thing and then this should be good to go!

PatAKnight (approved):
LGTM

PR #2970, opened 2026-04-29
[RHIDP-13060] harden lightspeed proxy against arbitrary routes

Jdubrick:

- Adds an explicit allowlist for routes to restrict authenticated users from hitting arbitrary routes on the LCORE side 
- Moved the pre-existing passthrough routes array to constants

Issue: https://redhat.atlassian.net/browse/RHIDP-13060

### DIFF
--- a/workspaces/lightspeed/.changeset/brown-rivers-build.md
+++ b/workspaces/lightspeed/.changeset/brown-rivers-build.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed-backend': patch
+---
+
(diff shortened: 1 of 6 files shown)


Jdubrick:
For reviewers in response to Qodo's review:

1. I confirmed LCORE does not normalize, so the encoded cases will return a 404 if they make it to LCORE 
2. `/v1/feedback` is in both constant arrays as there is the exact match `/v1/feedback` and the partial match `/v1/feedback/status` where one has its own handler and one does not

karthikjeeyar (approved):
Changes looks good!

Image


/approve
/lgtm

PR #3002, opened 2026-05-05
refactor: MigrationPhase as class with helper methods

eloycoto:
It helps with having all in the same place, and using always the same class instead of multiple strings in a lot of places.


elai-shalev (approved):
makes perfect sense

PR #3004, opened 2026-05-05
fix(bulk-import): correct dataFetcher return type and replace unsafe Response casts

jonkoops:
The `dataFetcher` method returns a raw `Response` on error but the type declared only `OrgAndRepoResponse`. Fixed the return type union and replaced all unsafe `as Response` casts with `instanceof` narrowing. Also fixed three no-op test assertions that never validated anything.


christoph-jerolimov (approved):
Lgtm!

PR #3046, opened 2026-05-05
refactor: add JobStatus value object replacing raw string comparisons

eloycoto:
Replace a few raw string status checks across x2a-common, x2a-node, x2a-backend, and x2a frontend with a flyweight JobStatus class following the Phase pattern from PR #3002.

- Add JobStatus class with predicates: isActive(), isFinished(), isPending(), isRunning(), isSuccess(), isError(), isCancelled()
- Replace ['pending','running'].includes() with isActive()
- Replace 3-way || chains with isFinished()
- Keep raw strings for DB writes and API responses (data, not logic)


elai-shalev (approved):
looks like a good improvement

PR #3050, opened 2026-05-05
fix: missing upgrade to BS `1.49.4` of the scorecards `sonarqube` plugin

davidfestal:
Fix the missing upgrade to BS `1.49.4` of the scorecards `sonarqube` plugin


hopehadfield (approved):
Assuming CI is green

PR #3070, opened 2026-05-07
fix(lightspeed): add return statement to address unhandled error in /v1/feedback 

Jdubrick:

Adds a `return` statement to the `/v1/feedback` handler when it gets a 500 error, it was the only one missing this out of all the route handlers.

Issue: https://redhat.atlassian.net/browse/RHIDP-13063

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
@@ -529,6 +529,8 @@ export async function createRouter(
         response.status(500).json({
           error: errormsg,
         });
+
+        return;
       }
 
       const data = await fetchResponse.json();
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
@@ -525,8 +525,7 @@ export async function createRouter(
         const errormsg = `Error from lightspeed-core server: ${errorBody.error?.message || errorBody?.detail?.cause || 'Unknown error'}`;
         logger.error(errormsg);
 
-        // Return a 500 status for any upstream error
-        response.status(500).json({
+        response.status(fetchResponse.status).json({
           error: errormsg,
         });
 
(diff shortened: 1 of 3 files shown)


workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -529,6 +529,8 @@ export async function createRouter(
         response.status(500).json({
           error: errormsg,
         });
+
+        return;
```
gabemontero, line 532: you didn't touch setting of the status (to 500) above, nor the setting of the status to `fetchResponse.status` below, but these 2 bits make me ask

- is fetchResponse.ok false but fetchResponse.status a non error return code ?
- if fetchResponse.status is an error return code, **why** do you want to return 500 instead of the upstream error code

gabemontero (commented):
one question (answers might lead to an actual ask for change)

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
Jdubrick: I think in terms of any error coming from the Lightspeed Core container we are just treating it as 500. The question could be raised to tune this a bit better in the future. wdyt?

for fetchResponse.ok if it is in the 200s it returns true, so we are catching all errors from that check, we could use fetchResponse.status but then we'd need to check for all 2xx codes to say its good I think, `.ok` is a catch-all for them

gabemontero: If you to convert the various 200s to just 200 I'm ok with that.

But traditionally converting all 300s 400s etc to 500 loses info.

And not sure why we need to wait until the future.  What is the concern for addressing these basics now?

gabemontero: ok @Jdubrick and I hoped on a quick call ... leaving router.ts change the same, but independent of the issue which drove him to fix things by adding the return, line 529 was non-optimal, and we'll fix that too, where instead of returning 500, we'll use fetchResponse.status

he'll then tweak the unit tests to adjust

Jdubrick: @gabemontero I updated it in this commit https://github.com/redhat-developer/rhdh-plugins/pull/3070/changes/b3589db55b8c5a9c496495cff7beba796312ecfd and as part of that fixed the other handlers that were hardcoding `500` as well.

gabemontero: awesome thanks @Jdubrick

gabemontero (approved):
lgtm

PR #3097, opened 2026-05-11
fix(orchestrator): Loki backend module security updates

lholmquist:

This fixes some of the security issues that were identified with the loki backend module so it can be made GA

The biggest issue was using the globalDispatcher to modify tls fetch options,  this PR removes the use of the globalDispatcher.

Also adds the limit config variable to further limit the amount of logs that are being returned.

Much of the validation for queries actually happens on the loki server itself, but using cursor, there is now baseUrl validation options and also the logStreamSelector inputs are sanitized 

addresses concerns in https://redhat.atlassian.net/browse/RHIDP-13478

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -29,6 +29,7 @@ export interface Config {
         /**
          * Auth Token for accessing the loki query url
          */
+        /** @visibility secret */
         token: string;
         /**
          * Set to false if the baseUrl has a self-signed certificate
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -32,6 +32,11 @@ export interface Config {
         /** @visibility secret */
         token: string;
         /**
+         * Limit the number of logs to fetch
+         * defaults to 100
+         */
+        limit?: number;
+        /**
          * Set to false if the baseUrl has a self-signed certificate
          * defaults to true
          */
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -46,8 +46,8 @@ export interface Config {
         /** @visibility secret */
         token: string;
         /**
-         * Limit the number of logs to fetch
-         * defaults to 100
+         * Limit the number of logs to fetch.
+         * Defaults to 100. Must not be negative if set.
          */
         limit?: number;
         /**
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -24,8 +24,22 @@ export interface Config {
         /**
          * Base URL of the Loki service.
          * /loki/api/v1/query_range will be appended to the baseUrl
+         *
+         * Must be an absolute http(s) URL without credentials, query, or fragment.
+         * In production (NODE_ENV=production), https is required unless allowInsecureHttp is true.
          */
         baseUrl: string;
+        /**
+         * Optional allowlist for the hostname parsed from baseUrl.
+         * Entries are matched case-insensitively. If an entry starts with `.`, the hostname must be
+         * that suffix or a subdomain (e.g. `.example.com` allows `loki.example.com`).
+         */
+        allowedHosts?: string[];
+        /**
+         * When true, allows http:// baseUrl when NODE_ENV is production.
+         * Defaults to false; local/dev typically uses NODE_ENV!=production where http is already allowed.
+         */
+        allowInsecureHttp?: boolean;
         /**
          * Auth Token for accessing the loki query url
          */
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -60,9 +60,12 @@ export interface Config {
         // new values will be appened
         logPipelineFilters?: Array<string>;
         logStreamSelectors?: Array<{
-          // label is the selector, something like 'app' or 'service_name', etc...
+          /** Prometheus-style name: [a-zA-Z_][a-zA-Z0-9_]* */
           label: string;
-          // value is the label matching operator, so something like: '=~".+"'
+          /**
+           * Label matcher only, e.g. `="application"` or `=~".+"` (quoted; regex may use backticks).
+           * Validated at startup.
+           */
           value: string;
         }>;
       };
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
@@ -61,12 +61,12 @@ export interface Config {
         logPipelineFilters?: Array<string>;
         logStreamSelectors?: Array<{
           /** Prometheus-style name: [a-zA-Z_][a-zA-Z0-9_]* */
-          label: string;
+          label?: string;
           /**
            * Label matcher only, e.g. `="application"` or `=~".+"` (quoted; regex may use backticks).
            * Validated at startup.
            */
-          value: string;
+          value?: string;
         }>;
       };
     };
--- a/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/LokiProvider.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/LokiProvider.ts
@@ -69,7 +69,11 @@ export class LokiProvider implements WorkflowLogProvider {
       config,
       `${LOKI_CONFIG_PATH}.logPipelineFilters`,
     );
-    this.limit = config.getOptionalNumber('limit') || 100;
+    const limitOpt = config.getOptionalNumber('limit');
+    if (limitOpt !== undefined && limitOpt < 0) {
+      throw new Error(`${LOKI_CONFIG_PATH}.limit must not be negative`);
+    }
+    this.limit = limitOpt ?? 100;
     this.agent = new Agent({
       connect: {
         rejectUnauthorized: this.rejectUnauthorized,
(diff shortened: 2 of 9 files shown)


workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
```diff
@@ -24,12 +24,32 @@ export interface Config {
...
         /**
          * Auth Token for accessing the loki query url
          */
+        /** @visibility secret */
         token: string;
+        /**
+         * Limit the number of logs to fetch
+         * defaults to 100
+         */
+        limit?: number;
```
PatAKnight, line 52: We might want to add some validation to this as it can accept both negative numbers and extremely large numbers.

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+    const ctx = `${configKeyPath}[${i}]`;
+    if (!element) {
+      throw new Error(`${ctx}: entry must not be empty or whitespace-only`);
+    }
+    if (/[\r\n\u2028\u2029]/.test(element)) {
+      throw new Error(`${ctx}: entry must not contain line breaks`);
+    }
+    if (element.includes('}')) {
+      throw new Error(`${ctx}: entry must not contain "}"`);
+    }
```
PatAKnight, line 208: Should we also include `{` as a way to fully close out the braces from being an option?

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
```diff
@@ -40,9 +60,12 @@ export interface Config {
...
         logStreamSelectors?: Array<{
-          // label is the selector, something like 'app' or 'service_name', etc...
+          /** Prometheus-style name: [a-zA-Z_][a-zA-Z0-9_]* */
           label: string;
-          // value is the label matching operator, so something like: '=~".+"'
+          /**
+           * Label matcher only, e.g. `="application"` or `=~".+"` (quoted; regex may use backticks).
+           * Validated at startup.
+           */
           value: string;
```
PatAKnight: It looks like we use `entry.getOptionalString` for these and have a default value in the event that they are missing. Should we also mark these as optional so that it better matches with the changes?

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/LokiProvider.ts
```diff
@@ -159,24 +192,20 @@ export class LokiProvider implements WorkflowLogProvider {
...
-          return {
-            id: val[0],
-            log: val[1],
-          };
-        });
+        .flatMap(entry => entry.values)
+        .map(([id, log]) => ({ id, log }));
     } catch (error) {
       throw new Error(`Problem fetching loki logs: ${error.message}`);
     }
```
PatAKnight: While not something that you added here in this PR, I am wondering if it might be better to try to move to `@backstage/errors` over plain `Errors`. Could do something like the following:

```ts
import { assertError, ServiceUnavailableError } from '@backstage/errors';
// ...
} catch (error: unknown) {
  assertError(error);
  throw new ServiceUnavailableError(
    `Problem fetching Loki logs: ${error.message}`,
    error,
  );
}
```

Would could start with updating the ones in `LokiProvider.ts` and also the newly added plain `Errors`, WDYT?

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+
+/** Prometheus / Loki stream label names: `[a-zA-Z_]\w*` (ASCII `\w` only; no `/u` flag). */
+const LOKI_LABEL_NAME_PATTERN = /^[a-zA-Z_]\w*$/;
+
+/**
+ * Label matcher fragment after the label name: `="..."`, `!=`, regex with `"` or `` ` ``.
+ * Requires properly closed quotes and escapes so the fragment cannot break out of `{...}`.
+ */
+const LOG_STREAM_SELECTOR_VALUE_PATTERN =
+  /^(?:(?:=|!=)"(?:[^"\\\r\n]|\\.)*"|(?:=~|!~)(?:"(?:[^"\\\r\n]|\\.)*"|`[^`\r\n]*`))$/;
```
PatAKnight: Maybe something like the following could fix this SonarCloud warning:

```ts
const LOG_STREAM_SELECTOR_VALUE_PATTERNS: RegExp[] = [
  /^=(?:"(?:[^"\\\r\n]|\\.)*")$/,
  /^!=(?:"(?:[^"\\\r\n]|\\.)*")$/,
  /^=~(?:"(?:[^"\\\r\n]|\\.)*"|`[^`\r\n]*`)$/,
  /^!~(?:"(?:[^"\\\r\n]|\\.)*"|`[^`\r\n]*`)$/,
];
```

and then you could update `assertValidLogStreamSelectorValue` with:

```
function assertValidLogStreamSelectorValue(value: string, context: string): void {
  const ok = LOG_STREAM_SELECTOR_VALUE_PATTERNS.some(pattern => pattern.test(value));
  if (!ok) {
    throw new Error(
      `${context}: value must be a LogQL label matcher (e.g. ="literal", !="...", =~"re", or =~\`re\`) with no raw line breaks outside escapes`,
    );
  }
}
```

I'll be honest though, regex is not my strongest so I used AI for this suggestion in its entirety

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+export interface ValidatedLogStreamSelector {
+  label: string;
+  value: string;
+}
+
+function assertValidLokiLabelName(label: string, context: string): void {
+  if (!LOKI_LABEL_NAME_PATTERN.test(label)) {
+    throw new Error(
+      `${context}: label must match Prometheus label name rules [a-zA-Z_]\\w* (got "${label}")`,
+    );
```
PatAKnight, line 143:
```suggestion
    throw new Error(
      `${context}: label must match Prometheus label name rules ${String.raw`[a-zA-Z_]\w*`} (got "${label}")`,
    );
```

Should be able to do something like this to get rid of this SonarCloud complaint

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/helpers.ts
```diff
@@ -0,0 +1,235 @@
...
+      'Workflow instance id contains characters that are not allowed in Loki line filters',
+    );
+  }
+}
+
+/**
+ * Escapes a string for use inside LogQL double-quoted line filter literals (`|="..."`).
+ */
+export function escapeLogQlDoubleQuotedLineLiteral(fragment: string): string {
+  return fragment.replaceAll('\\', '\\\\').replaceAll('"', '\\"');
```
PatAKnight:
```suggestion
export function escapeLogQlDoubleQuotedLineLiteral(fragment: string): string {
  return fragment
    .replaceAll(String.raw`\\`, String.raw`\\\\`)
    .replaceAll('"', String.raw`\\` + '"');
}
```

Maybe something like this to fix the SonarCloud complaint

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/LokiProvider.ts
```diff
@@ -97,21 +130,19 @@ export class LokiProvider implements WorkflowLogProvider {
...
-        ) => {
-          // something about that last comma
-          streamSelector += `${entry.label || 'openshift_log_type'}${entry.value || '=~".+"'}${index !== arr.length - 1 ? ',' : ''}`;
-        },
-      );
+      streamSelector = this.selectors
+        .map(
+          (entry, index) =>
+            `${entry.label}${entry.value}${
+              index !== this.selectors.length - 1 ? ',' : ''
```
PatAKnight:
```suggestion
              index === this.selectors.length - 1 ? '' : ','
```

Might be able to do something like this to get rid of the SonarCloud complaint.

PatAKnight (commented):
Overall looks pretty good. Some minor stuff that doesn't necessarily block this PR but would be nice to potentially address. 

- I think updating the readme with the new config values could be beneficial.
- Seems like there are a couple of places that were missed as part of the [code coverage](https://app.codecov.io/gh/redhat-developer/rhdh-plugins/pull/3097?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=redhat-developer)
- The review comments that I have included

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/config.d.ts
lholmquist: Yea, good call with this one. I'll add a check that he number is not negative.  

For the upper limit, it looks like that value is set on the loki server itself and defaults to 5000,  so if a user of this plugin sets it to something above that, they will error anyway.

this is what i got from google:

> Yes, Loki has a configurable upper limit for query_range (the number of log lines returned), with a default of 5,000 entries per query. This can be increased by updating the max_entries_limit_per_query parameter in the limits_config section of the Loki configuration, typically up to a practical maximum, though setting it too high (e.g., > 30,000–50,000) can cause 413 error responses or performance

lholmquist: updated with this commit https://github.com/redhat-developer/rhdh-plugins/pull/3097/commits/67c157a6abfe7ca58f9be471a5c042f047d3587b

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/helpers.ts
lholmquist: added here https://github.com/redhat-developer/rhdh-plugins/pull/3097/commits/52a61823268a5df92dc65d5a9cf4d5c75bc5c250

workspaces/orchestrator/plugins/orchestrator-backend-module-loki/src/workflowLogsProviders/LokiProvider.ts
lholmquist: https://github.com/redhat-developer/rhdh-plugins/pull/3097/commits/f641bd3f43d3900112e22a099606b6efb9cf5e19 and https://github.com/redhat-developer/rhdh-plugins/pull/3097/commits/50cb211e27f26a55c66fec3a2e47bd97a10e07ed should cover this

PatAKnight (approved):
LGTM

PR #3107, opened 2026-05-11
feat(translations): add pseudo-localization support

rohitratannagar:
## Summary

  **Resolves : [RHDHBUGS-3029](https://redhat.atlassian.net/browse/RHDHBUGS-3029)**
  
  Add pseudo-localization support to the translations plugin to help identify untranslated or hardcoded strings across the RHDH UI. Strings are transformed with diacritical marks and brackets (e.g., `"Settings"` →
   `"[Ṣḛḛţţḭḭṇḡṡ]"`), making untranslated text visually obvious.

  ### Usage in RHDH
  ```yaml
  dynamicPlugins:
    frontend:
      red-hat-developer-hub.backstage-plugin-translations:
        mountPoints:
          - mountPoint: application/provider
            importName: PseudoLocalizationProvider
```

  ## Activation

  **Option 1:** URL query parameter
  https://your-rhdh-instance.com/?pseudolocalization=true
  
  **Option 2:** App config
  ```yaml
  i18n:
    pseudolocalization:
      enabled: true
```

  ### demo
  
  

https://github.com/user-attachments/assets/e5702652-aac6-4288-a84d-22a6ef9828e8

  No static code changes required in RHDH.

### DIFF
--- a/workspaces/translations/.changeset/sweet-ends-dance.md
+++ b/workspaces/translations/.changeset/sweet-ends-dance.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-translations': minor
+---
+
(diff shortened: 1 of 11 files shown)


debsmita1 (commented):
We should think of migrating the translations plugin to NFS and use  `AppRootWrapperBlueprint` instead of `application/provider` .

christoph-jerolimov (approved):
Thanks a lot @rohitratannagar, this is great!

@debsmita1 is right, we need to double check to enable this for NFS. It's so useful.

But let us merge this old PR now and then we can cont.

Can you prob. upgrade the workspace and check NFS?

PR #3112, opened 2026-05-12
refactor: create CallbackToken class

eloycoto:
Replace SignatureValidator and generateCallbackToken with a self-validating CallbackToken class that encapsulates generation, HMAC signing, and signature verification.


mareklibra (changes requested):
Code-wise this LGTM.

I notice this continues the trend of refactoring toward OOP patterns. While I appreciate its principles in other languages, I'm careful about over-using those in JavaScript/TypeScript. JS is fundamentally functional - its classes are syntactic sugar over prototypal inheritance and the OOP support doesn't behave as one might expect coming from languages like Java (a commonly known issue ). Overusing class-based architecture in JS tends to introduce friction rather than reduce it. That said, I won't block on this, just flagging the direction for the team to be mindful of.

To meet requirements of this repo, the title and changeset need to be tweaked.

Considering no functionality is added, the title should start with `chores(x2a):`.

Please address SonarQube findings, they make sense this time.

eloycoto:
> I notice this continues the trend of refactoring toward OOP patterns. While I appreciate its principles in other languages, I'm careful about over-using those in JavaScript/TypeScript. JS is fundamentally functional - its classes are syntactic sugar over prototypal inheritance and the OOP support doesn't behave as one might expect coming from languages like Java (a commonly known issue ). Overusing class-based architecture in JS tends to introduce friction rather than reduce it. That said, I won't block on this, just flagging the direction for the team to be mindful of.

Hi!

To clarify, the goal isn't to move everything to OOP or introduce inheritance hierarchies. I'm well aware of the pitfalls that brings in JavaScript.

The motivation here is more pragmatic:

- Single Responsibility: Grouping related data and behavior together so each piece has a clear owner.
- Encapsulation: Replacing string comparisons scattered across multiple files with a well-defined type that's harder to misuse.
- Discoverability: Making it easier to find and understand what operations are available on a given concept.

It's less about "classes for the sake of classes" and more about reducing the surface area for bugs, especially now that we have a better understanding of the domain scope.

mareklibra:
Thanks for clarification, agreed. I wanted to be sure we are on the same page with using the OOP in JS.

Please have a look at the chores raised above and lets merge this.

mareklibra (approved):

PR #3130, opened 2026-05-13
feat(x2a): adding 'Edit project' action

mareklibra:
Fixes: FLPATH-3463

Adds the ability to edit a project's name, owner (createdBy), and description from the project details page.

**Backend**

- New PATCH /projects/:projectId endpoint with Zod-validated body (name, createdBy, description -- all optional, at least one required)
- updateProject database operation with ownership-aware permission enforcement (admin can update any project; regular users only their own)
- OpenAPI schema updated with the new PATCH operation and generated client/server types
- ENTITY_REF_RE regex added to x2a-common for shared entity-ref validation (user:… / group:…)
- dirName field exposed on the Project model; assertProjectHasDirName guard added before init runs
- DB migration 2026051200_add_project_dir_name
- The directory in the target repo can not be changed (especially not by changing the project's name)

**Frontend**
- EditProjectDialog component with an Autocomplete-based owner picker:
- Regular users see a dropdown populated from their ownershipEntityRefs (own ref + groups)
- Admins get freeSolo mode -- select from suggestions or type any valid entity ref with live format validation
- ProjectDetailsCard gains an edit button and wires the dialog
- Ownership transfer confirmation dialog warns users they may lose access
- Translations added for all new strings (EN, DE, ES, FR, IT)

- [ x Screenshots attached (for UI changes)

---

https://github.com/user-attachments/assets/b26c0563-a896-434a-b2d4-4cd91ec8e0d5


eloycoto (approved):

workspaces/x2a/plugins/x2a-common/src/constants.ts
```diff
@@ -101,3 +101,14 @@ export const RUN_INIT_DEEP_LINK_HASH = '#runinit';
...
+/**
+ * Pattern for a valid Backstage entity reference, e.g.
+ * `user:default/my-name` or `group:default/team-a`.
+ *
+ * Namespace and name segments are restricted to lowercase alphanumerics,
+ * dashes, underscores, and dots to match the Backstage catalog conventions.
+ *
+ * @public
+ */
+export const ENTITY_REF_RE = /^(user|group):[a-z0-9_.-]+\/[a-z0-9_.-]+$/;
```
elai-shalev, line 114: This is blocking upper case, but backstage technically is case sensative and does allow upper cases.

workspaces/x2a/plugins/x2a-backend/src/schema/openapi.yaml
```diff
@@ -150,6 +150,40 @@ paths:
...
+                  description: Description of the project
+      responses:
+        '200':
+          description: Updated project data.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Project'
+        '404':
+          description: Project not found.
```
elai-shalev, line 188: I think 400 error is also needed, for the  InputError option.

workspaces/x2a/plugins/x2a/src/translations/ref.ts
```diff
@@ -54,10 +54,24 @@ export const x2aPluginMessages = {
     name: 'Name',
     abbreviation: 'Abbreviation',
     status: 'Status',
-    createdBy: 'Created By',
+    createdBy: 'Owned By',
```
elai-shalev: Why has this changes to "Owned by"? Would it not make sense to add an additional "Owned by" but keep the old created by info?

elai-shalev (commented):
overall looks good, some comments. We still need to think about the dirname change, wither to implement it (not sure if we want that) or make sure to leave a "breadcrumbs" for info about the history of the project name (change log?) or link to dir name. Info can be lost when these projects are migrated from one user to another

workspaces/x2a/plugins/x2a-common/src/constants.ts
mareklibra: I think it is compared in a case-sensitive way but there is a requirement for keeping the refs as lowercase:
- https://backstage.io/api/stable/functions/_backstage_catalog-model.index.stringifyEntityRef.html
- https://backstage.io/docs/features/software-catalog/references/#string-references

So easing the rule here could cause more troubles.

workspaces/x2a/plugins/x2a/src/translations/ref.ts
mareklibra: Technically we can track both but is that needed?

IMO, keeping just a single field for that is enough for the use-cases considered so far. My preference is in keeping the UI simple, not overwhelming. But if you think otherwise, let's add it.

Anyway, your comment makes me think that we should rename the Created By to the Owned By completely, otherwise it can be confusing. I will add a commit unifying it.

elai-shalev: Lets just change them all to Owned by, and have only one field then

mareklibra:
> We still need to think about the dirname change

So far we have no hints that this will be needed, so let's implement the dir_name change later, manipulation with target repo will need to be included.

In a follow-up (about to push it), the `abbreviation` detail's item will be replaced by the `dir name`, it will be a click-able link to the target repo. Do you see such solution as sufficient?

elai-shalev (approved):

elai-shalev:
> > We still need to think about the dirname change
> 
> So far we have no hints that this will be needed, so let's implement the dir_name change later, manipulation with target repo will need to be included.
> 
> In a follow-up (about to push it), the `abbreviation` detail's item will be replaced by the `dir name`, it will be a click-able link to the target repo. Do you see such solution as sufficient?

that's sufficient yes, sounds good. lgtm

PR #3133, opened 2026-05-13
 fix(lightspeed): replace JSS makeStyles with emotion-based implement…

rohitratannagar:
## Summary
                                                                                                                                                                                              
 **Fixes: https://redhat.atlassian.net/browse/RHDHBUGS-3133**  
The lightspeed plugin uses JSS `makeStyles` which generates class names based on a global counter (e.g., `jss-1`). When multiple plugins load on the same page, their counters collide, causing styles to bleed across plugins.

  ### What changed                                                                                                                                                                            

  - **New `utils/makeStyles.ts`** — a custom `makeStyles` implementation built on top of emotion's `sx` engine (already bundled via MUI). It mimics the same API as MUI's `makeStyles` — identical call signature, returns `classes` with real CSS class names — so existing component code stays untouched. Class names are content-hashed (`ls-abc123`), so collisions are impossible regardless of plugin load order.
  - **Migrated all 23 component files** — only import lines changed. Zero changes to style definitions or `className` usage.
  - **Fixed `theme.palette.type` → `theme.palette.mode`** in `FileListItem.tsx` (MUI v4 → v5 API).

  ### Why not the seed approach?

  1. Lightspeed has multiple independent mount points (`LightspeedPage`, `LightspeedFAB`, `LightspeedDrawerProvider`) — each would need to be wrapped in its own `StylesProvider`, and their counters would still collide with each other.
  2. `makeStyles` from `@material-ui/core` and `@mui/styles` is already deprecated by MUI. Building on top of emotion aligns with MUI's recommended direction.

  Emotion avoids all of this — no provider, no counter, no wrapping needed.

### DIFF
--- a/workspaces/lightspeed/.changeset/dull-swans-do.md
+++ b/workspaces/lightspeed/.changeset/dull-swans-do.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed': patch
+---
+
+Replace JSS-based makeStyles with emotion-based implementation to fix CSS class name collisions when lightspeed plugin is loaded alongside other plugins. Added MUI ClassNameGenerator with `ls-` prefix for
--- a/workspaces/lightspeed/.changeset/dull-swans-do.md
+++ b/workspaces/lightspeed/.changeset/dull-swans-do.md
@@ -2,4 +2,4 @@
 '@red-hat-developer-hub/backstage-plugin-lightspeed': patch
 ---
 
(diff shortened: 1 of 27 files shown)


its-mitesh-kumar:
Closing this PR in favor of https://github.com/redhat-developer/rhdh-plugins/pull/3152/changes/2dee4c50bdb2e36f04acce27454209f65449f801 commit of this https://github.com/redhat-developer/rhdh-plugins/pull/3152 PR.

PR #3154, opened 2026-05-15
feat(orchestrator-form): evaluate conditional ui:hidden at step level

skylerto:
## What

Evaluate conditional `ui:hidden` expressions during wizard step filtering and add two new condition operators (`notContains`, `isNotEmptyList`) to `HiddenConditionObject`.

## Why

`getSortedStepEntries` only checks for static `ui:hidden: true` when filtering wizard steps. Conditional `ui:hidden` expressions (`anyOf`, `allOf`, `when`/`is`/`isNot`/`isEmpty`) are only evaluated at the individual field level by `FieldTemplate`, not during step filtering.

This means a step whose `ui:hidden` is a conditional object (or whose properties all have conditional `ui:hidden`) is never filtered out of the wizard stepper, resulting in **empty pages being shown**.

Additionally, `evaluateConditionObject` uses early-return logic which only evaluates the first matching operator in a condition object. This prevents combining multiple operators (e.g., `isNotEmptyList` + `notContains`) in a single condition with AND semantics.

## Changes

- **`HiddenCondition.ts`** — Add `notContains` and `isNotEmptyList` fields to `HiddenConditionObject`
- **`evaluateHiddenCondition.ts`** — Refactor `evaluateConditionObject` to AND all specified operators instead of early-returning on the first match; add evaluation for `isNotEmptyList` and `notContains`
- **`getSortedStepEntries.ts`** — Add optional `formData` parameter to `getSortedStepEntries` and `getActiveStepKey`; use `evaluateHiddenCondition` to evaluate conditional `ui:hidden` objects at both the step level and the "all properties hidden" level
- **`StepperObjectField.tsx`** — Pass `formData` to `getSortedStepEntries` so step filtering is reactive to form data changes
- **`OrchestratorFormWrapper.tsx`** — Pass `formData` to `getActiveStepKey`

## How to test

1. Create a workflow schema with a step that has a conditional `ui:hidden` (e.g., `{ when: "someField", is: "someValue" }`) at the step level
2. Verify the step is hidden/shown dynamically as the referenced field value changes
3. Create a step where all properties have conditional `ui:hidden` — verify the step disappears when all properties evaluate to hidden
4. Test the new operators:
   - `isNotEmptyList: true` hides when the field is a non-empty array
   - `notContains: "x"` hides when the array does not contain `"x"`
   - Combine both for "non-empty list that doesn't contain X" patterns

### DIFF
--- a/workspaces/orchestrator/.changeset/conditional-step-hiding.md
+++ b/workspaces/orchestrator/.changeset/conditional-step-hiding.md
@@ -1,0 +1,16 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator-form-react': minor
+---
+
+Evaluate conditional `ui:hidden` at the wizard step level and add `notContains`/`isNotEmptyList` operators.
+
+Previously, `getSortedStepEntries` only checked for static `ui:hidden: true` when filtering wizard steps. Conditional `ui:hidden` expressions (`anyOf`, `allOf`, `when`/`is`/`isNot`/`isEmpty`) were only evaluated at the individual field level, causing steps with conditional hidden logic to render as empty pages.
+
+Additionally, `evaluateConditionObject` used early-return logic that only evaluated the first matching operator, preventing multiple operators from being combined with AND semantics.
+
+This change:
+
+- Adds `notContains` and `isNotEmptyList` operators to `HiddenConditionObject`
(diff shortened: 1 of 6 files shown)


skylerto:
closing for PR against main: https://github.com/redhat-developer/rhdh-plugins/pull/3191

PR #3160, opened 2026-05-18
chore(x2a): Create ArtifactKind class

eloycoto:
Create a new class for ArtifactKind to make it simple to use it in all the places.

### DIFF
--- a/workspaces/x2a/plugins/x2a-common/src/domain/ArtifactKind.ts
+++ b/workspaces/x2a/plugins/x2a-common/src/domain/ArtifactKind.ts
@@ -27,14 +27,16 @@ export class ArtifactKind {
   static readonly PROJECT_METADATA = new ArtifactKind('project_metadata');
   static readonly ANSIBLE_PROJECT = new ArtifactKind('ansible_project');
 
+  private static readonly ALL = Object.freeze([
+    ArtifactKind.MIGRATION_PLAN,
+    ArtifactKind.MODULE_MIGRATION_PLAN,
+    ArtifactKind.MIGRATED_SOURCES,
+    ArtifactKind.PROJECT_METADATA,
+    ArtifactKind.ANSIBLE_PROJECT,
+  ]);
+
   private static readonly BY_VALUE = new Map<string, ArtifactKind>(
-    [
-      ArtifactKind.MIGRATION_PLAN,
-      ArtifactKind.MODULE_MIGRATION_PLAN,
-      ArtifactKind.MIGRATED_SOURCES,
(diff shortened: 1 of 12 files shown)


workspaces/x2a/plugins/x2a-common/src/domain/ArtifactKind.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (!kind) {
+      throw new Error(
+        `Invalid artifact type: "${raw}". Valid: ${ArtifactKind.values().join(', ')}`,
+      );
+    }
+    return kind;
+  }
+
+  static all(): readonly ArtifactKind[] {
+    return [
```
mareklibra: Nitpick: to simplify maintenance, we can declare `private static ALL = Object.freeze([...])`, simply return it here plus reuse it for the BY_VALUE. By doing that, maintenance will be simpler and returned arrays always the same (equal by reference)

mareklibra (changes requested):
There are still raw-string comparisons left in the code.
When we have decided to wrap it by classes, they should be refactored.

eloycoto:
@mareklibra comments has been fixed.

workspaces/x2a/plugins/x2a/src/components/ModulePage/ArtifactsCard.tsx
```diff
@@ -38,13 +39,13 @@ export const ArtifactsCard = ({
 }) => {
   const { t } = useTranslation();
   const moduleMigrationPlanArtifact = module?.analyze?.artifacts?.find(
-    artifact => artifact.type === 'module_migration_plan',
+    artifact => artifact.type === ArtifactKind.MODULE_MIGRATION_PLAN.toString(),
```
mareklibra: This goes against the reasoning behind having the ArtifactKind wrapped into a value object. To follow the idea behind, the `is*` predicates should be used instead.

mareklibra (commented):
Leaving one more comment.

IMHO, the former approach with proper TypeScript guarding is a better approach in this case, reducing one level of abstraction.

Code-wise, there is nothing wrong, so approving.

mareklibra (approved):

mareklibra (changes requested):
Taking back the approval.

Please have a look at `humanizeArtifactType` as well

eloycoto:
@mareklibra hope it's ok now.

eloycoto:
@mareklibra can this be merged?

mareklibra (approved):

PR #3178, opened 2026-05-19
fix(lightspeed): convert FAB from makeStyles/StylesProvider to sx prop to prevent global style leak

its-mitesh-kumar:
## Fixed 
- JIra : https://redhat.atlassian.net/browse/RHDHBUGS-3164
## Summary
- Replaced `makeStyles` + `StylesProvider` in both `LightspeedFAB` components (`src/components/` and `src/alpha/`) with MUI's `sx` prop
- Removed the seeded class name generator (`generateClassName`) from  LightSpeedDrawerProvider

## Problem
In RHDH cluster deployments, the `lightspeedFAB` module is registered as an `AppRootWrapperBlueprint` which wraps the **entire application** inside `LightspeedProvider` (i.e. `LightspeedDrawerProvider`). The `StylesProvider` with a `lightspeed` seed causes ALL child components' JSS classes to be prefixed with `lightspeed-`, breaking styles across every page.
This was not reproducible in local dev because JSS instance sharing differs between local and cluster environments.
## Fix
Convert the FAB's styles to MUI `sx` prop (emotion-based), eliminating the JSS dependency and the need for `StylesProvider` entirely. The FAB no longer injects a custom class name generator into the component tree.

### DIFF
--- a/workspaces/lightspeed/.changeset/fix-lightspeed-jss-style-leak.md
+++ b/workspaces/lightspeed/.changeset/fix-lightspeed-jss-style-leak.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed': patch
+---
+
+fix(lightspeed): prevent JSS class name prefix from leaking into host application styles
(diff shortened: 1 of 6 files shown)


karthikjeeyar (approved):
verified on rhdh-local


Image

Image

PR #3181, opened 2026-05-19
feat(x2a): Copy all files created

eloycoto:
When the export agent creates new files because of RULES, it will create files based on the rules, but was not copied on the phase:

Example:
```
-> tail -n 5 ../chef-examples/EXPORT-AGENTS.md
---

Every migrated Ansible project must include a GitHub Actions workflow at `.github/workflows/ansible-ci.yml` that runs `ansible-lint` on every pull request. The workflow runs on `ubuntu-latest`, checks out the code, installs `ansible-lint`, and executes `ansible-lint roles/ --strict`.%
$ -> ls ../chef-examples/.github/workflows
ansible-ci.yml
```

This change just go over the git files, and merged the files that are created, outside the ansible folder, and do all the files directly

Related to: FLPATH-3351

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
+++ b/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
@@ -129,7 +129,7 @@ copy_changed_files() {
 
   echo "=== Copying changed files to ${dest_dir}/ ==="
 
-  cd "${source_dir}"
+  pushd "${source_dir}" > /dev/null
 
   # Collect new (untracked) and modified files into a single list
   local changed_files=()
@@ -145,6 +145,7 @@ copy_changed_files() {
 
   if [[ ${#changed_files[@]} -eq 0 ]]; then
     echo "  No changed files detected"
+    popd > /dev/null
     return 0
   fi
 
@@ -169,6 +170,8 @@ copy_changed_files() {
   done
 
   echo "  Copied ${copied} file(s)"
+
+  popd > /dev/null
 }
 
 # Authenticated git wrappers using credential helper.
--- a/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
+++ b/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
@@ -109,6 +109,68 @@ run_x2a() {
   ERROR_MESSAGE=""
 }
 
+# Copy all new/modified files from the source directory to a destination.
+# Uses git to detect what the x2a tool created or changed, so we don't
+# hardcode file patterns and never miss output (e.g. .github/workflows/).
+#
+# Usage: copy_changed_files <source_dir> <dest_dir> [exclude_pattern ...]
+#   source_dir       — git repo where x2a wrote its output
+#   dest_dir         — where to copy the files
+#   exclude_pattern  — optional grep -E patterns for files to skip
+#                      (files we copied INTO source before running x2a)
+copy_changed_files() {
+  local source_dir="$1"
+  local dest_dir="$2"
+  shift 2
+  local exclude_patterns=("$@")
+
+  # Always exclude internal telemetry file
+  exclude_patterns+=('\.x2a-telemetry\.json$')
+
+  echo "=== Copying changed files to ${dest_dir}/ ==="
+
+  cd "${source_dir}"
+
+  # Collect new (untracked) and modified files into a single list
+  local changed_files=()
+  while IFS= read -r file; do
+    [[ -z "$file" ]] && continue
+    changed_files+=("$file")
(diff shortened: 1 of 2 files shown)


workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -300,15 +362,9 @@ case "${PHASE}" in
...
-    echo "Copying output to ${PROJECT_PATH}/"
-    cp -v "${SOURCE_BASE}/migration-plan.md" "${PROJECT_PATH}/"
-    # Copy any other generated files (like metadata, agent files)
-    cp -v "${SOURCE_BASE}"/*.json "${PROJECT_PATH}/" 2>/dev/null || true
-    cp -v "${SOURCE_BASE}"/*.yaml "${PROJECT_PATH}/" 2>/dev/null || true
-    cp -v "${SOURCE_BASE}/INPUT-AGENTS.md" "${PROJECT_PATH}/" 2>/dev/null || true
-    cp -v "${SOURCE_BASE}/EXPORT-AGENTS.md" "${PROJECT_PATH}/" 2>/dev/null || true
+    # Copy all files the x2a tool created/modified to the target project directory
+    # Exclude accepted rules we copied in before running x2a
+    copy_changed_files "${SOURCE_BASE}" "${PROJECT_PATH}" '^x2a-rules/'
```
elai-shalev, line 370: so we exclude both this and everything under the excluded patterns?

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -366,16 +422,10 @@ case "${PHASE}" in
...
-    cp -v "${SOURCE_BASE}/migration-plan-${MODULE_NAME_SANITIZED}.md" "${OUTPUT_DIR}/"
-    cp -v "${SOURCE_BASE}"/*.yaml "${OUTPUT_DIR}/" 2>/dev/null || true
-    cp -rv "${SOURCE_BASE}/migration-dependencies" "${OUTPUT_DIR}/" 2>/dev/null || true
-
-    # Update project-level Policyfile.lock.json - the source tool (like chef-cli) may have updated it
-    # during dependency resolution. Keep it at project root only, not per-module.
-    cp -v "${SOURCE_BASE}/Policyfile.lock.json" "${PROJECT_PATH}/" 2>/dev/null || true
+    # Copy all files the x2a tool created/modified to the module output directory
+    # Exclude files we copied into source before running x2a
+    copy_changed_files "${SOURCE_BASE}" "${OUTPUT_DIR}" \
```
mareklibra, line 430: Does that properly handle the `Policyfile.lock.json`? The OUTPUT_DIR is a module path, not the project root as it used to be.

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -109,6 +109,68 @@ run_x2a() {
...
+  local dest_dir="$2"
+  shift 2
+  local exclude_patterns=("$@")
+
+  # Always exclude internal telemetry file
+  exclude_patterns+=('\.x2a-telemetry\.json$')
+
+  echo "=== Copying changed files to ${dest_dir}/ ==="
+
+  cd "${source_dir}"
```
mareklibra: Missing revert - sort of pushd/popd clean-up would be safer for the future.

So far absolute paths are used, however leaving the working directory unchanged by the copy_changed_files would be a better practice.

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
mareklibra: Could INPUT-AGENTS.md be leaked into the module's directory?

mareklibra (commented):
Original idea behind having all the filenames hardcoded was in having control over the target content to include what is really needed, having the `artifacts`.

This patch changes the logic - we will need to explicitly exclude whatever is meant to be temporal.

I agree it makes more sense when the input agents evolved from the single `Chef`.

It's not an issue, just noting.

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
eloycoto: That files are to be copy, always

mareklibra (approved):

mareklibra (approved):

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
eloycoto: yup!

elai-shalev (approved):

PR #3196, opened 2026-05-21
feat(x2a): add action to resync project migration

mareklibra:
Fixes: FLPATH-3214

The user can newly update the project migration plan in an external flow and then let the X2A resync the module list changes.

---

https://github.com/user-attachments/assets/d0b3b4a6-34e5-4ab0-956d-74572932c606


workspaces/x2a/plugins/x2a-backend/src/router/phaseActions.ts
```diff
@@ -63,36 +63,102 @@ class InitPhaseAction implements PhaseAction {
...
+    // Modules in metadata but not in DB (or previously soft-deleted): add or restore
+    const toAdd: MetadataModule[] = [];
+    const toRestore: typeof existingModules = [];
+    const toUpdate: Array<{
+      id: string;
+      sourcePath?: string;
+      technology?: ReturnType<typeof normalizeSourceTechnology>;
+    }> = [];
+
+    for (const mm of metadataModules) {
```
eloycoto, line 121: This fucntion is almost impossible to follow, should we have an abstraction here? It's 100 lines more of code, and It does so many things, that I think that It should be break to avoid scenarios where this is going to be difficult to debug.

mareklibra: simplified

eloycoto (approved):

PR #3207, opened 2026-05-21
fix(augment): enforce lifecycle validation on publish/unpublish routes

rrbanda:
## Summary

- `PUT /agents/:id/publish` now enforces `staging → production` transition via `isValidTransition`
- `PUT /agents/:id/unpublish` now enforces `production → staging` transition
- `PUT /agents/bulk-publish` validates each agent individually and reports skipped agents that fail validation
- All routes now include `from`/`to` stages in audit log entries

Previously, these shortcuts allowed jumping from any lifecycle stage directly to production, bypassing required review and staging phases.

Part of the Agent Lifecycle Governance Epic.

## Test plan

- [ ] Try to publish an agent in "draft" stage -- should get validation error
- [ ] Promote an agent through draft → review → staging, then publish -- should succeed
- [ ] Unpublish a production agent -- should move to staging
- [ ] Try to unpublish a non-production agent -- should get validation error
- [ ] Bulk publish mixed agents (some in staging, some in draft) -- draft ones should be skipped

### DIFF
--- a/workspaces/augment/plugins/augment-backend/src/routes/agentRoutes.ts
+++ b/workspaces/augment/plugins/augment-backend/src/routes/agentRoutes.ts
@@ -431,7 +431,7 @@ export function registerAgentRoutes(
   );
 
   // ---------------------------------------------------------------------------
-  // PUT /agents/:agentId/publish -- shortcut: promote to production
+  // PUT /agents/:agentId/publish -- promote to production (validates transition)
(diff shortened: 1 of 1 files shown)


rrbanda:
Closing: superseded by Phase 1 (#3222) + Phase 4 (#3225) which consolidate lifecycle transition enforcement with the full hardening implementation.

PR #3232, opened 2026-05-24
test(mcp-integrations): add MCP plugin and backend integration tests

HusneShabbir:
## Summary

Adds integration tests for MCP overlay plugins: lightweight plugin wiring checks (action registration) and a backend HTTP suite against `/api/mcp-actions/v1` (`tools/list`, `tools/call`, auth).

Also adds `test:unit` / `test:integration` scripts (`--watch=false` so local runs exit instead of staying in Backstage’s default Jest watch mode) and backend test dependencies for the MCP client.

## Test file structure

```
workspaces/mcp-integrations/
├── plugins/
│   ├── software-catalog-mcp-extras/src/
│   │   └── plugin.integration.test.ts          # 1 test
│   ├── techdocs-mcp-extras/src/
│   │   └── plugin.integration.test.ts          # 1 test
│   └── scaffolder-mcp-extras/src/
│       └── plugin.integration.test.ts          # 1 test
└── packages/backend/src/
    └── mcp-tools.integration.test.ts           # 8 tests
```

**Total: 4 files · 11 integration tests** (included in CI via `yarn test:all`)

---

## Plugin tests — “does the plugin register the right actions?”

Each file starts `startTestBackend`, captures `actionsRegistry.register()`, and asserts action names and count.

| File | Actions |
|------|---------|
| `software-catalog-mcp-extras/.../plugin.integration.test.ts` | `query-catalog-entities` (1) |
| `techdocs-mcp-extras/.../plugin.integration.test.ts` | `fetch-techdocs`, `analyze-techdocs-coverage`, `retrieve-techdocs-content` (3) |
| `scaffolder-mcp-extras/.../plugin.integration.test.ts` | `execute-template`, `fetch-template-metadata`, `get-scaffolder-task-logs`, `list-scaffolder-actions`, `list-scaffolder-tasks`, `validate-scaffolder` (6) |

---

## Backend test — “does the MCP server work over HTTP?”

`packages/backend/src/mcp-tools.integration.test.ts` — Streamable HTTP MCP client against a real test backend (shared backend for full `pluginSources`; MCP session teardown via `terminateSession` + `client.close`).

| Group | Scenario |
|-------|----------|
| **tools/list** (all sources) | 10 overlay tools, names, descriptions, `inputSchema`, read-only/destructive MCP hints (actions that declare `attributes` only) |
| **tools/call** | `query-catalog-entities` → `{ entities: [] }` |
| | `analyze-techdocs-coverage` → zero coverage stats |
| | `fetch-template-metadata` → `{ templates: [] }` |
| **pluginSources** | Catalog-only config → single `query-catalog-entities` tool |
| **auth** | Missing / invalid Bearer → `401` with expected error message |
| | Valid Bearer → `tools/list` succeeds |

---

## Supporting changes

- **`attributes`** on `fetch-techdocs` and `retrieve-techdocs-content` (read-only MCP hints covered by the backend test; scaffolder read-only/destructive tools already had `attributes` on `main`)
- **Backend `devDependencies`:** `@backstage/backend-test-utils`, `@modelcontextprotocol/sdk`, `@backstage/backend-plugin-api`, `@backstage/errors`, `@backstage/plugin-catalog-node`
- **Scripts:** workspace + plugin `test:unit` / `test:integration`; backend `test:integration` for the HTTP suite; `--watch=false` on integration scripts so they exit locally (CI already sets `CI=true`)

## CI

When `workspaces/mcp-integrations` changes, **Test changed packages** runs `yarn test:all`, which includes these integration tests alongside existing unit tests (no separate job required).

## Resolves
https://redhat.atlassian.net/browse/RHIDP-14286


workspaces/mcp-integrations/packages/backend/src/mcp-tools.integration.test.ts
```diff
@@ -0,0 +1,477 @@
...
+        requireMcpToken: true,
+      });
+    });
+
+    it('rejects MCP requests without Authorization when an MCP token is required', async () => {
+      await expect(
+        withMcpClient(backend.server, async client =>
+          client.request({ method: 'tools/list' }, ListToolsResultSchema),
+        ),
+      ).rejects.toThrow();
```
johnmcollier: Should we check for a specific Authorization error here?

workspaces/mcp-integrations/packages/backend/package.json
```diff
@@ -16,7 +16,7 @@
     "start": "backstage-cli package start",
     "build": "backstage-cli package build",
     "lint": "backstage-cli package lint",
-    "test": "backstage-cli package test",
+    "test": "CI=true backstage-cli package test --forceExit --runInBand",
```
johnmcollier: Can we avoid hardcoding CI=true and otherwise modifying the test command? Since `yarn test` is also used for local development scenarios.

You might also want to think about adding a separate command to allow the unit and integration tests to be run separately.

johnmcollier: And why is `--forceExit` needed?

workspaces/mcp-integrations/plugins/techdocs-mcp-extras/src/plugin.integration.test.ts
```diff
@@ -0,0 +1,82 @@
...
+const EXPECTED_ACTIONS = [
+  'analyze-techdocs-coverage',
+  'fetch-techdocs',
+  'retrieve-techdocs-content',
+];
+
+describe('mcpTechdocsExtrasPlugin integration', () => {
+  let registeredActionNames: string[];
+
+  beforeAll(async () => {
```
johnmcollier, line 36: Qodo raised the lack of an `afterAll` / teardown function as why your tests possibly needed `--forceExit`, and I'm inclined to agree. 

If there wasn't another reason why `--forceExit` was needed, could we add teardowns/afterAll functions shutting down the MCP server? Some of `stopTestBackend` function I guess?

workspaces/mcp-integrations/plugins/scaffolder-mcp-extras/src/actions/createFetchTemplateMetadataAction.ts
```diff
@@ -29,6 +29,11 @@ export const createFetchTemplateMetadataAction = ({
   actionsRegistry.register({
     name: 'fetch-template-metadata',
     title: 'Fetch Software Template Metadata',
+    attributes: {
```
johnmcollier: I would separate this out into a separate commit with its own changeset since this affects functionality of the MCP tool/plugin

workspaces/mcp-integrations/plugins/software-catalog-mcp-extras/src/actions/createQueryCatalogEntitiesAction.ts
```diff
@@ -30,6 +30,11 @@ export const createQueryCatalogEntitiesAction = ({
   actionsRegistry.register({
     name: 'query-catalog-entities',
     title: 'Fetch Catalog Entities',
+    attributes: {
```
johnmcollier: Same comment as above

workspaces/mcp-integrations/plugins/techdocs-mcp-extras/src/actions/createAnalyzeTechDocsCoverageAction.ts
```diff
@@ -37,6 +37,11 @@ export const createAnalyzeTechDocsCoverageAction = ({
   actionsRegistry.register({
     name: 'analyze-techdocs-coverage',
     title: 'Analyze TechDocs Coverage',
+    attributes: {
```
johnmcollier: Same comment as above (and for the other MCP tools in this plugin)

johnmcollier (changes requested):
Overall looks pretty good, left some review comments. Also make sure you add changesets for the changed plugins

gabemontero (commented):
I have nothing to add review wise with what is here outside of what @johnmcollier already reported @HusneShabbir 

I will surface a discussion we had in https://redhat-internal.slack.com/archives/C08DZMUKU6R/p1779704191007899 wrt to MCP tools we have managed to transition from this repo to upstream.

For example, two such set of MCP tools are
- https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/actions/createRegisterCatalogEntitiesAction.ts 
- https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/actions/createUnregisterCatalogEntitiesAction.ts

Now, ideally integration tests could be added upstream alongside those, but that would involve first selling upstream on the value.

I wonder if similar to what we are doing with MCP tools themselves, if we could first produce the integrations tests for those upstream tools here, and then remove them from here if we are able to integration tests for them upstream.

Not to dissimilar to providing test coverage for the community plugins that RHDH contributes / maintains upstream.

WDYT @HusneShabbir ?

HusneShabbir:
Thanks for raising this @gabemontero 

On integration tests for `register-entity` / `unregister-entity` **in this PR**: we can’t add meaningful MCP integration coverage for those tools in `rhdh-plugins` **without** also carrying the action implementations here.

`app-config` already includes `catalog` in `pluginSources`, but our pinned `@backstage/plugin-catalog-backend` doesn’t ship those actions yet (they exist on Backstage `master`, not in the version we build against). MCP integration tests start a real backend and call `tools/list` / `tools/call` — if nothing registers `register-entity` / `unregister-entity`, the tests have nothing to exercise.

So the options are:

1. **Wait for upstream** — bump `catalog-backend` once those actions are released, then add integration tests against `catalog.register-entity` / `catalog.unregister-entity` (no overlay sources in this repo).
2. **Temporary overlay** — port the actions into `software-catalog-mcp-extras` *and* add tests here, then delete both when upstream ships them + tests (same pattern as the tools themselves).

I’m not in favor of (2) in *this* PR unless we explicitly want that bridge; this PR stays focused on the overlay MCP tools we already ship here. Happy to open a follow-up for upstream-style catalog tests once the dependency includes those actions, or to spike (2) in a separate PR if we want tests-before-upstream.

cc: @johnmcollier

gabemontero:
> Thanks for raising this @gabemontero
> 
> On integration tests for `register-entity` / `unregister-entity` **in this PR**: we can’t add meaningful MCP integration coverage for those tools in `rhdh-plugins` **without** also carrying the action implementations here.
> 
> `app-config` already includes `catalog` in `pluginSources`, but our pinned `@backstage/plugin-catalog-backend` doesn’t ship those actions yet (they exist on Backstage `master`, not in the version we build against). MCP integration tests start a real backend and call `tools/list` / `tools/call` — if nothing registers `register-entity` / `unregister-entity`, the tests have nothing to exercise.
> 
> So the options are:
> 
> 1. **Wait for upstream** — bump `catalog-backend` once those actions are released, then add integration tests against `catalog.register-entity` / `catalog.unregister-entity` (no overlay sources in this repo).
> 2. **Temporary overlay** — port the actions into `software-catalog-mcp-extras` _and_ add tests here, then delete both when upstream ships them + tests (same pattern as the tools themselves).
> 
> I’m not in favor of (2) in _this_ PR unless we explicitly want that bridge; this PR stays focused on the overlay MCP tools we already ship here. Happy to open a follow-up for upstream-style catalog tests once the dependency includes those actions, or to spike (2) in a separate PR if we want tests-before-upstream.
> 
> cc: @johnmcollier

fair enough @HusneShabbir  .... let's use a upcoming team meeting to discuss voice to voice with the rest of the team which of those potential f/up issues we pursue

otherwise consider my review done

I'll defer to @johnmcollier for any final approvals for merge you need

I see one requested change from him in this PR's status as I type this comment.

thanks

johnmcollier (approved):

PR #3238, opened 2026-05-26
fix(x2a): show correct phase duration and attempt count

mareklibra:
Fixes: FLPATH-3525

Use telemetry timestamps for phase duration instead of aggregated job-level times that include failed pod restarts.

Add attemptCount and firstAttemptAt enrichment fields to the Job schema.

Display attempts and total elapsed time as separate fields in PhaseDetails.

---

Screenshot From 2026-05-25 16-11-02

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/src/services/X2ADatabaseService/jobs.test.ts
+++ b/workspaces/x2a/plugins/x2a-backend/src/services/X2ADatabaseService/jobs.test.ts
@@ -14,12 +14,12 @@
  * limitations under the License.
  */
 
-import { mockCredentials } from '@backstage/backend-test-utils';
-
 import {
   Artifact,
   toSorted,
 } from '@red-hat-developer-hub/backstage-plugin-x2a-common';
+
+import { mockServices, mockCredentials } from '@backstage/backend-test-utils';
 
 import {
   artifactsFromValues,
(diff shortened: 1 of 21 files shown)


workspaces/x2a/plugins/x2a-backend/src/services/X2ADatabaseService/jobs.test.ts
```diff
@@ -773,3 +776,150 @@ describe('X2ADatabaseService – jobs', () => {
...
+          projectId: project.id,
+        });
+
+        const firstJob = await service.createJob({
+          projectId: project.id,
+          moduleId: module.id,
+          phase: 'analyze',
+          startedAt: new Date('2024-01-01T10:00:00Z'),
+        });
+        await delay(5);
```
elai-shalev: why are these needed?

elai-shalev (approved):
looks good, one merge conflicts resolve lgtm

workspaces/x2a/plugins/x2a-backend/src/services/X2ADatabaseService/jobs.test.ts
mareklibra: Thanks, it's a leftover. The `createJob` used to be called without `startedAt` but it was added later on.

Removing it.

PR #3260, opened 2026-06-01
fix(orchestrator): restore status and date filters on workflow runs

lokanandaprabhu:
### Summary

- Fixes an issue where filtering workflow runs by **status** or **date** could show an error instead of results
- Restores the correct filter types used when querying workflow instances after the recent query variable change

https://github.com/user-attachments/assets/6f563d62-4d69-44aa-bbb8-5d31a5032b6c


lholmquist (approved):
Looks good.  Ran a couple workflows then used the drop downs to filter and worked as expected.  No errors in the chrome dev tools either

PR #3265, opened 2026-06-02
Enhance Validations and add Code-Highlighter

asmasarw:
- FLPATH-4260 | [DCM] Rego code field has no monospace font or syntax highlighting
- FLPATH-4290 | [DCM] No duplicate field path validation for catalog items
- FLPATH-4286 | [DCM] No Rego syntax validation
- FLPATH-4291 | [DCM] No client-side validation of instance field values against JSON Schema
- FLPATH-4251 | [DCM] [Code-only] Invalid JSON in catalog item field default_value silently coerced to string
- FLPATH-4281 | [DCM] Rego code field uses proportional font

### DIFF
--- a/workspaces/dcm/plugins/dcm/src/pages/catalog-item-instances/instanceFormTypes.ts
+++ b/workspaces/dcm/plugins/dcm/src/pages/catalog-item-instances/instanceFormTypes.ts
@@ -40,6 +40,8 @@ export type UserValueRow = {
   schemaMin?: number;
   /** Inclusive upper bound (from validation_schema.maximum). */
   schemaMax?: number;
+  /** Whether the field must be filled (from validation_schema.required). */
+  required?: boolean;
 };
 
 export type InstanceForm = {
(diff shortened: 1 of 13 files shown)


workspaces/dcm/plugins/dcm/src/pages/catalog-item-instances/instanceFormTypes.ts
```diff
@@ -94,22 +98,83 @@ function defaultToString(val: unknown): string {
   }
 }
 
+function pickNumericBound(
```
mareklibra: This seems to be copy&paste with the catalogItemFormTypes.ts. Can it be shared? Like moving this to a file with utility functions.

workspaces/dcm/plugins/dcm/src/pages/catalog-items/CatalogItemsTabContent.tsx
```diff
@@ -248,7 +248,7 @@ export function CatalogItemsTabContent() {
     <Drawer
       anchor="right"
       open={open}
-      onClose={submitting ? undefined : onClose}
+      onClose={submitting || Boolean(error) ? undefined : onClose}
```
mareklibra, line 251: Isn't the user trapped in when an error is rendered? The user would need to use the `X` button first, then close. Sounds weird but it might be he intention

mareklibra (changes requested):

workspaces/dcm/plugins/dcm/src/pages/catalog-item-instances/instanceFormTypes.ts
asmasarw: Correct - forget to enhance this.

workspaces/dcm/plugins/dcm/src/pages/catalog-items/CatalogItemsTabContent.tsx
asmasarw: We have two scenarios:
- User clicks on "X" -> Drawer closes always unless they're submitting an api request
- User has an error and he click anywhere in the screen -> Drawer keeps open

And this is totally fine - we need to keep drawer open so user can fix the issues, unless he decided to go away and close it -> by pressing X icon.

mareklibra (approved):

PR #3296, opened 2026-06-04
Security: Sanitize Lightspeed Core error responses

rajin-kichannagari:

Fixes an information disclosure vulnerability where error responses from Lightspeed Core were leaking internal details to clients.

**The issue:**
When LCS returns errors, we were forwarding the full error message to clients, which could include model names, provider info, org IDs, stack traces, etc. Not great for security.

**What I changed:**
- Added a `sanitizeLcsError()` function that logs the full error server-side but only returns a generic message to the client
- Updated the error handling in 4 endpoints: `/v1/feedback`, `/v1/query/interrupt`, `/v1/query`, and `/v2/conversations/:id`
- Updated tests to make sure we're not leaking internal details

**Example:**
Before: `{"error": "Error from lightspeed-core server: Model gpt-4-0613 failed with OpenAI API error: rate limit exceeded for organization org-abc123"}`

After: `{"error": "Error from lightspeed-core server while processing query"}`

Full error details are still logged on the server for debugging.

- [x] A changeset describing the change and affected packages

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
@@ -66,6 +66,30 @@ interface StaticMcpServer {
 }
 
 /**
+ * Sanitizes Lightspeed Core Service (LCS) error responses to prevent
+ * information disclosure. Logs full error details server-side for debugging
+ * while returning only generic messages to clients.
+ *
+ * @param errorBody - The error response body from LCS
+ * @param logger - Logger instance for server-side logging
+ * @param context - Context string describing the operation (e.g., "sending feedback")
+ * @returns Generic error message safe to return to clients
+ */
+function sanitizeLcsError(
+  errorBody: any,
+  logger: any,
+  context: string,
+): string {
+  // Log full error details server-side for debugging
+  logger.error(
+    `Error from lightspeed-core server while ${context}: ${JSON.stringify(errorBody)}`,
+  );
+
+  // Return only generic message to client (no internal LCS details)
+  return `Error from lightspeed-core server while ${context}`;
+}
+
+/**
  * Build MCP-HEADERS for LCS.  Format matches the LCS "client" auth model:
  *   { "server-name": { "Authorization": "<token>" } }
  *
@@ -565,13 +589,15 @@ export async function createRouter(
       );
 
       if (!fetchResponse.ok) {
-        // Read the error body
         const errorBody = await fetchResponse.json();
-        const errormsg = `Error from lightspeed-core server: ${errorBody.error?.message || errorBody?.detail?.cause || 'Unknown error'}`;
-        logger.error(errormsg);
+        const sanitizedError = sanitizeLcsError(
+          errorBody,
+          logger,
+          'sending feedback',
+        );
 
         response.status(fetchResponse.status).json({
-          error: errormsg,
+          error: sanitizedError,
         });
 
         return;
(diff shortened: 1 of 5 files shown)


workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -65,6 +61,30 @@ interface StaticMcpServer {
...
+): string {
+  // Log full error details server-side for debugging
+  logger.error(
+    `Error from lightspeed-core server while ${context}: ${JSON.stringify(errorBody)}`,
+  );
+
+  // Return only generic message to client (no internal LCS details)
+  return `Error from lightspeed-core server while ${context}`;
+}
+
```
Jdubrick: Can we move this to a new file in /service called `utils.ts` or something? Then we can use this as a general sanitizing function for all routers (notebooks, main lightspeed) and any new ones that come in the future?

Jdubrick (commented):
1 small comment, rest looks good

JslYoon (changes requested):
Overall looks good, although the sanitizeLcsError() function uses any for both errorBody and logger
parameters. We should add proper TypeScript types:

- logger should use the LoggerService type (already imported in types.ts from
@backstage/backend-plugin-api).
- errorBody should have an interface that matches the LCS error response
structure (see the test mocks for examples of the shape).
- for the function sanitizeLcsError lets capitalize the LCS bit.

If you can also address the SonarCloud analysis and failing CI after rebasing, I can take another look.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param context - Context string describing the operation
+ * @param response - Express response object
+ */
+export async function handleLCSFetchError(
+  fetchResponse: Response,
+  logger: LoggerService,
+  context: string,
+  response: any,
+): Promise<void> {
+  const errorBody = await fetchResponse.json();
```
Jdubrick: Let's wrap this in a `try/catch` just on the off chance Lightspeed Core doesn't return JSON / doesn't have a body, on the catch we can just set the body to an empty object

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
```diff
@@ -0,0 +1,73 @@
...
+ * @param fetchResponse - The failed fetch response from LCS
+ * @param logger - Logger instance for server-side logging
+ * @param context - Context string describing the operation
+ * @param response - Express response object
+ */
+export async function handleLCSFetchError(
+  fetchResponse: Response,
+  logger: LoggerService,
+  context: string,
+  response: any,
```
Jdubrick: Is it possible to type this to `express.Response` or is ts being weird about it and that is why it's `any`?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
Jdubrick: Then let's add a test for this case as well

JslYoon (approved):
lgtm, make sure all other comments are address too before merging

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
rajin-kichannagari: It was, but I changed it from `any` to an inline type that matches the Express response interface we need. Avoided importing `express.Response` to keep the utils file lightweight since none of the other functions need Express types.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
rajin-kichannagari: Good call! Added a try/catch that sets errorBody to an empty object if JSON parsing fails, plus tests for both the normal JSON case and the non-JSON error case.

Jdubrick (approved):
/lgtm

workspaces/lightspeed/plugins/lightspeed-backend/src/service/utils.ts
Jdubrick: I personally think it shows that it is a Express Response very explicitly by using that type, but I won't hold the PR for that, it's up to you. I lgtm'd it

PR #3330, opened 2026-06-08
docs(augment): add OpenSpec proposal for per-user Kagenti token exchange

gabemontero:

Add OpenSpec artifacts for the kagenti-user-level-token-exchange change, which implements RFC 8693 OAuth2 Token Exchange for the Kagenti provider enabling per-user authorization instead of shared service-account tokens.

Artifacts created:
- proposal.md: motivation, capabilities (token-exchange, user-token-routing), impact
- design.md: 5 architectural decisions with alternatives, risks, and constraints
- specs/token-exchange/spec.md: 7 requirements, 15 scenarios (config, RFC 8693 execution, caching, dedup, streaming, fallback, no-impact guarantee)
- specs/user-token-routing/spec.md: 6 requirements, 8 scenarios (header routing, route extraction, interface widening, backward compatibility)
- tasks.md: 6 task groups, 19 implementation tasks ordered by dependency

Includes the preliminary implementation plan used as source material.

- [n/a] Screenshots attached (for UI changes)


workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,122 @@
...
+RFC 8693 OAuth2 token exchange for per-user Kagenti authorization.
+
+## ADDED Requirements
+
+### Requirement: Token exchange configuration schema
+
+The system SHALL support an optional `tokenExchange` configuration block nested under `augment.kagenti.auth` with the following fields:
+
+- `enabled` (boolean, default: `false`) — whether to attempt per-user token exchange
+- `audience` (string, default: value of `auth.clientId`) — the `audience` parameter in the RFC 8693 exchange request
```
pkliczewski: do you think it makes sense to list supported values?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- **THEN** the system SHALL use `kagenti-api` as the audience parameter in exchange requests
+
+#### Scenario: Token exchange audience defaults to clientId
+
+- **WHEN** `tokenExchange.enabled` is `true` and `audience` is not specified
+- **THEN** the system SHALL use the value of `auth.clientId` as the audience parameter
+
+### Requirement: RFC 8693 token exchange execution
+
+The `TokenExchangeManager` SHALL perform OAuth2 Token Exchange per RFC 8693 by sending a POST to the configured `tokenEndpoint` with:
```
pkliczewski: to be consistent with the document should we use `auth.tokenEndpoint`?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- `subject_token`: the user's OIDC access token
+- `subject_token_type`: `urn:ietf:params:oauth:token-type:access_token`
+- `audience`: the configured audience value
+- `client_id` and `client_secret`: from the parent auth config
+
+The exchanged token SHALL preserve the user's `sub` claim and add an `act` (actor) claim identifying the service.
+
+#### Scenario: Successful token exchange
+
+- **WHEN** a valid user OIDC token is provided and Keycloak supports token exchange
```
pkliczewski: what is the use case for Keycloak not supporting token exchange? according to the docs `token-exchange-standard:v2` is enabled by default for Keycloak.

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,122 @@
...
+- **THEN** the system SHALL return a Kagenti-scoped access token with the user's `sub` claim preserved
+
+#### Scenario: Exchange request format
+
+- **WHEN** a token exchange is initiated
+- **THEN** the POST body SHALL include `grant_type`, `subject_token`, `subject_token_type`, `audience`, `client_id`, and `client_secret` as form-urlencoded parameters
+
+### Requirement: Per-user token caching
+
+The `TokenExchangeManager` SHALL cache exchanged tokens keyed by user identity. Cached tokens SHALL be reused for subsequent requests from the same user until they expire or are invalidated.
```
pkliczewski: what about refresh token? should we keep the token up to date if not invalidated?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,122 @@
...
+#### Scenario: Token not evicted during active stream
+
+- **WHEN** a streaming request is using an exchanged token and the token's TTL expires
+- **THEN** the system SHALL NOT evict the token until the stream completes
+
+### Requirement: Graceful fallback to service-account token
+
+The system SHALL fall back to the existing `KeycloakTokenManager` service-account token whenever per-user token exchange cannot complete. Fallback SHALL occur silently with a warning log — requests SHALL NOT fail due to exchange issues.
+
+#### Scenario: Token exchange disabled
```
pkliczewski: this scenario seems to be the same as the one in line #17

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/user-token-routing/spec.md
```diff
@@ -0,0 +1,69 @@
+# Spec: user-token-routing
+
+Route-level extraction and forwarding of user OIDC tokens from configurable request headers.
+
+## ADDED Requirements
+
+### Requirement: Configurable user token header on RouteContext
+
+The `RouteContext` type SHALL include an optional `userTokenHeader` field containing the header name from which to extract the user's OIDC token. The value SHALL be sourced dynamically from the `KagentiProvider` configuration.
```
pkliczewski: do we need an example here?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
```diff
@@ -0,0 +1,89 @@
+# Design: Per-User OAuth2 Token Exchange for Kagenti Provider
+
+## Context
+
+The augment plugin's Kagenti provider authenticates to Kagenti using a single shared service-account token via Client Credentials Grant (`KeycloakTokenManager`). All backend requests to Kagenti carry the same identity regardless of which Backstage user initiated the request. The user's identity is passed only as an informational `X-Backstage-User` header.
```
pkliczewski, line 5: I am puzzled about this direction. In the orchestrator we implemented passing user identity to sonataFlow. Which means we use backstage providers defined and we force users to authenticate to pass valid tokens. Why not to follow similar approach here?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
```diff
@@ -0,0 +1,89 @@
...
+- Custom Keycloak realm or client configuration tooling
+- Token exchange for the `X-Backstage-User` header (informational only)
+
+## Decisions
+
+### Decision 1: Backend-only with configurable header source
+
+Accept the user's OIDC token from a configurable request header rather than coupling the frontend to OIDC auth. The token can be injected by:
+
+1. An auth proxy (oauth2-proxy, Keycloak Gatekeeper) — many RHDH deployments already have one
```
pkliczewski: how this will work with custom auth mechanism used by the customer?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
```diff
@@ -0,0 +1,89 @@
...
+
+**Alternative considered:** Backstage middleware extracting from session. Rejected because it couples to Backstage auth internals and requires session storage access.
+
+### Decision 2: New `TokenExchangeManager` modeled on `KeycloakTokenManager`
+
+Create a dedicated manager class rather than extending `KeycloakTokenManager`. The exchange manager handles per-user token caching (keyed by user), concurrent request deduplication, streaming-aware token lifetime, and exchange-specific error handling.
+
+**Alternative considered:** Extending `KeycloakTokenManager` with user-keyed methods. Rejected because the two have different lifecycles (system-wide vs. per-user), different cache semantics, and mixing them would complicate the clean fallback logic.
+
+### Decision 3: Graceful fallback at every stage
```
pkliczewski: why not to fail in this case? fallback maybe hard to debug and diagnose.

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
```diff
@@ -0,0 +1,89 @@
...
+        audience: kagenti-api # default: auth.clientId
+        userTokenHeader: X-Forwarded-Access-Token # default: x-user-oidc-token
+```
+
+This reuses the existing `tokenEndpoint`, `clientId`, and `clientSecret` from the parent auth block. No new top-level config keys.
+
+## Risks / Trade-offs
+
+- **Auth proxy not configured** → No OIDC token in header → falls back to service-account silently. Deployments without an auth proxy get no per-user auth until they add one or a frontend solution is built.
+- **Keycloak doesn't support token exchange** → Returns 400 `unsupported_grant_type`. Caught, warned, falls back. Requires Keycloak admin to enable token exchange on the realm.
```
pkliczewski, line 134: why not to document that it is needed?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
```diff
@@ -0,0 +1,89 @@
...
+- **OIDC token expired** → Exchange fails, caught, falls back. Short-lived tokens may cause frequent fallbacks; auth proxy should handle refresh.
```
pkliczewski: why not to use refresh token to keep OIDC token alive (valid)?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: claude and I agree ... will update with my next push

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: yep will be part of next push

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: yeah it is a defensive measure for older keycloak versions or non-Keycloak OIDC providers

I'll add a clarifying note in the next push, but do continue on this thread if you are not sold on the defensive measure

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: yeah during the openspec processing this got put in the deferred as a future enhancement category

for now the next push adds a clarifying note along these lines, but do continue on this thread if you are not sold on that decision

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: consolidated  in next push

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/user-token-routing/spec.md
gabemontero: added an example (but only in a parenthetical for now) 

after I push the changes let me know if the example is sufficient

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: Good question — we did consider following the Backstage auth provider pattern. The main blocker is that createApiFactory requires all deps to be resolvable at startup — there's no optional dependency concept. Adding oidcAuthApiRef as a hard dependency would crash deployments that don't use OIDC auth, and useApi() throws synchronously during render if the API isn't
  registered. This matters for us because the augment plugin needs to work across deployment configurations — not all RHDH deployments use Keycloak/OIDC as their auth provider, and we don't want to force that as a prerequisite just to enable the Kagenti integration.
  
  That's why Decision 1 in the design went with accepting the OIDC token from a configurable request header instead — it keeps token exchange opt-in without coupling the frontend to a specific auth provider.

  That said, if the orchestrator team found a way to handle the hard-dependency problem with createApiFactory (or if requiring OIDC auth was acceptable in that context), I'd love to understand the pattern you used. It could be a better approach here too, or at least an alternative alongside the header-based path.

  Could you point me at how the orchestrator passes user identity to SonataFlow? Specifically how it gets the user's Keycloak OIDC token in the frontend without the createApiFactory constraint being an issue.

gabemontero: Here is what claude/opus came up with when analyzing the orchestrator code:
```
I looked at the orchestrator's approach in useOrchestratorAuth.ts. The pattern there is different from our use case in a couple of ways:
  
  1. The orchestrator collects provider-scoped tokens (GitHub, GitLab, Microsoft) to forward to SonataFlow for external API access. The three built-in providers are always registered by Backstage, so useApi() works safely for those. For custom providers, it uses useApiHolder() with dynamic discovery to avoid the static createApiFactory dependency.
  2. For our case, we specifically need the user's Keycloak OIDC token to perform RFC 8693 token exchange against Keycloak. The orchestrator's findCustomProvider does reference internal.auth.oidc from RHDH (line 100-101 in useOrchestratorAuth.ts), but it accesses the API holder's internal map via @ts-ignore — it's reaching into a private API.

  The createApiFactory constraint matters for us because the augment plugin needs to work across deployment configurations — not all RHDH deployments use Keycloak/OIDC as their auth provider, and we don't want to force that as a prerequisite just to enable the Kagenti integration. That's why the design went with accepting the OIDC token from a configurable request
  header — it keeps token exchange opt-in and decoupled from the frontend auth provider configuration.

  That said, the orchestrator's useApiHolder() dynamic discovery pattern is worth exploring as an alternative frontend path in the future. If RHDH stabilizes a public API for accessing the OIDC auth provider, we could add a frontend option alongside the header-based approach.
```
so yeah, if the `useApiHolder` bit is part of the underlying orchestrator work you were referring to, perhaps we reassess and go down that path initiall vs. in the future 

WDYT @pkliczewski ?

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: adding a comment to clear this up

 the header source is decoupled from any specific auth provider. A customer with a custom auth mechanism just needs to get a valid OIDC-compatible token into the configured header, however they choose to do that.

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: I think it is a fair point that some cases might be hard to debug, and warrant louder signals, but I still think there are others where graceful fallback is acceptable

will update to provide some detail on the two cases

let's revisit this thread after I post the update, and see whether that is sufficient for you, or if we need to discuss more

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: fair point ... will attempt to address in next update

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: will attempt to clarify why in the next update (essentially don't have access to the refresh token at this point) ... somewhat related to an earlier comment 

see what you think after reading the clarification, and we can then see about discussing this further or not

gabemontero:
thanks for the thorough review @pkliczewski 

I've pushed the new commit with changes stemming from your comments

I've refrained from resolving any conversations .... certainly at least some of them will require more iteration on our part (and possibly non-trivial adjustments) after you see the updates

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
pkliczewski: I know we decided to stay backend. With the orchestrator we use frontend for asking user to login to specific provider if token is not available. In this case a user can login with any provider they want and we will ask them to login with a provider like Keycloak.

If backend only approach I agree with you but if we could ask user to login when connecting to kagenti for the first time it would simplify this change. Maybe we could come up with the flow to force user to login before we connect to kagenti.

pkliczewski:
@gabemontero thanks, resolved most of the comments

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/design.md
gabemontero: so claude and I explored the `useApiHolder` bit we discovered from our orchestrator examination and I believe we've incorporated it to provide what you are asking for when kagenti/oidc is available, but dynamically bypass this when it is not available

PTAL

gabemontero:
> @gabemontero thanks, resolved most of the comments

thanks @pkliczewski - remaining comment addressed and commit pushed

gabemontero:
fyi @pkliczewski I've gotten agreement from @PatAKnight to review this proposal as well

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
```diff
@@ -0,0 +1,124 @@
...
+### Requirement: Streaming-aware token lifetime
+
+The `TokenExchangeManager` SHALL support streaming-aware token lifetime management. Streaming requests SHALL hold a reference that prevents the token from being evicted while the stream is active.
+
+#### Scenario: Token not evicted during active stream
+
+- **WHEN** a streaming request is using an exchanged token and the token's TTL expires
+- **THEN** the system SHALL NOT evict the token until the stream completes
+
+### Requirement: Graceful fallback to service-account token
```
PatAKnight: I wonder if this should be something that is configurable? Maybe have it so that environments where they have more strict access and security postures can opt to have no fallback.

PatAKnight (commented):
I don't have the most solid grasp on Auth in this context so I might not be able to catch the more minor details but overall it looks fine to me. I did notice that in the middle you were talking about using the frontend with `useApiHolder` but the `proposal.md` and `tasks.md` all mention backend specific work.

gabemontero:
Thanks for the review @PatAKnight!

Good catch on the frontend/backend disconnect — the design.md was updated to include frontend OIDC discovery via `useApiHolder()` after pkliczewski's earlier feedback about the orchestrator pattern, but the proposal.md and tasks.md weren't synchronized to match.

I've clarified the phasing: **this change covers the backend-only implementation** (header-based OIDC token acquisition). Frontend OIDC discovery via `useApiHolder()` is documented as a planned follow-up change. Updated the proposal.md, design.md, tasks.md, and user-token-routing spec to make this clear.

Also addressed your inline comment about configurable fallback — see the reply there.

workspaces/augment/openspec/changes/kagenti-user-level-token-exchange/specs/token-exchange/spec.md
gabemontero: Great point — agreed. I've added a `fallbackToServiceAccount` boolean (default: `true`) to the `tokenExchange` config block.

When `true` (default): exchange failures fall back to the service-account token with a warning log — same as before, safe for initial rollout and transitioning deployments.

When `false` (strict mode): exchange failures return an error response (401 for missing token, 502 for exchange failure) instead of falling back. This gives environments with strict security postures the hard guarantee that requests won't silently downgrade to a shared service-account identity.

Updated the spec.md (config schema + new strict mode scenarios), design.md (Decision 3 rewritten for configurable behavior), tasks.md, and proposal.md to reflect this. Pushing shortly.

gabemontero:
@PatAKnight @pkliczewski @rrbanda  so after another round of deciding whether this current proposal is truly viable,  I think I've convinced myself that this staged approach of depending on a 3rd party to inject the user's OIDC access token is not sufficient (i.e. customers won't want to set that up, don't already have it), and we need to bring in user OIDC discovery back to the augment frontend.

About to push a couple a commit that does that

This also makes me wonder if we go ahead and pursue #3331 (finer grained RBAC) beforehand and independent of this change.

PR #3331, opened 2026-06-08
docs(augment): add OpenSpec proposal for fine-grained Backstage permissions

gabemontero:

Add OpenSpec artifacts for the fine-grained-backstage-permissions change, which replaces 12+ inline route-level authorization guards with proper Backstage fine-grained permissions, enabling deployers to configure RBAC policies for agent and tool lifecycle governance.

Artifacts created:
- proposal.md: motivation, 3 capabilities (permission-definitions, authorization-middleware, route-authorization), impact across 12 files
- design.md: 5 architectural decisions covering resource types, two-tier fallback, self-approval defense-in-depth, visibility filtering, and rule patterns
- specs/permission-definitions/spec.md: 11 requirements (2 resource types, 16 permissions, 3 permission rules with ownership/stage/self-approval)
- specs/authorization-middleware/spec.md: 5 requirements (two-tier auth, conditional evaluation, RouteContext integration)
- specs/route-authorization/spec.md: 14 requirements covering all agent, tool, and Kagenti route authorization replacements with backward compat
- tasks.md: 7 task groups, 30 implementation tasks ordered by dependency

Includes the preliminary implementation plan used as source material.

- [/] Tests for new functionality and regression tests for bug fixes
- [/] Screenshots attached (for UI changes)

### DIFF
--- a/workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
+++ b/workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
@@ -14,10 +14,9 @@ These permissions are independent of AgenticProvider authorization. Kagenti's pe
 
 - Replace all 12+ inline authorization decisions with Backstage permission framework calls
 - Enable deployers to configure fine-grained RBAC policies for agent and tool lifecycle operations and targeted access to infrastructure resources (vector stores, documents, MCP connections, prompts, models)
-- Provide an opt-in legacy fallback (`permissions.legacyAdminFallback`) so existing deployments using `augment.access` + `augment.admin` can migrate incrementally
+- Provide an opt-in legacy fallback (`augment.permissions.legacyAdminFallback`) so existing deployments using `augment.access` + `augment.admin` can migrate incrementally
 - Support conditional permission rules for ownership checks, self-approval prevention, and lifecycle stage gating
 - Keep self-approval prevention as defense-in-depth (permission rule supplements hard-coded check)
-- Emit structured audit log entries for all authorization decisions — recording the user, action, resource, outcome (allow/deny), and whether the `augment.admin` fallback was used. The authorization middleware is the single point through which all permission decisions flow, making it the natural place for audit logging.
 
 **Non-Goals:**
 
@@ -37,7 +36,7 @@ Define `augment-agent` and `augment-tool` as distinct resource types rather than
 
 ### Decision 2: Opt-in legacy fallback to `augment.admin`
 
-When `permissions.legacyAdminFallback` is enabled in the plugin config, `authorizeLifecycleAction` checks the fine-grained permission first. If DENY, it falls back to checking `augment.admin`. When the config flag is absent or false (the default), only fine-grained permissions are evaluated — no fallback occurs.
+When `augment.permissions.legacyAdminFallback` is enabled in the plugin config, `authorizeLifecycleAction` checks the fine-grained permission first. If DENY, it falls back to checking `augment.admin`. When the config flag is absent or false (the default), only fine-grained permissions are evaluated — no fallback occurs.
 
 **Why opt-in rather than default-on:** The augment plugin is in dev preview, which means no backward compatibility guarantees. Making the fallback default-on risks the "temporary becomes permanent" problem — once deployments rely on `augment.admin` as a catch-all, removing the fallback later becomes the very breaking change it was meant to prevent. Opt-in gives existing consumers a migration path (enable the flag, then incrementally adopt fine-grained policies, then disable the flag) without baking the fallback into every deployment's baseline.
 
@@ -51,11 +50,11 @@ The `IS_NOT_CREATOR` permission rule enables RBAC-level self-approval prevention
 
 **Alternative considered:** Removing the hard-coded check entirely and relying solely on the permission rule. Rejected because a misconfigured RBAC policy could silently disable self-approval prevention — defense-in-depth is safer for a governance control.
 
-### Decision 4: Visibility filtering via basic permission
+### Decision 4: Visibility filtering via resource-based permission with 3-tier evaluation
 
-`augment.agent.list` is a basic (non-resource) permission that controls visibility: ALLOW shows all agents, DENY filters to published + user's own agents. This avoids evaluating resource-based conditions against every agent in the list.
+`augment.agent.list` is a resource-based permission with resource type `augment-agent`, supporting 3-tier evaluation: ALLOW shows all agents, DENY shows no agents, CONDITIONAL applies the returned conditions as filters against each agent in the list (e.g., `IS_OWNER` returns only the user's own agents, `HAS_LIFECYCLE_STAGE(published)` returns only published agents). This aligns with the orchestrator plugin's existing CONDITIONAL policy patterns and gives deployers full flexibility over visibility rules via RBAC configuration.
 
-**Alternative considered:** Resource-based permission evaluated per-agent in list responses. Rejected due to performance — evaluating conditional rules against potentially hundreds of agents per list request is expensive and unnecessary when the distinction is binary (admin vs. filtered view).
+**Alternative considered:** Basic (non-resource) permission with binary ALLOW/DENY where DENY hardcodes a "published + own" filter. Rejected because (a) DENY-means-filtered is non-standard permission semantics — deployers expect DENY to mean no access, (b) the hardcoded filter cannot be customized per-deployment, and (c) the performance cost of evaluating conditions per-agent is comparable to the hardcoded filter iteration we'd be doing anyway.
 
 ### Decision 5: Permission rules modeled on extensions-backend pattern
 
@@ -63,7 +62,7 @@ Permission rules (`IS_OWNER`, `IS_NOT_CREATOR`, `HAS_LIFECYCLE_STAGE`) follow th
 
 ## Risks / Trade-offs
 
-- **Policy migration complexity** → Deployers must configure fine-grained RBAC policies. Existing deployments can enable `permissions.legacyAdminFallback` to preserve current `augment.admin` behavior during migration.
+- **Policy migration complexity** → Deployers must configure fine-grained RBAC policies. Existing deployments can enable `augment.permissions.legacyAdminFallback` to preserve current `augment.admin` behavior during migration.
 - **Conditional evaluation overhead** → Resource-based permissions require loading the resource to evaluate conditions. Mitigated by keeping list operations as basic permissions and only using resource-based permissions for mutation routes where the resource is already loaded.
 - **Rule mismatch on upgrade** → If a deployer configures a fine-grained policy with a `HAS_LIFECYCLE_STAGE` condition referencing a stage name that changes, the rule silently denies. Mitigated by documenting stage names as part of the permission contract.
 - **Defense-in-depth dual check** → The self-approval hard-coded check and `IS_NOT_CREATOR` rule are redundant by design. If one is relaxed without updating the other, behavior may be confusing. Mitigated by documenting this clearly.
--- a/workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
+++ b/workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
@@ -1,0 +1,66 @@
+# Design: Fine-Grained Backstage Permissions for Augment Lifecycle Governance
+
+## Context
+
+The augment plugin currently enforces 12+ authorization decisions via custom inline guards in route handlers — `checkIsAdmin`, `createdBy` comparisons, lifecycle stage checks, and self-approval prevention. These bypass Backstage's permission framework entirely. Only two coarse permissions exist (`augment.access` for read, `augment.admin` for admin operations), making fine-grained RBAC impossible.
+
+Backstage provides a permission framework (`@backstage/plugin-permission-node`) supporting basic permissions, resource-based permissions with conditional rules, and policy evaluation via the permission backend. The `extensions-backend` plugin in this workspace already follows this pattern and serves as a reference implementation.
+
+These permissions are independent of AgenticProvider authorization. Kagenti's per-user RBAC (via RFC 8693 token exchange) and OpenAI's organization RBAC govern access to external AI services. The permissions defined here govern access within the augment plugin's own governance system — lifecycle state transitions, ownership-scoped operations, self-approval prevention, and visibility filtering. The two layers operate on different verbs, different resources, and different data stores.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Replace all 12+ inline authorization decisions with Backstage permission framework calls
+- Enable deployers to configure fine-grained RBAC policies for agent and tool lifecycle operations
+- Maintain full backward compatibility — existing `augment.access` + `augment.admin` policies continue to work unchanged
+- Support conditional permission rules for ownership checks, self-approval prevention, and lifecycle stage gating
+- Keep self-approval prevention as defense-in-depth (permission rule supplements hard-coded check)
+
+**Non-Goals:**
+
+- Modifying `augment.access` or `augment.admin` behavior or semantics
+- Adding permissions for AgenticProvider-level operations (Kagenti API calls, OpenAI API calls)
+- Frontend permission checks (backend-only enforcement)
+- Custom permission policy plugins (works with standard Backstage RBAC)
+- Removing the hard-coded self-approval prevention check (kept as defense-in-depth alongside the `IS_NOT_CREATOR` rule)
+
+## Decisions
+
+### Decision 1: Separate resource types for agents and tools
+
+Define `augment-agent` and `augment-tool` as distinct resource types rather than a single `augment-resource` type. Agents and tools have different route handlers, different lifecycle transitions, and deployers may want independent RBAC scoping.
+
+**Alternative considered:** Single `augment-resource` type with a discriminator field. Rejected because it conflates two domain objects that are independently routed and independently targetable by policy.
+
+### Decision 2: Two-tier authorization with backward-compatible fallback
+
+`authorizeLifecycleAction` checks the fine-grained permission first. If DENY (the default when no policy exists), it falls back to checking `augment.admin`. This means deployments that haven't configured fine-grained policies get identical behavior to today.
+
+**Alternative considered:** Requiring all deployments to update their RBAC policies. Rejected because it would break every existing augment deployment on upgrade.
+
+**Alternative considered:** OR-combining fine-grained and admin in a single policy evaluation. Rejected because Backstage's permission framework evaluates permissions individually — the fallback must be explicit in code.
+
+### Decision 3: Self-approval prevention as defense-in-depth
+
+The `IS_NOT_CREATOR` permission rule enables RBAC-level self-approval prevention, but the existing hard-coded check (`createdBy !== userRef`) is retained. The hard-coded check is the primary guard; the permission rule allows deployers to customize or relax the policy via RBAC if desired.
+
+**Alternative considered:** Removing the hard-coded check entirely and relying solely on the permission rule. Rejected because a misconfigured RBAC policy could silently disable self-approval prevention — defense-in-depth is safer for a governance control.
+
+### Decision 4: Visibility filtering via basic permission
+
+`augment.agent.list` is a basic (non-resource) permission that controls visibility: ALLOW shows all agents, DENY filters to published + user's own agents. This avoids evaluating resource-based conditions against every agent in the list.
+
+**Alternative considered:** Resource-based permission evaluated per-agent in list responses. Rejected due to performance — evaluating conditional rules against potentially hundreds of agents per list request is expensive and unnecessary when the distinction is binary (admin vs. filtered view).
+
+### Decision 5: Permission rules modeled on extensions-backend pattern
+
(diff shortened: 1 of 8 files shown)


workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
```diff
@@ -0,0 +1,76 @@
+# Spec: authorization-middleware
```
pkliczewski, line 1: the use cases in this md file seems to be covered by RBAC plugin. Is there anything specific to augment?

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
```diff
@@ -0,0 +1,138 @@
...
+#### Scenario: Tool resource type exported
+
+- **WHEN** a consumer imports from `augment-common`
+- **THEN** the `RESOURCE_TYPE_AUGMENT_TOOL` constant SHALL be available with value `augment-tool`
+
+### Requirement: Agent permissions
+
+The system SHALL define the following agent permissions using `createPermission` with resource type `augment-agent`:
+
+| Permission ID             | Action                         |
```
pkliczewski: should we define permissions for all the resource types we want to support and not only agents and tools?

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/route-authorization/spec.md
```diff
@@ -0,0 +1,189 @@
+# Spec: route-authorization
```
pkliczewski, line 1: this seems to be scenarios and requirement for RBAC plugin. Is there anything augment specific here?

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
```diff
@@ -0,0 +1,66 @@
...
+
+Backstage provides a permission framework (`@backstage/plugin-permission-node`) supporting basic permissions, resource-based permissions with conditional rules, and policy evaluation via the permission backend. The `extensions-backend` plugin in this workspace already follows this pattern and serves as a reference implementation.
+
+These permissions are independent of AgenticProvider authorization. Kagenti's per-user RBAC (via RFC 8693 token exchange) and OpenAI's organization RBAC govern access to external AI services. The permissions defined here govern access within the augment plugin's own governance system — lifecycle state transitions, ownership-scoped operations, self-approval prevention, and visibility filtering. The two layers operate on different verbs, different resources, and different data stores.
+
+## Goals / Non-Goals
+
+**Goals:**
+
+- Replace all 12+ inline authorization decisions with Backstage permission framework calls
```
pkliczewski, line 15: not sure whether it is a goal for this doc but I think we should have an auditlog support

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Decisions
+
+### Decision 1: Separate resource types for agents and tools
+
+Define `augment-agent` and `augment-tool` as distinct resource types rather than a single `augment-resource` type. Agents and tools have different route handlers, different lifecycle transitions, and deployers may want independent RBAC scoping.
+
+**Alternative considered:** Single `augment-resource` type with a discriminator field. Rejected because it conflates two domain objects that are independently routed and independently targetable by policy.
+
+### Decision 2: Two-tier authorization with backward-compatible fallback
```
pkliczewski: we are working on a new plugin so I wonder why we need to be backward compatible?

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
gabemontero: The way I think about it, this spec is not reimplementing RBAC policy evaluation (that's the RBAC plugin's job). What it defines is the augment-specific integration layer between the plugin's route handlers and the Backstage permission framework:
  
  1. Two-tier fallback logic — check the fine-grained permission first, then fall back to augment.admin. This is augment domain logic — the RBAC plugin evaluates policies, but it doesn't know that augment wants to fall back to a coarse permission when no fine-grained policy is configured. This fallback is what lets existing deployments work unchanged after upgrade.
  2. Conditional evaluation against augment resource types — when the RBAC plugin returns CONDITIONAL (e.g., "allow if owner"), something needs to evaluate that condition against augment's agent/tool objects (createdBy, lifecycle stage). That's matchesAgentConditions / matchesToolConditions — domain-specific utilities the RBAC plugin can't provide.
  3. RouteContext integration — wiring these into the route context so handlers call ctx.authorizeLifecycleAction(...) instead of importing permission internals.

  Regarding how this relates to the existing permissions:
  - augment.access is unchanged — it gates plugin-level visibility and is orthogonal to fine-grained lifecycle permissions
  - augment.admin continues to gate admin routes (config, documents, vector stores, etc.) that aren't covered by the new permissions. For lifecycle operations (promote, approve, publish, etc.), it becomes a backward-compatible fallback rather than the primary gate. Deployers who adopt fine-grained policies can stop granting augment.admin for lifecycle operations — the fallback is a migration aid, not a permanent architecture.
  
  I've added a clarifying section to the spec to make this relationship explicit.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
gabemontero: I think we can add more later .... I'm not 100% clear what would comprise "all the resource types"

This does start with that was done with the governance system and its focus on agent lifecycle

I'll add some clarifying notes and at least some reference to potential future candidates.  TBH, wouldn't mind some feedback from the customer side before adding more.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/route-authorization/spec.md
gabemontero: If it isn't clear, I'll attempt to clarify in the next update.

But the content here is entirely about which augment routes use which augment permissions with which augment-specific conditions.  The spec here wants to define what the plugin asks the backstage permission framework.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
gabemontero: agreed ... attempted to add with the next push

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
gabemontero: My thinking was that we have an external customer already using the plugin with the existing RBAC model (augment.access + augment.admin). From their perspective, a new drop that requires reconfiguring their RBAC policies to maintain existing access is a cross-release breaking change — even though the plugin itself is still in active development.
  
  The two-tier fallback ensures these deployments continue to work on upgrade without any policy changes, while giving them the option to adopt fine-grained permissions incrementally. I've added a note to Decision 2 to make this rationale explicit.

gabemontero:
thanks again for the thorough review @pkliczewski 

I've pushed a new commit with my updates ... and left the comment threads unresolved

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
pkliczewski: thanks, it felt to me as we would like to define how RBAC should work. if you find it helpful let's keep it.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
pkliczewski: We have support levels, for now this plugin is a consulting work and dev preview should apply here. This means that code still can evolve but there should be no compatibility guarantees.

Let's see what the customer will say. If we need to be backward compatible so be it.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
pkliczewski: for now we have agents, skills, mcp, vector db and prompts. The question is whether we should define resource types for all or only for some.

pkliczewski:
@gabemontero thanks, resolved most of the comments

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
gabemontero: ok I ruminated some more and have been swayed to your inclination here 

made additions to the matrix ... one caveat, skills are not a resource managed by augment to date, including the recently introduced governance model, only something we might leverage from the skills marketplace and/or rhdh 2.1 ai catalog

"tools" are a thing ... i.e. MCP-invokable service deployed to Kagenti. Also has full lifecycle governance with the same stages. Has ChatToolConfig, dedicated lifecycle routes (toolLifecycleRoutes.ts), and Kagenti CRUD routes (kagentiToolRoutes.ts).

The additions will include 
 - augment.vectorstore.manage — vector store CRUD
 - augment.document.manage — document upload/deletion
 - augment.mcp.manage — MCP connections, tool creation/deletion, builds
 - augment.prompt.manage — system prompt generation/config
 - augment.model.manage — model listing, testing, config

 These are basic (non-resource) permissions for now — the operations don't have ownership or lifecycle semantics today, so conditional rules aren't needed. If ownership tracking is added later, they can be upgraded to resource-based permissions with rules like IS_OWNER.

 Each uses the same two-tier fallback to augment.admin, so existing deployments are unaffected. This also closes a gap where MCP tool creation (POST) was previously ungated.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
gabemontero: sounds good ... as part of preparing for that conversation, I decided to pivot the design / spec from defaulting to the old, coarse grained permissions of augment.access and augment.admin to having to opt into the behavior of falling back to the old permissions 

in essence force the discussion

gabemontero:
> @gabemontero thanks, resolved most of the comments

likewise thanks @pkliczewski 

I've posted replies to the remaining review threads and pushed a new commit with updates for those remaining threads

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
pkliczewski: you are right with the tools. thanks

gabemontero:
fyi @pkliczewski I've gotten agreement from @PatAKnight to review this proposal as well

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
```diff
@@ -0,0 +1,120 @@
...
+
+Augment-specific authorization middleware that integrates Backstage's permission framework with the plugin's domain model (agents, tools, lifecycle stages, ownership). This is NOT a reimplementation of RBAC policy evaluation — the RBAC plugin handles that. This middleware provides:
+
+1. **Opt-in legacy fallback logic** — when `permissions.legacyAdminFallback` is enabled, check fine-grained permission first, then fall back to `augment.admin` for backward compatibility during migration. When disabled (the default), only fine-grained permissions are evaluated. This is augment-specific business logic that the RBAC plugin cannot provide, since the fallback semantics (which coarse permission to check, when to check it) are domain decisions.
```
PatAKnight: I would recommended avoiding having the config value `permissions.legacyAdminFallback` and instead doing it under the `augment` config. This way you don't have to worry about the potential (albeit probably small) chance that backstage adds the same value that conflicts with this under `permission`

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
```diff
@@ -0,0 +1,120 @@
...
+- **THEN** access SHALL be denied
+
+### Requirement: Conditional permission evaluation
+
+When `authorizeLifecycleAction` receives a CONDITIONAL result from the permission framework, it SHALL evaluate the conditions against the provided resource using `matchesAgentConditions` or `matchesToolConditions` based on the resource type.
+
+#### Scenario: Conditional result with matching resource
+
+- **WHEN** the permission framework returns CONDITIONAL with an `IS_OWNER` condition, and the resource's `createdBy` matches the requesting user
+- **THEN** `authorizeLifecycleAction` SHALL evaluate the condition to ALLOW
```
PatAKnight: This might be a bit of a nit as I might not be fully understanding what the proposed `authorizeLifecycleAction` would end up doing under the hood. 

But the conditional rules would act more similar to a filter. So, you would get `AuthorizeResult.CONDITIONAL` as a result from the `PermissionService.authorizeConditional`. You would then take the conditions that are also included and convert that into your filter that would be then used to filter the resources that you want. So, I don't know if `SHALL evaluate the condition to ALLOW` would be correct terminology. Maybe more like `SHALL return the appropriate resources based on condition` or something like that.

Example of how the catalog plugin uses conditional evaluation for deletion: https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/service/AuthorizedEntitiesCatalog.ts#L207:L237

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/authorization-middleware/spec.md
```diff
@@ -0,0 +1,120 @@
...
+
+### Requirement: Audit logging for authorization decisions
+
+The `authorizeLifecycleAction` and `authorizeBasicWithFallback` functions SHALL emit a structured audit log entry for every authorization decision. Each entry SHALL include:
+
+- `user` — the requesting user's entity ref
+- `action` — the permission ID evaluated (e.g., `augment.agent.approve`)
+- `resource` — the resource identifier (agent/tool ID), if applicable
+- `outcome` — `allow` or `deny`
+- `grantedVia` — `permission` (fine-grained permission allowed), `fallback` (`augment.admin` fallback allowed), or `denied` (both denied)
```
PatAKnight: Completely up to you, but the RBAC plugin does already log this information: https://github.com/backstage/community-plugins/blob/main/workspaces/rbac/plugins/rbac-backend/src/auditor/auditor.ts#L65 and https://github.com/backstage/community-plugins/blob/main/workspaces/rbac/plugins/rbac-backend/docs/audit-log.md#audit-log-format. The only thing that you would be missing is the `grantedVia` section

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/specs/permission-definitions/spec.md
```diff
@@ -0,0 +1,169 @@
...
+| ------------------------- | ------------------------------ |
+| `augment.agent.list`      | read (basic, no resource type) |
+| `augment.agent.register`  | create                         |
```
PatAKnight: So, just want to call this out, but there are times where you might not need/want resource types for all of your permissions. Namely `create` permissions as technically the resource that you are creating might not exist yet and therefore a conditional policy (filter) might not work for it. This was an issue that the RBAC plugin run into not too long ago with our `policy.entity.create` permission. And it can create problems with things like the frontends ability to utilize `usePermission`.

Section from backstage on the topic with a note on the bottom mentioning this: https://backstage.io/docs/permissions/writing-a-policy#resource-types
`usePermission` from backstage: https://backstage.io/docs/permissions/plugin-authors/05-frontend-authorization#using-usepermission

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
```diff
@@ -0,0 +1,69 @@
...
+
+The `IS_NOT_CREATOR` permission rule enables RBAC-level self-approval prevention, but the existing hard-coded check (`createdBy !== userRef`) is retained. The hard-coded check is the primary guard; the permission rule allows deployers to customize or relax the policy via RBAC if desired.
+
+**Alternative considered:** Removing the hard-coded check entirely and relying solely on the permission rule. Rejected because a misconfigured RBAC policy could silently disable self-approval prevention — defense-in-depth is safer for a governance control.
+
+### Decision 4: Visibility filtering via basic permission
+
+`augment.agent.list` is a basic (non-resource) permission that controls visibility: ALLOW shows all agents, DENY filters to published + user's own agents. This avoids evaluating resource-based conditions against every agent in the list.
+
+**Alternative considered:** Resource-based permission evaluated per-agent in list responses. Rejected due to performance — evaluating conditional rules against potentially hundreds of agents per list request is expensive and unnecessary when the distinction is binary (admin vs. filtered view).
```
PatAKnight: I may be misunderstanding the intent here but I would think that you are roughly filtering anyways during the deny approach which (in theory) should perform similarly to the conditional access filtering that happens. Albiet, the difference is that you do not have to wait on the permission framework and RBAC plugins to return back the information needed.

But you could then have an allow (sees everything), deny (sees nothing), and conditional (which sees based on the filter that is assigned to the user). And these filters (conditional rules) would be the ones available to the resource type that you assign to the `augment.agent.list` permission.

Unless you are only ever planning on filtering for one single rule no matter what (I think in this case it was for ownership)

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/fine-grained-permissions-prelim-implementation-plan.md
```diff
@@ -0,0 +1,243 @@
...
+
+**`plugins/augment-backend/src/routes/toolLifecycleRoutes.ts`** — Same pattern for 4 tool lifecycle decisions.
+
+**`plugins/augment-backend/src/routes/kagentiAgentRoutes.ts`** — Replace `requireAdminAccess` on infra routes with `augment.kagenti.admin` + fallback.
+
+### Phase 5: Plugin Wiring (2 files)
+
+**`plugins/augment-backend/src/plugin.ts`** — Register all new permissions via `permissionsRegistry.addPermissions`. Add `@backstage/plugin-permission-node` dependency.
+
+**`plugins/augment-backend/src/router.ts`** — Create `permissionIntegrationRouter` via `createPermissionIntegrationRouter` with both resource types and all rules. Mount it on the router. Thread `authorizeLifecycleAction` and `authorizeBasicWithFallback` into `RouteContext`.
```
PatAKnight, line 153: I believe that this is no longer needed now with the preferred approach being to use the `PermissionsRegistryService` within `plugin.ts`

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/fine-grained-permissions-prelim-implementation-plan.md
```diff
@@ -0,0 +1,243 @@
...
+- PUT withdraw → `augment.agent.withdraw` with IS_OWNER condition
+- DELETE → `augment.agent.delete` with IS_OWNER + HAS_LIFECYCLE_STAGE conditions
+
+**`plugins/augment-backend/src/routes/toolLifecycleRoutes.ts`** — Same pattern for 4 tool lifecycle decisions.
+
+**`plugins/augment-backend/src/routes/kagentiAgentRoutes.ts`** — Replace `requireAdminAccess` on infra routes with `augment.kagenti.admin` + fallback.
+
+### Phase 5: Plugin Wiring (2 files)
+
+**`plugins/augment-backend/src/plugin.ts`** — Register all new permissions via `permissionsRegistry.addPermissions`. Add `@backstage/plugin-permission-node` dependency.
```
PatAKnight, line 151: I believe you will also need to include `permissionsRegistry.addResourceType` as well for each of your resource types in addition to `permissionsRegistry.addPermissions` for any of your basic permissions.

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
gabemontero: To your question @PatAKnight -  yes, the current design only ever filters by one hardcoded rule (ownership + published status). 

That said, I think you are on the right track in that could be overly simplistic, where different deployments might want different visibility semantics, 

Hence, your "3-tier" model (feel free to provide a different name if you don't think "3-tier" properly captures it :-) ) is the right call.

gabemontero:
thanks for the expert review @PatAKnight 

claude and I have attempted to make updates around all your comments .... pushed as a separate commit ... PTAL when you have the chance 

thanks again

workspaces/augment/openspec/changes/fine-grained-backstage-permissions/design.md
PatAKnight: Nah, love the "3-tier" name works best for what I was trying to expand

PatAKnight (approved):
LGTM

gabemontero:
> LGTM

awesome / thanks @PatAKnight

PR #3333, opened 2026-06-09
FLPATH-4249 | [DCM] Provider name field is editable in edit mode - possible identity mutation risk

asmasarw:
**Disable the Provider Name field when editing an existing provider.**

The Name field in the Edit Provider dialog is now read-only. Provider names are
immutable identifiers and should not be changed after creation.


y0n1 (approved):
LGTM in general.
- Adding a nice `ProviderFormFields.test.tsx` would be even better 😉.
- Should `helperText` remain as-is?

jkilzi (approved):

PR #3336, opened 2026-06-09
FLPATH-4272 | [DCM] Hardcoded colors may break dark mode

asmasarw:
Replace hardcoded colors and shadows with theme-aware values for dark mode support.

- ProviderStatus: chip border now uses `theme.palette.divider` instead of `rgba(0,0,0,0.2)`
- CopyButton: "copied" checkmark color now uses `theme.palette.success.main` instead of `#28a745`
- dcmStyles: overview card box-shadow is suppressed in dark mode via `theme.palette.type` check

### DIFF
--- a/workspaces/dcm/plugins/dcm/src/components/dcmStyles.ts
+++ b/workspaces/dcm/plugins/dcm/src/components/dcmStyles.ts
@@ -300,7 +300,10 @@ export const useDcmStyles = makeStyles(theme => ({
     '& .MuiPaper-root': {
       border: `1px solid ${theme.palette.divider}`,
       borderRadius: theme.shape.borderRadius,
-      boxShadow: '0 1px 2px rgba(0, 0, 0, 0.06)',
+      boxShadow:
+        theme.palette.type === 'dark'
+          ? 'none'
+          : '0 1px 2px rgba(0, 0, 0, 0.06)',
       backgroundColor: theme.palette.background.paper,
     },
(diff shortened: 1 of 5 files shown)


workspaces/dcm/plugins/dcm/src/components/dcmStyles.ts
```diff
@@ -300,7 +300,10 @@ export const useDcmStyles = makeStyles(theme => ({
     '& .MuiPaper-root': {
       border: `1px solid ${theme.palette.divider}`,
       borderRadius: theme.shape.borderRadius,
-      boxShadow: '0 1px 2px rgba(0, 0, 0, 0.06)',
+      boxShadow:
+        theme.palette.type === 'dark'
```
mareklibra: In a previous PR, a `isDarkMode()` function has been added. It would be good to reuse it here instead

mareklibra (changes requested):
Besides of the comment about isDarkMode(), it would be good to find other occurances difficult for the dark mode. 

Hint: CatalogItemFormFields.tsx (background) 

But there might be others

workspaces/dcm/plugins/dcm/src/components/dcmStyles.ts
asmasarw: Nice Catch :)

mareklibra (approved):

PR #3339, opened 2026-06-09
Policy Field: Add MaxLen, Required Priority and Fix Decimal issues

asmasarw:
### Tickets:

- FLPATH-4270 | [DCM] Policy description field has no max-length validation
- FLPATH-4269 | [DCM] Policy priority field accepts non-integer values
- FLPATH-4271 | [DCM] Policy priority field not marked as required
- FLPATH-4244 | [DCM] Normalize policy enabled checks for defensive coding
- FLPATH-4255 | [DCM] Policy priority uniqueness constraint not communicated
- FLPATH-4250 | [DCM] Whitespace-only values pass form validation for required policy fields

---

### Changes:

Fix policy form validation for Description and Priority fields, normalize enabled
checks, and reject whitespace-only required fields.

- Add a 255-character max-length Yup rule to the Description field, with inline error
  feedback matching the Display Name pattern (validation error on blur rather than
  hard-blocking input at the DOM level).
- Add `.integer()` and `.required()` Yup rules to the Priority field so decimal values
  (e.g. 500.5) and an empty field are rejected with an explicit error message instead of
  silently coercing to an integer or defaulting to 500.
- Add `step={1}` to the Priority number input and block `.`/`e`/`E` characters via
  `onKeyDown` and `onChange` guards to prevent decimal strings from bypassing the Yup
  integer check through JavaScript's `Number()` coercion.
- Update the Priority label to "Priority \*" to indicate it is a required field.
- Update the Priority helper text to surface the API uniqueness constraint
  (priority must be unique per policy type) so users are informed before hitting a 409.
- Normalize the three inconsistent `enabled` checks in `PoliciesTabContent` to
  `p.enabled ?? true` so the Enabled column, Actions toggle, and edit form all agree
  when `enabled` is undefined.
- Add `.trim()` to the `display_name` and `rego_code` Yup rules so whitespace-only
  values are rejected client-side instead of passing validation and being silently
  rejected by the API after submit.

### DIFF
--- a/workspaces/dcm/plugins/dcm/src/pages/policies/PoliciesTabContent.tsx
+++ b/workspaces/dcm/plugins/dcm/src/pages/policies/PoliciesTabContent.tsx
@@ -119,7 +119,7 @@ export function PoliciesTabContent() {
       setTogglingIds(addTogglingId(id));
       try {
         const updated = await policyApi.updatePolicy(id, {
-          enabled: !p.enabled,
+          enabled: !(p.enabled ?? true),
         });
         setItems(replacePolicyById(id, updated));
       } catch (err) {
--- a/workspaces/dcm/plugins/dcm/src/pages/policies/PoliciesTabContent.tsx
+++ b/workspaces/dcm/plugins/dcm/src/pages/policies/PoliciesTabContent.tsx
@@ -172,7 +172,7 @@ export function PoliciesTabContent() {
         title: 'Enabled',
         field: 'enabled',
         render: p => {
-          const isEnabled = p.enabled === true;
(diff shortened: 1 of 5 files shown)


workspaces/dcm/plugins/dcm/src/pages/policies/PoliciesTabContent.tsx
mareklibra: when p.enabled is undefined, !p.enabled is true, so “disable” still sends enabled: true. 

```suggestion
enabled: !(p.enabled ?? true),
```

mareklibra (approved):

PR #3342, opened 2026-06-09
FLPATH-4262 | [DCM] Provider health status is static with no refresh or polling

asmasarw:
### Tasks:

FLPATH-4262 | [DCM] Provider health status is static with no refresh or polling

---

### Changes:

Add a manual refresh button to the Providers table to update health status without a full page reload.

A sync icon button now appears next to the search field in the Providers card header. Clicking it re-fetches the provider list (including `health_status`) while keeping the table visible. A spinner is shown on the button during the request. The initial page load behaviour is unchanged.

### DIFF
--- a/workspaces/dcm/plugins/dcm/src/hooks/useCrudTab.ts
+++ b/workspaces/dcm/plugins/dcm/src/hooks/useCrudTab.ts
@@ -99,6 +99,8 @@ export interface UseCrudTabResult<T, F extends Record<string, unknown>> {
   items: T[];
   setItems: React.Dispatch<React.SetStateAction<T[]>>;
   loading: boolean;
+  /** True only during a manual reload after the first successful load. The table stays visible. */
+  refreshing: boolean;
   loadError: string | null;
   reload: () => void;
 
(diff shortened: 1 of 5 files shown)


jkilzi (commented):
👀

workspaces/dcm/plugins/dcm/src/hooks/useCrudTab.ts
```diff
@@ -232,16 +236,26 @@ export function useCrudTab<T, F extends Record<string, unknown>>(
...
+      .then(result => {
+        setItems(result);
+        hasLoadedRef.current = true;
+      })
       .catch(err => {
         setLoadError(extractApiError(err));
         setItems([]);
       })
-      .finally(() => setLoading(false));
+      .finally(() => {
```
jkilzi, line 256: `hasLoadedRef` is never reset

jkilzi (changes requested):
After a failed refresh, Retry shows the empty state instead of a spinner because hasLoadedRef stays true while items is empty. Consider using loading when items.length === 0, or preserving items on refresh failure. Also worth a unit test asserting refreshing === true during manual reload.

jkilzi (approved):

PR #3347, opened 2026-06-09
feat(lightspeed): add DCR authentication support for MCP servers

maysunfaisal:

  - See screenshots below 

## Issue
https://redhat.atlassian.net/browse/RHIDP-11657

## Summary

Adds Dynamic Client Registration (DCR) authentication for MCP servers in Lightspeed, allowing the backend to mint per-user plugin request tokens instead of requiring static tokens.

### What changed

**Backend (`lightspeed-backend`)**
- New `auth: dcr` config option for MCP servers — when set, the backend mints a Backstage service token on behalf of the user via `getPluginRequestToken`
- Added `AuthService` injection into the router
- `GET /mcp-servers` and `PATCH /mcp-servers/:name` responses now include the `auth` field
- DCR servers report `hasToken: true` always (tokens are minted per-request, no manual entry needed)
- `buildMcpHeaders` sends the minted token in `MCP-HEADERS` for DCR servers

**Frontend (`lightspeed`)**
- MCP Settings modal shows a read-only message for DCR servers explaining tokens are automatic
- New translation strings for DCR status/description

**Bug fixes along the way**
- Fixed kebab menu not auto-closing after clicking "MCP Settings" — caused by duplicate `@patternfly/react-core` versions (added resolution to force `6.4.3`)
- Removed stale `attachButtonPosition` prop that no longer exists in `@patternfly/chatbot`
- Fixed RBAC backend v7+ requiring `@backstage/plugin-permission-backend` registered separately (breaking change from v5/v6)

**Other**
- Added `@backstage/plugin-auth` to frontend for DCR OAuth2 consent page
- Added `dist-dynamic` to `.prettierignore`
- Reference `rbac-policy.csv` with Lightspeed permissions for local RBAC testing
- Updated API reports

### How to test

1. **Unit tests** — `cd workspaces/lightspeed && yarn test:all` (769 tests pass)
2. **Local DCR testing** — Requires:
   - `@backstage/plugin-mcp-actions-backend` added to `index.ts`
   - `experimentalDynamicClientRegistration.enabled: true` in `app-config.yaml`
   - `mcpServers: [{ name: "...", auth: dcr }]` in lightspeed config
   - An MCP client (Cursor/VS Code) connecting to `http://localhost:7007/api/mcp-actions/v1` — DCR OAuth consent should pop up in browser
3. **OpenShift** — Deploy custom OCI images with DCR changes; set `auth: dcr` in `values.yaml` under `lightspeed.mcpServers`

### Config example

```yaml
lightspeed:
  mcpServers:
    - name: mcp-integration-tools
      auth: dcr          # tokens minted automatically
    - name: legacy-server
      token: ${TOKEN}    # static token (existing behavior)
```


gabemontero (commented):
hey @maysunfaisal - so claude and I looked at two error scenarios:

1) If the user's RBAC policy denies a specific tool invocation

for this, claude's analysis was that the rejection happens inside @backstage/plugin-mcp-actions-backend, not in this code. Lightspeed streams back whatever LCS returns — so the user would see whatever error or refusal the LCS/MCP Actions layer produces in the chat stream. The Lightspeed backend is fully opaque to per-tool authorization outcomes.

Assuming this is correct, I'm curious if it makes sense to catch the LCS/MCP actions layer error and display a potentially more user friendly indication of permission issues.

WDYT?

2) Unexpected error minting a Backstage token (`getPluginRequestToken` throws)

Claude claims the entire chat query fails (I'll post the details in a moment).  It then goes on that is a chat query results in querying multiple MCP servers, none of those servers gets called.

I suppose a user prompt could result in multiple queries, but even if so, it seems unpredictable whether calling the other mcp servers is productive i.e. how dependent would the output of the backstage mcp server be.

I'm inclined to leave this be, but thought I would surface it in the review as due diligence to get your and everyone else's take.

Here is the code analysis from claude on `gitPluginRequestToken`

```
  There are two call sites for getPluginRequestToken, and neither has a local try/catch around it:

  A) During /v1/query (chat flow) — buildMcpHeaders() calls getPluginRequestToken at ~line 133. If it throws, the exception propagates unhandled out of buildMcpHeaders and is caught by the outer catch in the /v1/query handler (~line 775):

  } catch (error) {
      const errormsg = `Error fetching completions from ${provider}: ${error}`;
      logger.error(errormsg);
      response.status(500).json({ error: errormsg });
  }

  The entire chat query fails with a 500. Even if only one of several MCP servers uses auth: dcr, a token-minting failure for that one server aborts the whole request — the error is not isolated per-server. The other MCP servers (static-token or other DCR servers) never get their headers built.

  B) During POST /mcp-servers/:name/validate (~line 393) — same pattern: no local catch, the outer handler returns 500 with "Validation failed".

  Note: there is a guard for when authService/credentials are absent (logs a warning, skips the server), but that only handles the case where the services weren't injected — not a runtime failure in getPluginRequestToken itself.

```

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -83,12 +119,31 @@ async function buildMcpHeaders(
...
-      headers[server.name] = { Authorization: `${token}` };
+    if (server.auth === 'dcr') {
+      if (server.token && options?.logger) {
+        options.logger.warn(
+          `MCP server '${server.name}' has both auth: dcr and a static token; ` +
+            `the static token is ignored when auth is dcr`,
+        );
+      }
+      if (options?.authService && options?.credentials) {
+        const { token } = await options.authService.getPluginRequestToken({
```
yangcao77: claude suggests:
getPluginRequestToken() is called inside a for loop with no try-catch. If token minting fails for one DCR
  server (e.g., mcp-actions plugin not registered, auth service down), the entire function throws. This
  prevents MCP headers from being built for any server — including healthy static-token ones. A single
  misconfigured DCR server takes down all MCP integration for every user.                  

  Fix: Wrap the getPluginRequestToken call in a per-server try-catch, log the error, and skip that server
  while continuing to process others.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -83,12 +119,31 @@ async function buildMcpHeaders(
...
+          onBehalfOf: options.credentials,
+          targetPluginId: 'mcp-actions',
+        });
+        headers[server.name] = { Authorization: `${token}` };
+      } else {
+        options?.logger?.warn(
```
yangcao77, line 137: this is only a logger in backend, frontend won't receive this warning. so if the auth is not available, should this be sent to & shown in UI as well?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/mcp-server-types.ts
```diff
@@ -44,6 +44,8 @@ export interface McpServerResponse {
   toolCount: number;
   hasToken: boolean;
   hasUserToken: boolean;
+  /** Authentication mode — `'dcr'` means tokens are minted automatically. */
+  auth?: string;
```
yangcao77: you defined `type McpServerAuth = 'dcr';` in router, should this be `McpServerAuth`?

yangcao77: same for other `auth: string` in this PR

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -56,23 +61,54 @@ import { QueryRequestBody, RouterOptions } from './types';
...
+ *
+ * Servers the user has disabled are always excluded.
  */
 async function buildMcpHeaders(
   servers: StaticMcpServer[],
   store: McpUserSettingsStore,
   userEntityRef: string,
+  options?: {
+    authService?: AuthService;
+    credentials?: BackstageCredentials;
```
yangcao77: from claude:
These two fields are always needed together (both required for getPluginRequestToken), but the type allows providing one without the other. Bundling them as `dcrAuth?: { authService, credentials }` would make partial-provision a compile-time error.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -83,12 +119,31 @@ async function buildMcpHeaders(
...
 
-    // User's personal token (DB) takes precedence over admin default (app-config).
-    // If the user hasn't set one, falls back to the config token.
-    // If neither exists, the server is excluded from MCP-HEADERS.
-    const token = setting?.token || server.token;
-    if (token) {
-      headers[server.name] = { Authorization: `${token}` };
+    if (server.auth === 'dcr') {
+      if (server.token && options?.logger) {
+        options.logger.warn(
```
yangcao77: from claude:

When a server has both auth: dcr and a static token, the warning fires on every chat query (per-request). Asimilar warning already fires at startup during config parsing (lines 212-216). The per-request duplicate is log noise.
  
Fix: Remove the per-request warning (rely on the startup warning), or gate it behind a warnedOnce set.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -321,8 +397,17 @@ export async function createRouter(
           return;
         }
 
-        const setting = await settingsStore.get(name, userEntityRef);
-        const effectiveToken = setting?.token || server.token;
+        let effectiveToken: string | undefined;
+        if (server.auth === 'dcr') {
+          const { token: dcrToken } = await auth.getPluginRequestToken({
```
yangcao77: from claude:

f getPluginRequestToken throws, the outer catch returns { error: 'Validation failed' }. The user sees "Validation failed" and thinks the MCP server is broken, when the actual problem is Backstage auth misconfiguration. A specific catch around the minting call should return a distinct error message.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Good catch. This scenario would affect all DCR servers equally and isn't something an end user can act on (PE configure MCP Servers). Surfacing it in the UI would require either adding a status field to the /mcp-servers response or injecting a warning into the chat stream, both of which touch more surface area than I'd like in this PR. For now the backend log warning gives the warning. I can capture showing auth availability in the MCP Settings modal as a follow-up enhancement

maysunfaisal:
> hey @maysunfaisal - so claude and I looked at two error scenarios:
> 
> 1. If the user's RBAC policy denies a specific tool invocation
> 
> for this, claude's analysis was that the rejection happens inside @backstage/plugin-mcp-actions-backend, not in this code. Lightspeed streams back whatever LCS returns — so the user would see whatever error or refusal the LCS/MCP Actions layer produces in the chat stream. The Lightspeed backend is fully opaque to per-tool authorization outcomes.
> 
> Assuming this is correct, I'm curious if it makes sense to catch the LCS/MCP actions layer error and display a potentially more user friendly indication of permission issues.
> 
> WDYT?
> 
> 2. Unexpected error minting a Backstage token (`getPluginRequestToken` throws)
> 
> Claude claims the entire chat query fails (I'll post the details in a moment). It then goes on that is a chat query results in querying multiple MCP servers, none of those servers gets called.
> 
> I suppose a user prompt could result in multiple queries, but even if so, it seems unpredictable whether calling the other mcp servers is productive i.e. how dependent would the output of the backstage mcp server be.
> 
> I'm inclined to leave this be, but thought I would surface it in the review as due diligence to get your and everyone else's take.
> 
> Here is the code analysis from claude on `gitPluginRequestToken`
> 
> ```
>   There are two call sites for getPluginRequestToken, and neither has a local try/catch around it:
> 
>   A) During /v1/query (chat flow) — buildMcpHeaders() calls getPluginRequestToken at ~line 133. If it throws, the exception propagates unhandled out of buildMcpHeaders and is caught by the outer catch in the /v1/query handler (~line 775):
> 
>   } catch (error) {
>       const errormsg = `Error fetching completions from ${provider}: ${error}`;
>       logger.error(errormsg);
>       response.status(500).json({ error: errormsg });
>   }
> 
>   The entire chat query fails with a 500. Even if only one of several MCP servers uses auth: dcr, a token-minting failure for that one server aborts the whole request — the error is not isolated per-server. The other MCP servers (static-token or other DCR servers) never get their headers built.
> 
>   B) During POST /mcp-servers/:name/validate (~line 393) — same pattern: no local catch, the outer handler returns 500 with "Validation failed".
> 
>   Note: there is a guard for when authService/credentials are absent (logs a warning, skips the server), but that only handles the case where the services weren't injected — not a runtime failure in getPluginRequestToken itself.
> ```

@gabemontero Thanks for flagging this. You're correct, RBAC tool-level denials happen inside @backstage/plugin-mcp-actions-backend, and Lightspeed is opaque to those outcomes; it streams back whatever LCS returns. Surfacing a user-friendly permission error would require parsing the LCS response stream for authorization-specific error patterns and rendering them differently in the chat UI and that touches the frontend streaming logic and response parsing, which is a bigger scope than this PR. I'll capture it as a follow-up enhancement (parsing MCP/RBAC errors from LCS and rendering actionable messages in the Lightspeed chat). For now, the user will see whatever LCS surfaces in the streamed response.

Addressed your second concern

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Updated

workspaces/lightspeed/plugins/lightspeed-backend/src/service/mcp-server-types.ts
maysunfaisal: Updated

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Updated

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Updated

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Updated

gabemontero:
> 
> @gabemontero Thanks for flagging this. You're correct, RBAC tool-level denials happen inside @backstage/plugin-mcp-actions-backend, and Lightspeed is opaque to those outcomes; it streams back whatever LCS returns. Surfacing a user-friendly permission error would require parsing the LCS response stream for authorization-specific error patterns and rendering them differently in the chat UI and that touches the frontend streaming logic and response parsing, which is a bigger scope than this PR. I'll capture it as a follow-up enhancement (parsing MCP/RBAC errors from LCS and rendering actionable messages in the Lightspeed chat). For now, the user will see whatever LCS surfaces in the streamed response.
> 
> Addressed your second concern

sounds good @maysunfaisal thanks

maysunfaisal:
@gabemontero Here is the issue to address surfacing user friendly MCP permission denial msgs https://redhat.atlassian.net/browse/RHIDP-14907

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: Opened issue https://redhat.atlassian.net/browse/RHIDP-14908

maysunfaisal:
# Screenshots - Local

## RBAC Example - catalog.example.read = deny

### rbac policy file
Screenshot 2026-06-08 at 5 08 32 PM

### Lightspeed (prior to rebrand) interacting with Backstage/RHDH MCP with DCR - does not return Catalog resources
Screenshot 2026-06-08 at 5 06 47 PM
Screenshot 2026-06-08 at 5 08 21 PM


### MCP Settings for DCR auth server does not let you edit token from the UI
Screenshot 2026-06-08 at 5 06 33 PM

### Cursor MCP Settings - backstage-actions pointing to Backstage/RHDH MCP with DCR
Screenshot 2026-06-08 at 5 11 01 PM

### Redirected to Backstage/RHDH for DCR auth and successful auth
Screenshot 2026-06-08 at 5 10 48 PM

### Cursor MCP backstage-actions now shows a successful connection
Screenshot 2026-06-08 at 5 11 17 PM
Screenshot 2026-06-08 at 5 11 11 PM

### Query the Backstage/RHDH Catalog, denied due to rbac policy
Screenshot 2026-06-08 at 5 12 19 PM

## RBAC Example - catalog.example.read = allow

### rbac policy file
Screenshot 2026-06-08 at 5 13 38 PM

### Lightspeed query now passes for the same prompt using MCP with DCR
Screenshot 2026-06-08 at 5 13 44 PM

### Cursor query now passes for the same prompt using MCP with DCR
Screenshot 2026-06-08 at 5 14 27 PM

maysunfaisal:
# Screenshots - OpenShift

## RBAC Example - catalog.example.read = allow

### Lightspeed (prior to rebrand) MCP Settings does not let you edit for MCP Server with DCR
Screenshot 2026-06-09 at 3 26 56 PM

### Lightspeed successfully returns Catalog resources for MCP with DCR
Screenshot 2026-06-09 at 3 30 37 PM

### Cursor DCR auth with MCP from OpenShift
Screenshot 2026-06-09 at 3 28 25 PM
Screenshot 2026-06-09 at 3 28 08 PM

### Cursor successfully returns catalog resources for Backstage/RHDH MCP with DCR
Screenshot 2026-06-09 at 3 29 28 PM

## RBAC Example - catalog.example.read = deny

### Updated rbac policy config map on OpenShift for deny
Screenshot 2026-06-09 at 3 45 55 PM

### Lightspeed does not return catalog entities for the same prompt
Screenshot 2026-06-09 at 3 45 41 PM

### Cursor does not return catalog entities for the same prompt
Screenshot 2026-06-09 at 3 45 36 PM

workspaces/lightspeed/plugins/lightspeed-backend/src/plugin.ts
```diff
@@ -45,12 +46,15 @@ export const lightspeedPlugin = createBackendPlugin({
...
         http,
         httpAuth,
+        auth,
         userInfo,
         permissions,
         database,
       }) {
         await migrate(database);
 
+        await migrate(database);
```
yangcao77: duplicate?

workspaces/lightspeed/plugins/lightspeed/src/translations/ref.ts
```diff
@@ -346,6 +346,9 @@ export const lightspeedMessages = {
   'mcp.settings.status.oneTool': '{{count}} tool',
   'mcp.settings.status.manyTools': '{{count}} tools',
   'mcp.settings.status.unknown': 'Unknown',
+  'mcp.settings.status.autoManaged': 'Authenticated via Backstage',
```
yangcao77: is this ever referenced anywhere?

yangcao77 (changes requested):
generally looks good to me. my previous comments have been resolved. 
please remove the duplicate migrate call.  

others are some minor comments., and I'm open to merge this PR without those changes.

workspaces/lightspeed/plugins/lightspeed-backend/src/plugin.ts
maysunfaisal: fixed

workspaces/lightspeed/plugins/lightspeed/src/translations/ref.ts
maysunfaisal: fixed

yangcao77 (approved):
generally lgtm

maysunfaisal:
@redhat-developer/rhdh-ui @redhat-developer/rhdh-plugins-maintainers Could you PTAL at this PR? Thanks!

HusneShabbir (commented):
cc: @ciiay

HusneShabbir (commented):
https://github.com/maysunfaisal/rhdh-plugins/pull/1 covers the E2E tests for the MCP–DCR integration. One observation from the testing is that the new DCR alert is not fully localized yet.

Image

workspaces/lightspeed/plugins/lightspeed/src/translations/ref.ts
```diff
@@ -345,6 +345,8 @@ export const lightspeedMessages = {
   'mcp.settings.status.oneTool': '{{count}} tool',
   'mcp.settings.status.manyTools': '{{count}} tools',
   'mcp.settings.status.unknown': 'Unknown',
+  'mcp.settings.modalDescriptionDcr':
+    'This server uses Dynamic Client Registration (DCR). Tokens are minted automatically using your Backstage identity — no manual token is needed.',
   'mcp.settings.toggleServerAriaLabel': 'Toggle {{serverName}}',
```
its-mitesh-kumar, line 350: The new mcp.settings.modalDescriptionDcr key is only added to the English reference file (ref.ts). Please add the corresponding translated strings in all other supported language files as well:

- de.ts (German)
- es.ts (Spanish)
- fr.ts (French)
- it.ts (Italian)
- ja.ts (Japanese)

its-mitesh-kumar (changes requested):

workspaces/lightspeed/plugins/lightspeed/src/translations/ref.ts
maysunfaisal: @its-mitesh-kumar how would one acquire the translation for the other languages? Is there a process?

its-mitesh-kumar: Please do add these strings in other language(can be generated through claude/cursor). Later UI team sent it to TMS(translation team) for verification. And if required we do update it.
| Language | File | Translated String |
|----------|------|-------------------|
| German | `de.ts` | `Dieser Server verwendet Dynamic Client Registration (DCR). Token werden automatisch mit Ihrer Backstage-Identität erstellt — kein manuelles Token erforderlich.` |
| Spanish | `es.ts` | `Este servidor utiliza Dynamic Client Registration (DCR). Los tokens se generan automáticamente usando tu identidad de Backstage — no se necesita un token manual.` |
| French | `fr.ts` | `Ce serveur utilise Dynamic Client Registration (DCR). Les jetons sont générés automatiquement à partir de votre identité Backstage — aucun jeton manuel n'est nécessaire.` |
| Italian | `it.ts` | `Questo server utilizza Dynamic Client Registration (DCR). I token vengono generati automaticamente utilizzando la tua identità Backstage — non è necessario alcun token manuale.` |
| Japanese | `ja.ts` | `このサーバーは Dynamic Client Registration (DCR) を使用しています。トークンはあなたの Backstage ID を使用して自動的に発行されるため、手動でのトークン入力は不要です。` |

maysunfaisal: @its-mitesh-kumar thanks for the explanation!

HusneShabbir (commented):
/lgtm

maysunfaisal:
> ## Review
> ### Findings
> #### Medium
> * **[api-contract]** `workspaces/lightspeed/plugins/lightspeed-backend/src/plugin.ts` — `RouterOptions.auth` is now a required field (`auth: AuthService`), which is a breaking change for any external consumer of `createRouter()` that does not pass it. Within the Backstage plugin system, `coreServices.auth` is always injected by the framework, so internal usage is safe. However, JavaScript consumers bypassing TypeScript checks could hit a runtime TypeError. The changeset declares a minor version bump, which may be insufficient for a breaking API change.
>   Remediation: Consider making `auth` optional (`auth?: AuthService`) with a runtime guard, or document as a breaking change.

This is an internal plugin — coreServices.auth is always injected by the Backstage plugin framework. Making it optional would add unnecessary null checks for a code path that can't occur in practice.

> * **[test-inadequate]** `workspaces/lightspeed/plugins/lightspeed-backend/src/service/mcp-server.test.ts` — The new DCR tests cover `GET /mcp-servers` (auth field presence) and the `/v1/query` flow (token minting and mixed DCR/static headers). However, two code paths remain untested: (1) `POST /mcp-servers/:name/validate` with a DCR-configured server, and (2) the `PATCH /mcp-servers/:name` endpoint's 502 error path when `auth.getPluginRequestToken` throws.
>   Remediation: Add tests for the validate endpoint's DCR path and the PATCH 502 error path.

Partially addressed. Added test for the DCR validate endpoint path. The 502 error case (auth service throws) requires overriding coreServices.auth mock internals which is non-trivial with the current test infrastructure.

> 
> #### Low
> * **[edge-case]** `workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts` — In `buildMcpHeaders`, when a DCR server is configured but `dcrAuth` is unavailable, the function logs a warning and silently skips the server. The `/v1/query` request proceeds without that server's authorization header, and LCORE will likely return a 401. In practice, this path only triggers if `createRouter` is used by an external caller without the auth service.

Intentional design — graceful degradation. The warning log is sufficient; crashing the entire request when one server's auth is unavailable would be worse UX.

> * **[api-contract]** `workspaces/lightspeed/plugins/lightspeed-backend/config.d.ts` — The `auth` field is typed as `string` rather than the literal type `'dcr'`. While runtime validation exists (unsupported values trigger a warning and fall back to static-token mode), the schema could be tightened for better developer experience.

Addressed. Changed type from string to 'dcr'. 

> * **[data-exposure]** `workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts` — The `auth` field is returned in the `GET /mcp-servers` API response, but `config.d.ts` marks it with `@visibility backend`. The field value is non-sensitive (just the string `'dcr'` or undefined) and is needed by the frontend to render the correct modal UI. The `@visibility` annotation should be updated to `frontend` for consistency.

Addressed. Changed @visibility backend to @visibility frontend.

> * **[authorization]** `workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts` — `targetPluginId` is hardcoded to `'mcp-actions'` for all DCR servers. This is by design — DCR mode targets Backstage-internal MCP servers that all route through the mcp-actions backend plugin, which enforces per-resource permissions via RBAC. A documentation note clarifying this design decision would be helpful.

By design — all DCR-enabled MCP servers in RHDH are proxied through @backstage/plugin-mcp-actions-backend, so the target is always mcp-actions.

> * **[input-validation]** `workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts` — The `auth` config value is checked via strict equality (`authValue === 'dcr'`). A typo like `'DCR'` or `'dcr '` falls back to static-token mode with a warning log, which is reasonable. Case-insensitive comparison would be a minor UX improvement.
> 

Addressed. Added .toLowerCase() so DCR, Dcr, etc. all work.

> **Labels:** PR adds a new DCR authentication feature for MCP servers with security-relevant auth changes and test additions.

workspaces/lightspeed/plugins/lightspeed-backend/src/plugin.ts
```diff
@@ -103,6 +105,7 @@ export const lightspeedPlugin = createBackendPlugin({
             logger,
             database,
             httpAuth,
```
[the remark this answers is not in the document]
maysunfaisal: Not changing — coreServices.auth is always injected by the Backstage plugin framework. Making it optional would add null checks for a code path that can't occur.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/mcp-server.test.ts
```diff
@@ -257,6 +273,28 @@ describe('MCP server management endpoints', () => {
```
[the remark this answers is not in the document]
maysunfaisal: Added test for the DCR validate endpoint path. The 502 error case requires overriding coreServices.auth mock internals which is non-trivial with the current test infrastructure.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -85,12 +119,32 @@ async function buildMcpHeaders(
...
+            await options.dcrAuth.authService.getPluginRequestToken({
+              onBehalfOf: options.dcrAuth.credentials,
+              targetPluginId: 'mcp-actions',
+            });
+          headers[server.name] = { Authorization: `${token}` };
+        } catch (err) {
+          options?.logger?.error(
+            `Failed to mint DCR token for MCP server '${server.name}': ${err}`,
+          );
+        }
```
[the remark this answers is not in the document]
maysunfaisal: Intentional — graceful degradation. The warning log is sufficient; crashing the request when one server's auth is unavailable would be worse UX.

workspaces/lightspeed/plugins/lightspeed-backend/config.d.ts
```diff
@@ -48,6 +48,13 @@ export interface Config {
        * @visibility secret
        */
       token?: string;
+      /**
+       * Authentication mode for this MCP server.
+       * Set to 'dcr' for Dynamic Client Registration (user-bound tokens minted per-request).
```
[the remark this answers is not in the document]
maysunfaisal: Fixed — tightened to literal type 'dcr'.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -262,8 +334,11 @@ export async function createRouter(
             enabled: setting ? Boolean(setting.enabled) : true,
             status: setting?.status ?? 'unknown',
             toolCount: setting?.tool_count ?? 0,
-            hasToken: !!(setting?.token || server.token),
+            // DCR servers always have a token (minted per-request); static servers check DB + config.
+            hasToken:
+              server.auth === 'dcr' || !!(setting?.token || server.token),
             hasUserToken: !!setting?.token,
```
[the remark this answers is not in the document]
maysunfaisal: Fixed — updated to @visibility frontend.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -85,12 +119,32 @@ async function buildMcpHeaders(
...
-    const token = setting?.token || server.token;
-    if (token) {
-      headers[server.name] = { Authorization: `${token}` };
+    if (server.auth === 'dcr') {
+      if (options?.dcrAuth) {
+        try {
+          const { token } =
+            await options.dcrAuth.authService.getPluginRequestToken({
+              onBehalfOf: options.dcrAuth.credentials,
+              targetPluginId: 'mcp-actions',
```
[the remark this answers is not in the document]
maysunfaisal: By design — all DCR MCP servers in RHDH are proxied through @backstage/plugin-mcp-actions-backend, so the target is always mcp-actions.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -138,17 +193,34 @@ export async function createRouter(
...
+  //
   // URLs come from LCS (GET /v1/mcp-servers), not from app-config.
   const mcpServersConfig = config.getOptionalConfigArray(
     'intelligent-assistant.mcpServers',
   );
   const staticServers: StaticMcpServer[] = [];
   if (mcpServersConfig) {
     for (const mcpServer of mcpServersConfig) {
+      const authValue = mcpServer.getOptionalString('auth');
+      if (authValue && authValue !== 'dcr') {
```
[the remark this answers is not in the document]
maysunfaisal: Fixed — auth value is now lowercased before comparison.

maysunfaisal:
@its-mitesh-kumar Addressed the fullsend reviews

HusneShabbir (commented):
/lgtm

its-mitesh-kumar (approved):
/lgtm based on above reviewer has verified the functionality.

workspaces/lightspeed/plugins/lightspeed/src/components/LightSpeedChat.tsx
```diff
@@ -1768,7 +1768,6 @@ export const LightspeedChat = ({
           onSendMessage={sendMessage}
           isSendButtonDisabled={isSendButtonDisabled}
           hasAttachButton
-          attachButtonPosition="start"
           handleAttach={handleAttach}
```
its-mitesh-kumar, line 1771: @maysunfaisal Why did we removed it ?
@HusneShabbir We did missed while testing, due to this change attach button start appearing in the right side of message bar, while it should be on the left of the model selctor.
Lets be careful.

Image

I will be fixing this as part of my new PR

PR #3349, opened 2026-06-10
fix(scorecard): restrict entities to components in filecheck module

djanickova:

The filecheck scorecard module currently pulls metrics for all entities that contain the `backstage.io/source-location `annotation. Since this annotation is automatically added by backstage, filecheck pulls metric even for entities where it is not needed, eg. entities of type Package, and therefore does a lot of requests and pulls a lot of unnecessary data. While the change in this PR is not the final solution for this and further work will need to be done to check the performance issues of the filecheck module (Jira ticket created [here](https://redhat.atlassian.net/browse/RHDHBUGS-3334)), it restrics the entity type being pulled to Components only. This will help pull less data and especially unblock the failing e2e tests. 

Related to: [RHDHBUGS-3197](https://redhat.atlassian.net/browse/RHDHBUGS-3197)

### DIFF
--- a/workspaces/scorecard/.changeset/dirty-symbols-mate.md
+++ b/workspaces/scorecard/.changeset/dirty-symbols-mate.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-filecheck': patch
+---
+
+Restrict filecheck metric collection to Component catalog entities by adding a kind: component catalog filter.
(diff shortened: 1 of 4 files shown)


workspaces/scorecard/.changeset/dirty-symbols-mate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-filecheck': patch
+---
+
+Restrict filecheck metric collection to Component catalog entities by adding a kind: component catalog filter.
```
dzemanov:
```suggestion
**BREAKING**: Restrict filecheck metric collection to Component catalog entities by adding a kind: component catalog filter.
```

dzemanov (changes requested):
Verified that from now filecheck is calculated only for components, thank you for the changes!

dzemanov (approved):

PR #3359, opened 2026-06-10
feat: Add input size validation for queries and attachments 

rajin-kichannagari:
Implement backend validation to prevent users from submitting arbitrarily large queries or attachments that could consume backend memory and LCS resources.

## Changes:
- Add size limit constants: MAX_QUERY_LENGTH (32K), MAX_ATTACHMENT_SIZE_BYTES (20MB), MAX_TOTAL_ATTACHMENTS_SIZE_BYTES (50MB)
- Extend validateCompletionsRequest middleware with query length and attachment size validation
- Add attachments field to QueryRequestBody interface
- Add query length validation to notebooks query endpoint
- Add comprehensive test coverage for all validation scenarios

## Checklist

Note: Changeset not applicable for security fix. Frontend changes may be needed in a follow-up issue to communicate limits to users.


Jdubrick (changes requested):
I believe we will need to update the `express.json(..)` portion when defining the routers to override the allowable body limit, or else it will default to whatever Express' default is. Adding an integration test to test this might be beneficial too, since right now we just unit test the validation itself

Notebook: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts#L70
Lightspeed: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts#L108

Jdubrick:
I also want to ping @karthikjeeyar to verify if there will need to be any follow up UI changes to display to users when their attachments / body was too large?

HusneShabbir (commented):
cc: @its-mitesh-kumar

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
```diff
@@ -67,7 +68,8 @@ export async function createNotebooksRouter(
 ): Promise<Router> {
   const { logger, config, httpAuth, userInfo, permissions } = options;
   const notebooksRouter = Router();
-  notebooksRouter.use(express.json());
+  // Set explicit body size limit for query endpoint requests
+  notebooksRouter.use(express.json({ limit: '60mb' }));
```
Jdubrick: I think I was confusing earlier on, if we apply this `60mb` to the whole router then even routes that don't need this kind of size can allow it. We should override the limit on the applicable routes, like query routes, etc that would need them.

You can add it to the route itself, like:

```
 router.post(
    '/v1/query',
    express.json({ limit: '60mb' }),
    validateCompletionsRequest,
    requirePermission(lightspeedChatCreatePermission),
    async (request, response) => { ... },
  );
```

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -105,7 +105,8 @@ export async function createRouter(
   const { logger, config, database, httpAuth, userInfo, permissions } = options;
 
   const router = Router();
-  router.use(express.json());
+  // Set explicit body size limit to accommodate attachments (up to 50MB total)
+  router.use(express.json({ limit: '60mb' }));
```
Jdubrick: Same as this comment https://github.com/redhat-developer/rhdh-plugins/pull/3359/changes#r3415087306

workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
Jdubrick: And lets move the value itself to `constant.ts` so we can keep all these types of values consistently located

Jdubrick (commented):
One comment about how we should handle the allocation of the updated body size, otherwise looks good

Jdubrick:
Can we add a changeset too @rajin-kichannagari ? Since there is a possibility someone sending a query with the new changes might hit the limit, we should document it in the changes

Jdubrick (approved):
lgtm

PR #3385, opened 2026-06-11
Add stream error handlers to prevent process crashes

rajin-kichannagari:
## Summary
Attach error handlers to streaming responses to prevent unhandled error events from crashing the Node.js process when the LCS connection drops mid-stream.

Addresses security vulnerability identified in Red Hat Product Security threat model (dated Mar 15, 2026).

## Changes
- Add error handler to nodeStream in POST `/v1/query` endpoint (router.ts)
- Add error handler to response object to detect client disconnections
- Add error handlers to both source and transform streams in notebooks query endpoint (`/v1/sessions/:sessionId/query`)
- Ensure streams are properly destroyed on error to prevent hanging connections

## Affected Endpoints
- `/v1/query` endpoint (Lightspeed router)
- `/v1/sessions/:sessionId/query` endpoint (Notebooks router)

## Checklist

- [ ] A changeset describing the change and affected packages

**Testing note:** Verified with manual tests simulating LCS stream failures and client disconnections. Error handlers work correctly and prevent process crashes. Existing unit tests verify the happy path.

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
@@ -471,12 +471,15 @@ export async function createNotebooksRouter(
       };
 
       for (let retries = 0; retries <= MAX_QUERY_RETRIES; retries++) {
+        const abortController = new AbortController();
+
         const response = await fetch(
           `${lightspeedBaseUrl}/v1/responses?user_id=${encodeURIComponent(userId)}`,
           {
             method: 'POST',
             headers: { 'Content-Type': 'application/json' },
             body: JSON.stringify(lightspeedRequest),
+            signal: abortController.signal,
           },
         );
 
@@ -528,7 +531,6 @@ export async function createNotebooksRouter(
             userId,
           );
 
-          // Handle stream errors (RHIDP-13064)
           body.on('error', (error: Error) => {
             logger.error(
               `Upstream stream error while processing notebook query: ${error}`,
@@ -538,6 +540,7 @@ export async function createNotebooksRouter(
                 .status(500)
                 .json({ status: 'error', error: 'Stream error occurred' });
             }
+            abortController.abort();
             transformStream.destroy();
           });
 
@@ -551,15 +554,16 @@ export async function createNotebooksRouter(
                 error: 'Processing error occurred',
               });
             }
+            body.destroy();
+            abortController.abort();
           });
 
-          // Handle client disconnection (RHIDP-13064)
           res.on('error', (error: Error) => {
             logger.warn(
               `Client disconnected while processing notebook query: ${error}`,
             );
+            abortController.abort();
             body.destroy();
-            transformStream.destroy();
           });
 
           body.pipe(transformStream).pipe(res);
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
@@ -535,12 +535,12 @@ export async function createNotebooksRouter(
             logger.error(
               `Upstream stream error while processing notebook query: ${error}`,
             );
-            if (!res.headersSent) {
+            if (res.headersSent) {
+              res.destroy();
+            } else {
               res
                 .status(500)
                 .json({ status: 'error', error: 'Stream error occurred' });
-            } else {
-              res.destroy();
             }
             abortController.abort();
             transformStream.destroy();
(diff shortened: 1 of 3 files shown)


workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts
```diff
@@ -510,11 +510,48 @@ export async function createNotebooksRouter(
...
+            );
+            if (!res.headersSent) {
+              res.status(500).json({
+                status: 'error',
+                error: 'Processing error occurred',
+              });
+            }
+          });
+
+          // Handle client disconnection (RHIDP-13064)
```
JslYoon: No need to add in task #

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -652,6 +652,23 @@ export async function createRouter(
         // Pipe the response back to the original response
         if (fetchResponse.body) {
           const nodeStream = Readable.fromWeb(fetchResponse.body as any);
+
+          // Handle stream errors (RHIDP-13064)
```
JslYoon: No need for (RHIDP-13064)

JslYoon (commented):
Found 3 issues:

  1. **Transform stream error handler doesn't destroy upstream body stream**
  (resource leak)
  
     https://github.com/redhat-developer/rhdh-plugins/blob/39ee72848af2b4352ddd4cdca375e201bf8a2a8d/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts#L526-L531
  
     When `transformStream` errors, `body.destroy()` is never called. The
  upstream stream keeps reading data, causing memory leaks.
  
  2. **Double-destroy race condition in response error handler**
  
 https://github.com/redhat-developer/rhdh-plugins/blob/39ee72848af2b4352ddd4cdca375e201bf8a2a8d/workspaces/lightspeed/plugins/lightspeed-backend/src/service/notebooks/notebooksRouters.ts#L533-L539
  
   `body.destroy()` triggers `body.on('error')` which calls
  `transformStream.destroy()`, then line 536 calls it again.
  `transformStream.destroy()` executes twice.

  3. **Upstream fetch not cancelled on client disconnect** (connection pool
  exhaustion)
  
     https://github.com/redhat-developer/rhdh-plugins/blob/39ee72848af2b4352ddd4cdca375e201bf8a2a8d/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts#L659-L672
  
     `nodeStream.destroy()` doesn't cancel the underlying fetch. The upstream
  HTTP connection stays open, draining data from Lightspeed backend even after
  client disconnects.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -652,6 +655,23 @@ export async function createRouter(
         // Pipe the response back to the original response
         if (fetchResponse.body) {
           const nodeStream = Readable.fromWeb(fetchResponse.body as any);
+
+          nodeStream.on('error', (error: Error) => {
```
Jdubrick, line 661: I was wrong in the issue description saying to use 'error' as the event, we should use `close`. Then when it catches we need to make sure it was a premature close before we emit any error logs and destroy the stream

Jdubrick: This would affect all the `xyz.on('error')` pieces added in this PR

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -652,6 +655,23 @@ export async function createRouter(
...
+            logger.error(
+              `Upstream stream error while processing query: ${error}`,
+            );
+            if (!response.headersSent) {
```
Jdubrick: I think we also may need to add an `else` to this `headersSent` check, since if the stream closed prematurely, I believe the pipe won't properly be destroyed on the response side, so we can destroy it if the stream closed mid-stream. Qodo raised something similar as the 2nd point, which is applicable, but it left the `res` cleanup out

`response.destroy()`

Jdubrick: Would be applicable in all cases where we check this

Jdubrick (commented):
Can we also add some quick tests to verify these handlers?

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
rajin-kichannagari: Changed it to response.on('close') with a !response.writableFinished guard so we only log/destroy on premature closes. Applied across all the handlers in both routers.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
rajin-kichannagari: Added else { response.destroy() } to all the headersSent checks

JslYoon (approved):
lgtm, once Jordan's comments and sonarqube warnings are addressed can be merged

Jdubrick (approved):
lgtm

PR #3386, opened 2026-06-11
fix:  This adds the ability to send nested filters with the same field name

lholmquist:
This is the backend change that addresses https://redhat.atlassian.net/browse/RHIDP-14337 when the user wants to create a filter using either the `variables.initiatorEntity` or `variables.targetEntity`

These values are "nested" values and should use the NestedFilter format, similar to how it is done in the router here: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/orchestrator/plugins/orchestrator-backend/src/service/router.ts#L884-L892

The payload should look something like this if using just one filter:

```
{
    "paginationInfo": {
        "pageSize": 21,
        "offset": 0,
        "orderBy": "start",
        "orderDirection": "DESC"
    },
    "filters": {
        "field": "variables",
        "nested": {
            "operator": "EQ",
            "value": "user:development/guest",
            "field": "initiatorEntity"
        }
    }
}
```

If using multiple filters, the payload might look something like this:

```
{
    "paginationInfo": {
        "pageSize": 21,
        "offset": 0,
        "orderBy": "start",
        "orderDirection": "DESC"
    },
    "filters": {
        "operator": "AND",
        "filters": [
            {
                "field": "variables",
                "nested": {
                    "operator": "EQ",
                    "value": "component:default/my-component",
                    "field": "targetEntity"
                }
            },
            {
                "field": "variables",
                "nested": {
                    "operator": "EQ",
                    "value": "user:development/guest",
                    "field": "initiatorEntity"
                }
            }
        ]
    }
}
```

The code will now group up the field paramter in a nested query if the field is defined multiple times and will then format the query correctly.

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
@@ -96,17 +96,29 @@ function handleNestedFilter(
   type: ProcessType,
   filter: NestedFilter,
 ): FilterClause {
-  const subClauses = buildFilterCondition(
-    introspection,
-    type,
-    filter.nested,
-    true,
-  );
-
-  const filterClause: FilterClause = {
-    clauseVariable: subClauses.clauseVariable,
-    clause: `${filter.field}: {${subClauses.clause}}`,
-  };
+  // check if the nested param is an array or an object
+  // this makes sure that we can handle the case where there are multiple filters for the same field
+  let filterClause: FilterClause;
+  if (Array.isArray(filter.nested)) {
+    const subClauses = filter.nested.map(n =>
+      buildFilterCondition(introspection, type, n, true),
+    );
+    filterClause = {
+      clauseVariable: subClauses.flatMap(cl => cl.clauseVariable),
+      clause: `${filter.field}: {${subClauses.map(cl => cl.clause).join(', ')}}`,
+    };
+  } else {
+    const subClauses = buildFilterCondition(
+      introspection,
+      type,
+      filter.nested,
+      true,
+    );
+    filterClause = {
+      clauseVariable: subClauses.clauseVariable,
+      clause: `${filter.field}: {${subClauses.clause}}`,
+    };
+  }
 
   return filterClause;
 }
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
@@ -425,3 +425,36 @@ function getGraphQLOperator(operator: FieldFilterOperatorEnum): string {
 function nonSecureRandomAlphaNumeric() {
   return randomBytes(8).toString('hex');
 }
+
+// For nested filters, there might be more than one filter for the same field
+// so we need to group them by the field and then combine the nested filters into an array
+export function processFilters(filter: Filter[]): Filter[] {
+  if (filter && 'filters' in filter) {
+    const grouped = (filter.filters as Filter[]).reduce<
+      Record<string, Filter[]>
+    >((acc: Record<string, Filter[]>, item: Filter) => {
+      if ('field' in item) {
+        acc[item.field] ??= [];
+        acc[item.field].push(item);
+      }
+      return acc;
+    }, {});
+
+    const newFilters: Filter[] = [];
+
+    for (const [field, filtersForField] of Object.entries(grouped)) {
+      if (filtersForField.length > 1) {
+        const nested = filtersForField
+          .filter((f): f is NestedFilter => 'nested' in f)
+          .map(f => f.nested);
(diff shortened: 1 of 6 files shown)


workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
```diff
@@ -413,3 +425,37 @@ function getGraphQLOperator(operator: FieldFilterOperatorEnum): string {
...
+// For nested filters, there might be more than one filter for the same field
+// so we need to group them by the field and then combine the nested filters into an array
+export function processFilters(filter: Filter): Filter {
+  if (isLogicalFilter(filter)) {
+    const grouped = filter.filters.reduce<Record<string, Filter[]>>(
+      (acc, item) => {
+        if ('field' in item) {
+          acc[item.field] ??= [];
+          acc[item.field].push(item);
+        }
```
ciiay: Looks like the grouping logic only keeps children that have a `field` property which could cause a regression issue. Children that are `LogicalFilter` (no `field`) are never added to grouped and are silently dropped from the output.

In `router.ts` (~897–901), when a non-admin calls getInstances, the backend combines filters like this:
```
filters = {
  operator: 'AND',
  filters: [nestedVariablesFilter, requestFilters],  // requestFilters is often a LogicalFilter
};
```
From the frontend (WorkflowRunsTabContent.tsx), when the user applies table filters (status, date, etc.) alongside an entity-scoped targetEntity filter, `requestFilters` is wrapped as:
```
{
  operator: 'AND',
  filters: [
    { field: 'variables', nested: { field: 'targetEntity', ... } },
    { field: 'processId', ... },
    { field: 'state', ... },
  ],
}
```
So the full filter tree becomes:
```
AND [
  { field: 'variables', nested: initiatorEntity },
  AND [
    { field: 'variables', nested: targetEntity },
    { field: 'processId', ... },
    { field: 'state', ... },
  ],
]
```

After `processFilters`, only the `initiatorEntity` filter survives, the inner `AND` (and with it `targetEntity`, `processId`, `state`) is lost.

On upstream/main, this nested-AND shape actually worked because duplicate variables keys were at different nesting levels:
```
and: {
  variables: { initiatorEntity: ... },
  and: { variables: { targetEntity: ... }, processId: ..., state: ... }
}
```

ciiay (commented):
Thanks for the PR 👏 

Good direction on supporting nested filters with the same field name. The `handleNestedFilter` array handling looks right. Just left a comment about a possible regression in `processFilters` function.

workspaces/orchestrator/plugins/orchestrator-backend/src/helpers/filterBuilder.ts
lholmquist: Thanks for the review.  I'll be honest, i was never really thrilled with using that group by to process the filters. 

~~I think it would probably be better if the payload already had the `nested` parameter as an array so we wouldn't have to do all that group by stuff.  I think to do that, i need to update the API a little to allow for that since when i first tried it,  i was getting errors on the router.~~

~~So i'm thinking that the payload sent to the backend might look something like this then:~~

I updated the function that processes the filter first,  i think this should take care of the payload you mentioned

lokanandaprabhu (commented):
Tested this PR with the All runs UI (nested filters for variables.targetEntity and variables.initiatorEntity).

**What works -** 
                       Combined filters with different nested fields work after groupNestedFilters (e.g. router auto initiatorEntity + UI targetEntity, or targetEntity + status).
Single targetEntity nested filter returns 200 (no GraphQL validation error).

**What doesn’t work — Run by filter**
                        For non-admin users, router.ts always injects an initiatorEntity filter before calling getInstances:

router.ts (Lines 882-902)

```
        if (!isUserAuthorizedForInstanceAdminView) {
          const initiatorEntityFilter: FieldFilter = {
            operator: 'EQ',
            value: initiatorEntity,
            field: 'initiatorEntity',
          };
          // ... merged with requestFilters via AND
        }
```

`isUserAuthorizedForInstanceAdminView` is currently hardcoded to false (permission check commented out at L871–876).

When the UI sends a Run by filter (variables.initiatorEntity), the backend merges two initiatorEntity filters under variables. After grouping, GraphQL still ends up with duplicate initiatorEntity fields:

`Validation Error: There can be only one field named 'initiatorEntity'
`
Request: Please enable isUserAuthorizedForInstanceAdminViewPermission so instance admins skip auto-scoping and can filter by Run by via the API. Non-admins should remain auto-scoped to their own runs (UI should not send initiatorEntity for them).

lholmquist:
> isUserAuthorizedForInstanceAdminView is currently hardcoded to false (permission check commented out at L871–876).

yea,  i was testing something out and forgot to switch it back.  sorry about that.  🤦

lholmquist:
@lokanandaprabhu I reverted the hardcoding,  could you try it again

lokanandaprabhu (approved):
/lgtm

PR #3436, opened 2026-06-17
chore(x2a): job-script gitignore issues

yray-pixel:
Fixes FLPATH-4383

  ### Problem
  
  When source/target repositories have `.gitignore` rules blocking files like `migration-plan.md`, the x2a job script
  would create these files but silently fail to commit them to the target repository.

  ### Solution

  Add detection and logging for files that exist but weren't committed due to `.gitignore` rules:
  - Use `git ls-files --others --ignored --exclude-standard` to find gitignored files in PROJECT_DIR
  - Log clear warnings for each gitignored file
  - Display summary count of affected files

  ### Testing

  ✅ Tested with repos containing `*.md` in `.gitignore`
  ✅ Verified warnings appear for gitignored files
  ✅ Verified normal files still commit correctly
  ✅ Verified empty PROJECT_DIR doesn't cause errors

  ### Changes

  - Added gitignore detection logging in `cleanup()` function


eloycoto (approved):
congrats on your first PR in 🎉

mareklibra (commented):
LGTM, please just add changeset.

In the workspaces/x2a call `yarn changeset`. Then choose just the x2a-backend plugin and `patch` level. Write a short comment there.

Git add the newly created file.

These changesets are used for both the CHANGELOG files and triggering release.

PR #3425, opened 2026-06-16
feat: add Codecov components grouped by support level

gustavolira:
## Summary

This PR implements **per-support-level coverage reporting** on Codecov dashboard using the **Components** feature. Coverage is now grouped by plugin maturity:

- **GA Plugins** (17 workspaces)
- **Tech-Preview Plugins** (8 workspaces)  
- **Community Plugins** (39 workspaces)
- **Dev-Preview Plugins** (6 workspaces)

## Changes

✅ **Added component_management to codecov.yml**
- Mapped 70 workspace paths across 4 support-level components
- Preserves existing per-workspace flags (no historical data loss)
- Some workspaces appear in multiple components (e.g., `backstage` has packages at all 4 levels)

✅ **Created automation script**: `scripts/generate-codecov-components.sh`
- Extracts workspaces by support level from `rhdh-plugin-export-overlays` metadata
- Generates YAML fragment for `codecov.yml` component definitions
- Enables easy sync when support levels change

✅ **Documentation**: `docs/codecov-support-levels.md`
- Dashboard usage and interpretation
- Maintenance procedures (automated + manual)
- FAQ for multi-level workspaces
- **Known limitations** section documenting coverage skew

## Implementation Details

**How it works:**
- Support level metadata lives in `rhdh-plugin-export-overlays/workspaces/*/metadata/*.yaml` as `spec.support`
- Script extracts unique workspaces per support level and generates component paths
- Codecov aggregates coverage across all matching paths for each component

**Multi-level workspaces:**
Workspaces like `backstage`, `homepage`, and `analytics` appear in **multiple components** because they contain packages with different support levels. This is **correct** — coverage from those workspaces contributes to all relevant support-level components.

## Known Limitation: Coverage Skew

**4 workspaces have mixed support levels**, which causes coverage skew in component metrics:

- `backstage`: 43% GA (12 packages) + 57% non-GA (16 packages)
- `homepage`: 17% GA + 83% TP
- `analytics`: 50% GA + 50% Community  
- `roadie-backstage-plugins`: 50% GA + 50% Community

**Impact:** The `ga-plugins` component includes coverage from non-GA packages in these workspaces, so the component coverage is a **weighted average** rather than pure GA coverage.

**Example:** If `backstage` GA packages have 85% coverage but TP/Community have 60%, the `ga-plugins` component shows ~65%.

**When this matters:**
- ✅ **For visibility and trends**: Component metrics are still useful
[...]

PR #3463, opened 2026-06-18
chore: sync codecov components dynamically from overlay support levels

gustavolira:
## Summary

Follow-up to #3425. Makes the Codecov support-level **components** self-update from `rhdh-plugin-export-overlays` metadata (`spec.support`) instead of being hand-maintained, so a plugin promoted from e.g. tech-preview → GA is moved to the right component automatically.

While wiring this up I found that #3425 hard-coded **70 component paths**, but this repo only has **24 workspaces** — most of those paths (`workspaces/3scale/`, `quay/`, `acr/`, `backstage/`, …) don't exist here; they're overlay workspace names sourced from `backstage/community-plugins` and other upstreams, so Codecov tracked nothing for them. The generated block now contains only the workspaces that actually exist in this repo.

## Changes

- **`scripts/generate-codecov-components.sh`** — rewritten:
  - Reads `spec.support` from overlay metadata and **filters to workspaces present in this repo**.
  - Three modes: print (default), `--write` (rewrites only the `component_management` block in `codecov.yml`, preserving flags/coverage/ignore), `--check` (exits 1 + diff on drift).
  - Deterministic output (no timestamps) so `--check` is stable; portable across GNU/BSD; shellcheck-clean; back-compatible with the old positional invocation.
- **`.github/workflows/sync-codecov-components.yml`** — new:
  - **Weekly cron + `workflow_dispatch`**: clones the overlay, runs `--write`, opens a PR when a support level changed (`peter-evans/create-pull-request`).
  - **`pull_request` drift check**: runs `--check` on PRs touching `codecov.yml` and fails if it's out of sync.
  - Reuses the repo's pinned action SHAs, `RHDH_BOT_TOKEN`, harden-runner, and repo-guard conventions.
- **`codecov.yml`** — regenerated to the filtered set (GA 6 / Tech-Preview 4 / Community 3 / Dev-Preview 5 = 18 real paths), dropping paths for non-existent workspaces.
- **`docs/codecov-support-levels.md`** — updated for the automated flow and corrected counts/examples.

## Verification

- Print / `--write` / `--check` all exercised against a real overlay clone; `--write` is idempotent; negative drift test exits 1 and pinpoints the path.
- `codecov.yml` remains valid YAML with all existing flags intact.
- `shellcheck` clean.

## Related

- Follow-up to #3425 — RHIDP-13511 / Epic RHIDP-13497 / RHDHPLAN-851


hopehadfield (approved):
Nice! LGTM

PR #3512, opened 2026-06-20
feat(lightspeed): expose dedicated NFS module federation entry points for FAB and translations

its-mitesh-kumar:
## Description

This PR adds dedicated Module Federation entry points for the Lightspeed FAB (Floating Action Button) module and the Translations module, and fixes a React Context duplication bug that occurs across MF chunk boundaries in the New Frontend System (NFS).

These changes are required because `dynamicFrontendFeaturesLoader()` in Backstage NFS **only loads default exports** from each exposed module. The existing `lightspeedFABModule` and `lightspeedTranslationsModule` are **named exports** from `./src/alpha/index.tsx`, which means they cannot be discovered and loaded by the NFS dynamic plugin loader without a dedicated entry point that re-exports them as `default`.

### Gaps / Reasons for Each Change

| File | Gap / Reason |
|------|-------------|
| `src/alpha/lightspeedFABModuleExport.ts` | **NFS loader only loads default exports.** The `lightspeedFABModule` (which provides the `AppRootWrapperBlueprint` for the FAB button) was a named export and therefore invisible to `dynamicFrontendFeaturesLoader`. This file re-exports it as `default`. |
| `src/alpha/lightspeedTranslationsModuleExport.ts` | **Same gap for translations.** The `lightspeedTranslationsModule` (which registers `TranslationBlueprint` resources for multi-language support) was also only a named export. This file provides the required default export. |
| `package.json` (exports, typesVersions, scalprum) | **Module Federation container registration.** Without entries in `scalprum.exposedModules`, the RHDH CLI cannot build these files as separate MF containers. Without `exports`/`typesVersions`, TypeScript consumers cannot import the subpath modules. |
| `src/components/LightspeedDrawerContext.tsx` | **React Context duplication across MF boundaries.** In Module Federation, each remote chunk gets its own module scope, causing `createContext()` to create duplicate Context instances. The "Dock to Window" display mode failed with `useLightspeedDrawerContext must be used within a LightspeedDrawerProvider` because the provider and consumer used different Context objects. The `globalThis` singleton pattern ensures a single shared Context instance across all MF chunks. |

### How to enable in NFS mode

After deploying the rebuilt dynamic plugin, add to `app-config.yaml`:

```yaml
app:
  extensions:
    - app-root-wrapper:app/drawer
    - app-root-wrapper:app/lightspeed-fab
    - translation:app/lightspeed-translations
    - api:app/app-language:
        config:
          availableLanguages: ['en', 'de', 'es', 'fr', 'it', 'ja']
          defaultLanguage: 'en'
```

## Fixed
- TODO (no Jira ticket provided)

## ✔️ Checklist


its-mitesh-kumar:
Closing in favor of https://github.com/redhat-developer/rhdh-plugins/pull/3513

PR #3528, opened 2026-06-22
chore(scorecard): align examples to see all at once

christoph-jerolimov:

`yarn start` before:

image

With this change:

image


Eswaraiahsapram (approved):
Looks good to me

workspaces/scorecard/examples/components/all-scorecards.yaml
```diff
@@ -1,29 +1,12 @@
...
-  name: all-scorecards-service-different-owner
-  annotations:
-    github.com/project-slug: redhat-developer/rhdh-plugins
-    backstage.io/source-location: url:https://github.com/redhat-developer/rhdh-plugins
-    jira/project-key: RSPT
-    jira/label: JupiterTeam
-spec:
-  type: service
-  owner: rhdh-team
+  owner: group:development/guests
```
dzemanov: I think we wanted to leave at least one scorecard to have owner as a user instead of group to cover that use case. It doesn't necessarily have to be all-scorecards.yaml. Is there any other scorecard covering it?

dzemanov (dismissed):

workspaces/scorecard/examples/components/all-scorecards.yaml
christoph-jerolimov: Nope. But that was the problem why the catalog switches to show just this catalog.

But I can investigate why it doesn't show everything because the guest user is in both groups.

Thanks for the feedback, will take a look.

christoph-jerolimov: I understand the issue now, the problem was that the app-config included a `userEntityRef` configuration for the guest provider. But if this is configured it needs to specific also the `ownershipEntityRefs` for the groups. I removed that configuration so that ownership entity refs is loaded from the catalog.

PR #3536, opened 2026-06-22
feat(extensions): replace disabled with enabled field in plugin configuration

hopehadfield:
### Summary

 Replaces the `disabled` field with `enabled` in the extensions plugin UI and backend, per [RHIDP-14727](https://redhat.atlassian.net/browse/RHIDP-14727) and [RHIDP-11983](https://redhat.atlassian.net/browse/RHIDP-11983). The backend API accepts both fields for backward compatibility (`enabled` takes precedence), but the UI and storage layer exclusively use `enabled`.

### Changes

**extensions-common** — Added enablePackage/enablePlugin to the API interface and backend client. Removed disablePackage/disablePlugin.

**extensions-backend** — Router accepts both {enabled} and {disabled} in request body via `resolveEnabledField()`, with `enabled` taking precedence. Storage writes `enabled` to YAML. Validation accepts both fields.

**extensions (frontend)** — `useEnablePlugin` hook and all components send {enabled} exclusively. YAML generation templates use `enabled: true/false`.

 ### Backward compatibility

 - Existing YAML files with disabled: true/false still pass validation
 - API clients sending {disabled: true} still work (resolved to enabled: false by the router)
 - New configs written by the UI use enabled exclusively


workspaces/extensions/plugins/extensions/src/components/InstalledPackages/RowActions.tsx
its-mitesh-kumar: We need to update `disabled` to `enabled` here too

its-mitesh-kumar (commented):
Everything else look good to me codewise. Passing the revoew to @debsmita1 as she has more context on it.

debsmita1 (commented):
Tested this locally

/lgtm

dzemanov:
The route paths are staying the same? (e.g not deprecated and switched to enable route)
- /plugin/:namespace/:name/configuration/disable
- '/package/:namespace/:name/configuration/disable',

dzemanov (approved):

debsmita1 (approved):

PR #3540, opened 2026-06-23
feat(software-catalog-mcp-extras): add full-text search via queryEntities

savitojs:
The `query-catalog-entities` action only supports exact-match filters via `catalog.getEntities()`, making it impossible to find entities by partial name or description. Users must know the exact `metadata.name` to find anything - searching "argocd" won't find `devhub-argo-infra`.

The `CatalogService` interface already exposes `queryEntities()` with `fullTextFilter` support, but this plugin never uses it. This PR adds a `search` input parameter that enables substring matching across name, title, and description.

Changes:
- Add `search` input parameter to the tool schema with description clarifying partial/substring matching
- When `search` is provided, use `catalog.queryEntities()` with `fullTextFilter` targeting `metadata.name`, `metadata.title`, and `metadata.description`
- The `search` parameter can be combined with existing exact-match filters (kind, type, owner, lifecycle, tags)
- When `search` is not provided, behavior is identical to the current implementation (`catalog.getEntities()` with exact filters)
- Update tool description with search examples
- Add tests for search-only, search+kind combo, verbose+search, and backward compatibility

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/workspaces/mcp-integrations/.changeset/add-fulltext-search.md
+++ b/workspaces/mcp-integrations/.changeset/add-fulltext-search.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-software-catalog-mcp-extras': minor
+---
+
+Add full-text search parameter to query-catalog-entities action using catalog.queryEntities() with fullTextFilter support for partial matching across entity name, title, and description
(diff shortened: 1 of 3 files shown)


johnmcollier (commented):
Change looks good @savitojs but FWIW, there's a `query-catalog-entities` tool now available upstream with similar query functionality (https://github.com/backstage/backstage/blob/master/plugins/catalog-backend/src/actions/createQueryCatalogEntitiesAction.ts)

We'll likely deprecate this action in favour of that one in 2.1 given the overlap. You may want to see if the upstream action's filtering functionality meets your needs

PR #3546, opened 2026-06-23
feat(homepage): add `unless`  exclusion and `tags` for RBAC conditional filtering

karthikjeeyar:
### Fixes:

https://redhat.atlassian.net/browse/RHIDP-14648
https://redhat.atlassian.net/browse/RHIDP-14649

Adds two new config-level features to homepage default widgets:

 - **`unless` config block**: inverse of `if` — uses the same shape (`users`, `groups`, `permissions`) but acts as a denylist. Checked before `if` so deny always wins. On group nodes, prunes the entire subtree.
  - **`tags` on widget nodes**: optional string array (e.g. `['admin', 'developer']`) passed through to the API response. Used with the new `HAS_TAG` permission rule for RBAC conditional policy filtering. Widgets without tags bypass tag-based filtering.
  - **`HAS_TAG` permission rule**: new rule registered alongside `HAS_WIDGET_ID` for the `homepage-default-widget` resource type. Matches 
  widgets that have at least one overlapping tag.
  - **Refactored permission flow**: moved `authorizeConditional` from `buildUserContext` into `DefaultWidgetsService` so the top-level RBAC filtering runs.
  - **README updated** with `unless`, `tags`, filtering layer and permission rules table.
  - **Frontend**: mount point default props are now merged with widget-level props (`{ ...mountPoint.config?.props, ...widget.props }`).

## Screenshots:

image

## How to test

Test widgets included in app-config.yaml, to switch different personas

  1. `examples/org.yaml` — change the guest user's memberOf to match:
```yaml
  spec:
    memberOf: [guests, default/developers] //  To test as admin change it to `default/admins`
  ```
2. Modify the `conditional-policies.yaml` tags to filter the widgets.
3. 
  

### DIFF
--- a/workspaces/homepage/.changeset/unless-and-tags.md
+++ b/workspaces/homepage/.changeset/unless-and-tags.md
@@ -1,0 +1,11 @@
+---
+'@red-hat-developer-hub/backstage-plugin-homepage': patch
+'@red-hat-developer-hub/backstage-plugin-homepage-backend': minor
+'@red-hat-developer-hub/backstage-plugin-homepage-common': minor
+---
+
+Add `unless` exclusion block and `tags` for RBAC conditional policy filtering to homepage default widgets.
+
+`unless` is the denylist counterpart to `if` — it uses the same shape (`users`, `groups`, `permissions`) and hides a widget when any condition matches. Deny wins over `if`, and on group nodes it prunes the entire subtree.
+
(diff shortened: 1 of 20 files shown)


karthikjeeyar:
Thank you @christoph-jerolimov, the simplifying permission changes looks good to me

christoph-jerolimov (approved):
Thanks for the idea and great implementation @karthikjeeyar 

It was required to add the admin and user now to `rbac-policy.csv`, for me this looks right because you've added tags also to the default widgets.

I think we need to cleanup the test here, but I think we can do that in another PR and when adding support for the NFS.

PR #3551, opened 2026-06-23
chore(x2a): change publish phase actions order

yray-pixel:
Fix: Ensure AAP syncs after GitHub push completes (FLPATH-4398)

  Problem

  AAP was unable to discover migrated Ansible playbooks because it was syncing the Git repository before GitHub
  finished processing the push..

  Root Cause

  The job script was calling publish-aap (which triggers AAP project sync) immediately after generating the Ansible
  project files, but before committing and pushing to GitHub. This created a race condition where AAP's sync often
  completed before the new playbooks were available on GitHub.

  Solution

  This PR refactors the job script to:

  1. Commit and push to GitHub first - Ensures the Ansible project is available on the remote repository
  2. Then call publish-aap - AAP now syncs against the latest commit with all playbooks present

  Changes

  New git_commit_and_push() helper function:
  - Encapsulates common git commit/push logic
  - Handles both new and existing branches

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
+++ b/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
@@ -207,6 +207,67 @@ git_target_repo() {
   git -c "credential.helper=!f() { test \"\$1\" = get && printf 'username=${username}\\npassword=${password}\\n'; }; f" "$@"
 }
 
+# Commit and push changes to target repository.
+# Returns: Sets COMMIT_ID to the pushed commit SHA, or exits on push failure.
+git_commit_and_push() {
+  local commit_message="$1"
+  local fail_on_error="${2:-true}"  # Default: exit on push failure
+
+  if [ ! -d /workspace/target/.git ]; then
+    echo "ERROR: Target repository not initialized"
+    return 1
+  fi
+
+  cd /workspace/target
+
+  # Sanitize secrets from output files before committing
+  sanitize_secrets "${PROJECT_PATH:-/workspace/target}"
+
+  git add "${PROJECT_DIR}" 2>/dev/null || git add -A || true
+
+  # Check if there are changes to commit
+  if git diff --cached --quiet; then
+    echo "No changes to commit"
+    COMMIT_ID=$(git rev-parse HEAD 2>/dev/null || echo "")
+    return 0
+  fi
+
+  git commit -m "${commit_message}" || echo "Commit failed, but continuing"
+
+  # Push to remote
+  if [ "${TARGET_BRANCH_IS_NEW}" = "true" ]; then
+    echo "Pushing new branch '${TARGET_REPO_BRANCH}' to ${TARGET_REPO_URL}..."
+    if ! git_target_repo push -u origin "${TARGET_REPO_BRANCH}"; then
+      local error_msg="Failed to push new branch '${TARGET_REPO_BRANCH}' to ${TARGET_REPO_URL}"
+      echo "ERROR: ${error_msg}"
+      if [ "${fail_on_error}" = "true" ]; then
+        ERROR_MESSAGE="${error_msg}"
+        exit 1
+      else
+        PUSH_FAILED="${error_msg}"
+        return 1
+      fi
+    fi
+  else
+    echo "Pushing to existing branch '${TARGET_REPO_BRANCH}'..."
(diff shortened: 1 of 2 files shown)


workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -584,6 +624,20 @@ case "${PHASE}" in
     echo "=== Ansible project contents ==="
     find "${ANSIBLE_PROJECT_DIR}" -type f | head -50
 
+    # Commit and push the Ansible project to git BEFORE calling publish-aap
+    # This ensures AAP syncs the latest commit with the new playbooks
+    echo ""
+    echo "=== Committing and pushing Ansible project to git ==="
+    git_commit_and_push "x2a: ${PHASE} phase for ${MODULE_NAME}
```
eloycoto: can we define the variable for the commit message? it's easy to read the function and the input-variable.

eloycoto: like the cleanup: local commit_msg

eloycoto (commented):
In the publish, it commits twice, no?

```
cleanup() {
  local exit_code=$?
  set +e

  # Always try to commit and push whatever is in the working directory
  if [ -d /workspace/target/.git ]; then
    local commit_msg="x2a: ${PHASE} phase for ${MODULE_NAME:-project}

Phase: ${PHASE}
Project: ${PROJECT_ID}
Module: ${MODULE_NAME:-N/A}
Job: ${JOB_ID}

Co-Authored-By: ${GIT_AUTHOR_NAME} <${GIT_AUTHOR_EMAIL}>
"
    git_commit_and_push "${commit_msg}" false  # Don't exit on error in cleanup
  fi

  if [ "$TERMINATED" = true ]; then
    report_result "error" "Job was terminated"
  elif [ ${exit_code} -ne 0 ]; then
    report_result "error" "${ERROR_MESSAGE:-Script failed with exit code ${exit_code}}"
  elif [ -n "${PUSH_FAILED}" ]; then
    report_result "error" "${PUSH_FAILED}"
  else
    report_result "success" ""
  fi
}

```

so, should be a check that SKIP_COMMIT entry, no?

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
yray-pixel: 1. The git_commit_and_push() function  prevents double commits - it checks `git diff --cached --quiet` at line 229 and returns early with "No changes to commit".
So when cleanup runs after we've already committed in the publish phase, it won't commit. 
2. Extracted the commit message into a `local commit_msg` variable for better readability

workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
```diff
@@ -215,37 +276,16 @@ cleanup() {
...
   # Always try to commit and push whatever is in the working directory
   if [ -d /workspace/target/.git ]; then
-    cd /workspace/target
-
-    # Sanitize secrets from output files before committing
-    sanitize_secrets "${PROJECT_PATH:-/workspace/target}"
-
-    git add "${PROJECT_DIR}" 2>/dev/null || git add -A || true
-    git commit -m "x2a: ${PHASE} phase for ${MODULE_NAME:-project}
+    local commit_msg="x2a: ${PHASE} phase for ${MODULE_NAME:-project}
```
mareklibra: This is a repeating pattern, what about introducing sort of `build_commit_message()` function?

yray-pixel: good point, changed it to use separate function to reduce duplication

eloycoto (changes requested):
I tried to run this, and I got this issue:

```
/workspace/target/wqeqweqwe-41527f/ansible-project/roles/nginx_multisite/templates/sysctl-security.conf.j2
/workspace/target/wqeqweqwe-41527f/ansible-project/run_nginx_multisite.yml
=== Committing and pushing Ansible project to git ===
/bin/bash: line 638: local: can only be used in a function
=== Sanitizing secrets from output files ===
  No secrets found in output files
No changes to commit
Reporting result: status=error, phase=publish
```

eloycoto (approved):

PR #3554, opened 2026-06-23
feat(install-dynamic-plugins): add enabled field in DP config

hopehadfield:
## Description
Duplicate of https://github.com/redhat-developer/rhdh/pull/4937, which was dropped during migration.

Adds the enabled field with backwards compatibility

See https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2577 for relevant change in DPDY generation

## Which issue(s) does this PR fix
Fixes [RHIDP-11988D](https://redhat.atlassian.net/browse/RHIDP-11988)

### DIFF
--- a/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
+++ b/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
@@ -298,12 +298,14 @@ function copyPluginFields(
     if (skipSet.has(k) || isForbiddenKey(k)) continue;
     safeSet(dst, k, v);
   }
-  // When the override introduces one activation field, clear the opposite
-  // so the merged record never carries both (which would trigger a spurious
-  // "specifies both 'enabled' and 'disabled'" warning in isPluginDisabled).
-  if ('enabled' in src && 'disabled' in dst)
+  // When the override introduces a valid boolean activation field, clear the
+  // opposite so the merged record never carries both (which would trigger a
+  // spurious "specifies both" warning in isPluginDisabled).  Only act on
+  // actual booleans — a non-boolean value is treated as "unset" by
+  // isPluginDisabled and should not displace a valid value on dst.
+  if (typeof src.enabled === 'boolean' && 'disabled' in dst)
     delete (dst as Record<string, unknown>).disabled;
-  if ('disabled' in src && 'enabled' in dst)
+  if (typeof src.disabled === 'boolean' && 'enabled' in dst)
     delete (dst as Record<string, unknown>).enabled;
 }
 
--- a/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
+++ b/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
@@ -423,7 +423,7 @@ function processOciEntry(
 ): void {
   const pkg = plugin.package;
   if (typeof pkg !== 'string' || !pkg.startsWith(OCI_PROTO)) return;
-  const disabled = isPluginDisabled(plugin, log);
+  const disabled = isPluginDisabled(plugin);
   const parsed = tryParseOciRegistryAndPath(pkg);
   if (!parsed) {
     logInvalidOciFormat(pkg, sourceFile, disabled);
--- a/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
+++ b/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
@@ -298,6 +298,13 @@ function copyPluginFields(
     if (skipSet.has(k) || isForbiddenKey(k)) continue;
     safeSet(dst, k, v);
   }
+  // When the override introduces one activation field, clear the opposite
+  // so the merged record never carries both (which would trigger a spurious
+  // "specifies both 'enabled' and 'disabled'" warning in isPluginDisabled).
+  if ('enabled' in src && 'disabled' in dst)
+    delete (dst as Record<string, unknown>).disabled;
+  if ('disabled' in src && 'enabled' in dst)
+    delete (dst as Record<string, unknown>).enabled;
 }
 
 function isEqual(a: unknown, b: unknown): boolean {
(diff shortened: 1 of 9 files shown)


workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
```diff
@@ -112,3 +122,52 @@ export function effectivePullPolicy(plugin: {
...
+ *  4. When neither is set         → default to `false` (not disabled).
+ *
+ * Non-boolean values (e.g. `enabled: 'false'` from unquoted YAML) are
+ * treated as unset and a warning is emitted, preventing JS truthiness
+ * from silently flipping activation state.
+ *
+ * The `warn` callback receives the warning message string.  Pass `log` or
+ * leave it out for silent resolution (unit tests, hashing).
+ */
+export function isPluginDisabled(
```
gustavolira, line 143: This helper both resolves the state and emits warnings, and it gets called at every stage of the pipeline: `preMergeOciDisabledState` -> `processOciEntry`, then `categorize`, then `installNpmPlugin`/`installOciPlugin`. For a plugin that resolves to enabled (so it isn't skipped early), the same "both fields"/"non-boolean" warning fires in `categorize` and again in the installer, and for OCI a third time in the pre-merge pass. One mis-set field reads like two or three separate problems. Could we warn once during merge/validation and let the hot-path callers resolve silently?

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
```diff
@@ -112,3 +122,52 @@ export function effectivePullPolicy(plugin: {
...
+  if (plugin.disabled !== undefined && !hasDisabled) {
+    warn?.(
+      `WARNING: Plugin ${plugin.package} has non-boolean 'disabled: ${String(plugin.disabled)}'. ` +
+        `Expected true or false; ignoring the field.`,
+    );
+  }
+
+  if (hasEnabled && hasDisabled) {
+    warn?.(
+      `WARNING: Plugin ${plugin.package} specifies both 'enabled' and 'disabled'. ` +
```
gustavolira, line 165: `copyPluginFields` in merger.ts copies every field on override without clearing the complement, so if an included file sets `disabled: true` and the main config overrides with `enabled: true`, the merged record ends up carrying both. Then this branch warns "specifies both 'enabled' and 'disabled'" even though the user set them in separate files, which is exactly the override this feature is meant to support. The new merger.test.ts case ("handles enabled overriding disabled from a lower level") confirms the stale `disabled` survives. Should the override clear the opposite field, or should this warning be scoped to a single source spec rather than the post-merge record?

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.ts
```diff
@@ -551,7 +552,7 @@ export function filterDisabledOciPlugins(
         log(`\n======= Disabling OCI plugin ${pkg}`);
         continue;
       }
-      if (!parsed && plugin.disabled === true) {
+      if (!parsed && isPluginDisabled(plugin)) {
```
gustavolira, line 564: nit: `processOciEntry` calls `isPluginDisabled(plugin, log)` but here it's called without the callback. If that's deliberate to avoid duplicating the warning the pre-merge pass already logged, fair enough, but the asymmetry is invisible to the next reader. A one-line comment saying why this path stays silent would stop someone from "fixing" it back to a double warning.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
```diff
@@ -112,3 +122,52 @@ export function effectivePullPolicy(plugin: {
     ? PullPolicy.ALWAYS
     : PullPolicy.IF_NOT_PRESENT;
 }
+
+/**
+ * Resolve the effective disabled state from the `enabled` and `disabled`
+ * fields on a plugin spec.  Precedence rules (per RHIDP-11983):
```
gustavolira, line 128: nit: This cites RHIDP-11983, but the PR is fixing RHIDP-11988. One of them is a typo, worth confirming which ticket actually owns this so the breadcrumb points somewhere real.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
```diff
@@ -112,3 +122,52 @@ export function effectivePullPolicy(plugin: {
...
+  if (hasEnabled && hasDisabled) {
+    warn?.(
+      `WARNING: Plugin ${plugin.package} specifies both 'enabled' and 'disabled'. ` +
+        `The 'enabled' field takes precedence; please use only 'enabled'.`,
+    );
+    return !plugin.enabled;
+  }
+  if (hasEnabled) return !plugin.enabled;
+  if (hasDisabled) return plugin.disabled === true;
+  return false;
```
gustavolira, line 172: nit: When `enabled` is non-boolean (e.g. YAML `enabled: flase` parsing to the string "flase") and there's no `disabled`, this returns false, so a typo'd enable flag defaults to the plugin running. The warning above is what makes that acceptable, just flagging it so it's a deliberate choice and not a surprise.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
hopehadfield: The epic is linked here (RHIDP-11983) whereas RHIDP-11988 is the child issue.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
hopehadfield: Just updated so that the warning is only emitted once.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/types.ts
hopehadfield: Updated so that the override clears the opposite field.

gustavolira:
/review
--pr_reviewer.inline_code_comments=true
-i
--pr_reviewer.require_score_review=true
--pr_reviewer.require_can_be_split_review=true
--pr_reviewer.num_code_suggestions="2"

gustavolira (approved):

PR #3571, opened 2026-06-24
RHIDP-14173: Implement scalar aggregation KPI types

imykhno:

Implemented logic to support the following aggregate metric types: `average`, `sum`, `max`, `min`, and `count` for Scorecard aggregated KPI cards.

## This PR for:
* [RHIDP-14173](https://redhat.atlassian.net/browse/RHIDP-14173)

## :heavy_check_mark: Checklist

### Preparing
* If using `app-config.local.yaml` add new aggregation types under `scorecard.aggregationKPIs` (example added to the `app-config.yaml`):
``` yaml
scorecard:
  aggregationKPIs:
    totalOpenBugs:
      title: Total Open Bugs
      description: Sum of open issues across owned entities
      type: sum
      metricId: jira.openIssues
      options:
        thresholds:
          rules:
            - key: success
              expression: '>=80'
              color: '#6bb300' # green
            - key: warning
              expression: '<80'
              color: 'rgb(224, 189, 108)' # light orange
    avgOpenPrs:
      title: Average Open PRs
      description: Mean open PR count per entity
      type: average
      metricId: github.openPRs
    entitiesWithOpenIssues:
      title: Entities with Open Issues
      description: Count of entities with a stored open-issues value
      type: count
      metricId: jira.openIssues
    maxOpenPrs:
      title: Maximum Open PRs
      description: Maximum open PR count per entity
      type: max
      metricId: github.openPRs
    minOpenPrs:
      title: Minimum Open PRs
      description: Minimum open PR count per entity
      type: min
      metricId: github.openPRs
```
* Set environment variables for convenience
``` shell
export BACKSTAGE_URL="http://localhost:7007"
export TOKEN="your-auth-token-here"
```

### Get aggregation result when `sum` aggregation type was used (`totalOpenBugs`):
``` shell
curl -X GET "${BACKSTAGE_URL}/api/scorecard/aggregations/totalOpenBugs" \
  -H "Authorization: Bearer ${TOKEN}" | jq
```

### Get aggregation result when `average` aggregation type was used (`avgOpenPrs`):
``` shell
curl -X GET "${BACKSTAGE_URL}/api/scorecard/aggregations/avgOpenPrs" \
  -H "Authorization: Bearer ${TOKEN}" | jq
```

### Get aggregation result when `count` aggregation type was used (`entitiesWithOpenIssues`):
``` shell
curl -X GET "${BACKSTAGE_URL}/api/scorecard/aggregations/entitiesWithOpenIssues" \
  -H "Authorization: Bearer ${TOKEN}" | jq
```

### Get aggregation result when `max` aggregation type was used (`maxOpenPrs`):
``` shell
curl -X GET "${BACKSTAGE_URL}/api/scorecard/aggregations/maxOpenPrs" \
  -H "Authorization: Bearer ${TOKEN}" | jq
```

### Get aggregation result when `min` aggregation type was used (`minOpenPrs`):
``` shell
curl -X GET "${BACKSTAGE_URL}/api/scorecard/aggregations/minOpenPrs" \
  -H "Authorization: Bearer ${TOKEN}" | jq
```

### Get aggregation results when some entities lack values due to an error
The environment must contain a mix of entities: some with valid values for aggregation, and others without values to simulate an error state.

Use one of the proposed requests below to confirm that:
- Aggregation only occurs for entities with valid values.
- The total count of entities with errors is correctly populated in the calculationErrorCount attribute.

### Get aggregation results when all entities lack values due to an error
The environment must contain a all of entities without values to simulate an error state.

Use one of the proposed requests below to confirm that:
- The total count of entities with errors is correctly populated in the calculationErrorCount attribute.

### When the threshold is incorrectly configured for new aggregation types
Configure a threshold with incorrect ranges for a new aggregation type. The application is expected to throw an error upon startup.

For example:
``` yaml
    wrongThresholds:
      title: Wrong Thresholds
      description: Wrong thresholds
      type: sum
      metricId: jira.openIssues
      options:
        thresholds:
          rules:
            - key: success
              expression: '>=80'
              color: '#6bb300' # green
            - key: warning
              expression: '<20'
              color: 'rgb(224, 189, 108)' # light orange
```

### When the threshold is correctly configured for new aggregation types
A correctly configured threshold should be present in the response under `result.thresholds`. If no threshold is configured under `app-config.yaml`, the default threshold should be used.

### When a new aggregation type is configured for a boolean provider
When a metricId for a boolean metric provider is supplied, an error should be thrown when the application starts up.

For example:
``` yaml
    licenseFileExistsKpi:
      title: License File Exists KPI
      type: sum
      description: This KPI is provide information about whether the license file exists in the repository.
      metricId: filecheck.license
```


workspaces/scorecard/plugins/scorecard-backend/src/utils/aggregation/buildAggregationConfig.ts
```diff
@@ -70,11 +79,14 @@ export function buildAggregationConfig(
...
   } as AggregationConfig;
 
-  if (aggregationConfig.type === aggregationTypes.average) {
-    aggregationConfig.options = {
-      statusScores: buildStatusScores(config),
-      thresholds: buildAggregationThresholdsConfig(config),
-    };
+  if (aggregationConfig.type === aggregationTypes.weightedStatusScore) {
+    aggregationConfig.options = buildWeightedStatusScoreOptions(config);
+  } else if (isScalarAggregationType(aggregationConfig.type)) {
```
gustavolira, line 51: With `average` changing meaning from weighted-status-score to scalar mean, an un-migrated config that still has `type: average` + `options.statusScores` won't fail startup — this branch only copies `thresholds`, so the leftover `statusScores` is silently dropped before the strict Zod schema can reject it, and the KPI quietly starts returning a scalar mean with a different result shape. Should we fail startup when a scalar-type KPI still carries `options.statusScores`, with a "did you mean weightedStatusScore?" hint? Passing the raw options through to the schema would also catch it via `strictObject`.

workspaces/scorecard/plugins/scorecard/README.md
```diff
@@ -6,7 +6,7 @@ The plugin supports both the **legacy** Backstage frontend and the **New Fronten
 **Features:**
 
 - **Entity scorecard tab** — View scorecard metrics on catalog entity pages (components, websites, etc.).
-- **Scorecard homepage card** — Show aggregated KPIs on the home page (e.g. GitHub open PRs, Jira open issues). Supports **`statusGrouped`** (multi-slice pie) and **`average`** (weighted health donut) KPI types configured under **`scorecard.aggregationKPIs`**.
+- **Scorecard homepage card** — Show aggregated KPIs on the home page (e.g. GitHub open PRs, Jira open issues). The built-in card renders **`statusGrouped`** (multi-slice pie), **`weightedStatusScore`** (weighted health donut), and scalar types (`sum`, `average`, `max`, `min`, `count`).
```
gustavolira: This says the built-in card renders the scalar types, but `AggregatedMetricCard.tsx` only handles `statusGrouped` and `weightedStatusScore` — `sum`/`average`/`max`/`min`/`count` all fall through to `UnsupportedAggregationType`. If the scalar card UI is a follow-up, maybe say scalar KPIs are API-only for now, so users configuring a `sum` KPI don't file bugs about the error panel on their homepage.

workspaces/scorecard/plugins/scorecard-backend/src/validation/schemas/aggregationConfigSchemas.ts
```diff
@@ -15,7 +15,20 @@
...
+  aggregationTypes,
+  scalarAggregationTypes,
+} from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
+
+const thresholdsConfigSchema = z.object({
+  rules: z.array(
+    z.object({
+      key: z.string(),
+      expression: z.string(),
+      color: z.string(),
```
gustavolira: `color` is required here, but thresholds.md section 5 says the rules take an optional `color`, and `AggregationThresholdRule` in scorecard-common keeps it optional too (`Pick<ThresholdRule, ...>` where `color?`). So a scalar KPI rule without a color fails at startup even though config.d.ts and the docs both allow it. Should this be `z.string().optional()`, or should the docs/type change?

workspaces/scorecard/plugins/scorecard-backend/src/validation/validateAggregationConfig.ts
```diff
@@ -19,14 +19,39 @@ import type { Config } from '@backstage/config';
...
 import { buildAggregationConfig } from '../utils/buildAggregationConfig';
 import { validateThresholdsForAggregation } from '@red-hat-developer-hub/backstage-plugin-scorecard-node';
 import {
-  type AggregationConfig,
   aggregationTypes,
+  scalarAggregationTypes,
 } from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
 
-function parseAggregationConfig(config: unknown): AggregationConfig {
+function isScalarAggregationType(type: string): boolean {
```
gustavolira: Isn't this the same helper as `src/utils/isScalarAggregationType.ts` added in this PR? `isScalarAggregationRuntimeConfig` repeats the `.includes` pattern a third time. Could we import the one util everywhere — and make it a real type guard (`type is (typeof scalarAggregationTypes)[number]`) while at it?

workspaces/scorecard/plugins/scorecard-backend/src/database/DatabaseMetricValues.ts
```diff
@@ -39,11 +44,123 @@ type ReadEntityMetricsWithFiltersOptions = {
...
+
+    const aggregateRow = await aggregateQuery
+      .select(
+        this.dbClient.raw(`${aggregateExpression} as value`),
+        this.dbClient.raw('COUNT(*) as total'),
+        this.dbClient.raw('MAX(timestamp) as max_timestamp'),
+      )
+      .first();
+
+    const value = aggregateRow?.value ? Number(aggregateRow.value) : 0;
```
gustavolira: The truthiness check conflates "NULL aggregate" with "aggregate equals 0" — it works today because a numeric `0` falls through to the same `0`, but `aggregateRow?.value != null ? Number(aggregateRow.value) : 0` says what's meant and won't break if the fallback ever changes. Same for `total` on the next line.

workspaces/scorecard/plugins/scorecard-backend/src/database/DatabaseMetricValues.ts
```diff
@@ -196,6 +275,62 @@ export class DatabaseMetricValues {
...
+    metric_id: string,
+    aggregationFn: ScalarAggregationFn,
+  ): Promise<DbScalarAggregatedMetric | undefined> {
+    if (catalog_entity_refs.length === 0) {
+      return undefined;
+    }
+
+    // `value` is a JSON column. Depending on database/driver, a "missing" metric value can
+    // arrive either as SQL NULL or as JSON literal null (`CAST(value AS TEXT) = 'null'`).
+    const metricValueIsMissingExpr =
```
gustavolira: nit: same string + comment as in `readAggregatedMetricByEntityRefs` above. Since `getLatestIdsSubquery`/`normalizeTimestamp`/`mergeMaxTimestamp` were already extracted in this PR, this one could become a private class constant too.

workspaces/scorecard/plugins/scorecard-backend/src/service/mappers.ts
```diff
@@ -14,25 +14,31 @@
...
   Metric,
   aggregationTypes,
   AggregationResultByType,
-  type AggregationConfig,
+  ScalarAggregatedMetric,
 } from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
 import { DbAggregatedMetric } from '../database/types';
+import type { DbScalarAggregatedMetric } from '../database/types';
+
+function normalizeTimestamp(timestamp?: Date): string {
```
gustavolira: This shares a name with `database/utils/normalizeTimestamp.ts` but has a different contract (`Date? -> string` vs `unknown -> Date`) — easy to auto-import the wrong one. Rename to something like `toIsoTimestamp`?

workspaces/scorecard/plugins/scorecard-common/src/types/aggregation.ts
```diff
@@ -70,19 +89,29 @@ export type StatusGroupedAggregationResult = Omit<
...
+    weightedStatusScore: number;
+    weightedStatusSum: number;
+    weightedStatusMaxPossible: number;
+    aggregationChartDisplayColor: string;
+  };
+
+/**
+ * @public
+ */
+export type ScalarAggregationResult = Omit<AggregatedMetric, 'values'> & {
```
gustavolira: nit: this ends up field-for-field identical to `ScalarAggregatedMetric & { thresholds: ThresholdConfig }`, but deriving from `AggregatedMetric` means the two can drift, and the JSDoc on `entitiesConsidered`/`calculationErrorCount` only lives on one of them. Defining it in terms of `ScalarAggregatedMetric` keeps a single source of truth.

workspaces/scorecard/plugins/scorecard-backend/src/service/aggregations/AggregationService.ts
```diff
@@ -67,12 +68,14 @@ export class AggregationsService {
...
         type: aggregationTypes.statusGrouped,
         metricId: aggregationId,
-      } as AggregationConfig;
+      };
     }
 
-    return buildAggregationConfig(aggregationId, {
-      config,
-    });
+    return parseValidatedAggregationConfig(
```
gustavolira: This now runs the Zod parse plus the interval-coverage analysis in `validateThresholdsForAggregation` on every request (three router call sites), on config that startup validation already checked. Correct since config is static, but worth a comment that the re-parse is for type narrowing — or a small per-id cache if these endpoints get hot.

workspaces/scorecard/plugins/scorecard-backend/src/service/aggregations/strategies/ValueAggregationStrategy.ts
```diff
@@ -0,0 +1,76 @@
...
+  type ScalarAggregationResult,
+} from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
+import type { ScalarAggregationFn } from '../../../database/types';
+import { AggregatedMetricMapper } from '../../mappers';
+import type { AggregatedMetricLoader } from '../AggregatedMetricLoader';
+import type { AggregationOptions } from '../types';
+import { isScalarAggregationRuntimeConfig } from '../utils/aggregationRuntimeConfig';
+import type { AggregationStrategy } from './types';
+
+export class ValueAggregationStrategy implements AggregationStrategy {
```
gustavolira: nit: everything else in this family is "scalar" — `scalarAggregationTypes`, `ScalarAggregationConfig`, `readScalarAggregatedMetricByEntityRefs` — but the strategy is `ValueAggregationStrategy`. `ScalarAggregationStrategy` would keep the family greppable, and it's cheap to rename now vs after release.

workspaces/scorecard/plugins/scorecard-backend/docs/aggregation.md
```diff
@@ -34,26 +34,61 @@ KPIs under **`scorecard.aggregationKPIs`** declare a **`type`** that selects an
...
+| **`average`**             | Mean of latest metric values across owned entities                                                         | “Average open PRs per entity.”                  |
+| **`max`**                 | Maximum latest metric value across owned entities                                                          | “Worst-case / highest value in the portfolio.”  |
+| **`min`**                 | Minimum latest metric value across owned entities                                                          | “Best-case / lowest value in the portfolio.”    |
+| **`count`**               | Number of entities with a non-null latest stored value                                                     | “How many entities have data for this metric.”  |
+
+**Scalar types** (`sum`, `average`, `max`, `min`, `count`) aggregate the **numeric `value`** from each owned entity’s **latest** stored `metric_values` row for the configured **`metricId`**, instead of bucketing by threshold status. Clients can detect scalar responses by checking **`metadata.aggregationType`** against the scalar type literals (or `scalarAggregationTypes` from scorecard-common).\*\*\*\*
```
gustavolira: nit: this paragraph and the one at line 39 both open with "**Scalar types** (`sum`, ...)" and partially repeat each other — worth merging into one. There's also a stray escaped `\*\*\*\*` at the end of this line.

workspaces/scorecard/plugins/scorecard-backend/docs/aggregation.md
```diff
@@ -34,26 +34,61 @@ KPIs under **`scorecard.aggregationKPIs`** declare a **`type`** that selects an
...
+**`options.thresholds`:** **number**-style rules (same shape as metric thresholds) evaluated against **`weightedStatusScore`** on the **0–100** scale (higher = better for typical setups). The first matching rule supplies **`result.aggregationChartDisplayColor`**. If omitted from **`scorecard.aggregationKPIs`**, it stays unset in the built KPI config; **`WeightedStatusScoreAggregationStrategy`** then applies **`DEFAULT_WEIGHTED_STATUS_SCORE_KPI_RESULT_THRESHOLDS`** from [`src/constants/aggregationKPIs.ts`](../src/constants/aggregationKPIs.ts) when each aggregation runs and logs **info** that the default 0–100% health scale is used: **<30%** = error, **30–79%** = warning, **≥80%** = success. Full detail: [thresholds.md — Aggregation KPI result thresholds](./thresholds.md#4-aggregation-kpi-result-thresholds-weighted-status-score-type) and [backend README — Aggregation KPIs](../README.md#aggregation-kpis-homepage-and-get-aggregations).
```
gustavolira: nit: this anchor doesn't match the heading — "### 4. Aggregation KPI result thresholds (`weightedStatusScore` type)" slugs to `#4-aggregation-kpi-result-thresholds-weightedstatusscore-type` (the backend README uses the correct form).

workspaces/scorecard/plugins/scorecard-backend/README.md
```diff
@@ -159,19 +159,44 @@ scorecard:
...
+| `options`     | Type-specific options. For **`weightedStatusScore`**: **required** **`options.statusScores`** (status key → weight); optional **`options.thresholds`** (see [thresholds.md — Aggregation KPI result thresholds (weightedStatusScore)](./docs/thresholds.md#4-aggregation-kpi-result-thresholds-weightedstatusscore-type)). For **scalar types**: optional **`options.thresholds`** (see [thresholds.md — Aggregation KPI result thresholds (scalar types)](./docs/thresholds.md#5-aggregation-kpi-result-thresholds-scalar-types)). All scalar types require a **number** metric. Scalar responses include **`value`**, **`total`**, **`entitiesConsidered`**, **`calculationErrorCount`**, **`timestamp`**, and **`thresholds`** — see [aggregation.md — Scalar result fields](./docs/aggregation.md#scalar-result-fields). |
```
gustavolira: nit: `aggregation.md#scalar-result-fields` points to a heading that doesn't exist — there's no "Scalar result fields" section in aggregation.md (thresholds.md has the same link in section 5). Either add that section or point at the scalar response-shape bullets.

workspaces/scorecard/plugins/scorecard-backend/src/database/DatabaseMetricValues.test.ts
```diff
@@ -1356,4 +1356,253 @@ describe('DatabaseMetricValues', () => {
       },
     );
   });
+
+  describe('readScalarAggregatedMetricByEntityRefs', () => {
+    describe.each(databases.eachSupportedId())(
+      'should %p raw metric values across latest rows',
```
gustavolira: nit: this `describe.each` title renders as "should POSTGRES_13 raw metric values across latest rows", which reads oddly next to the per-`it` titles that already describe the operation. Also, the avg assertion below (`value: 5.666666666666667`) relies on exact float equality across pg and SQLite — `toBeCloseTo` would be safer.

workspaces/scorecard/plugins/scorecard-backend/src/utils/isScalarAggregationType.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { scalarAggregationTypes } from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
+import { AggregationType } from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
```
gustavolira: nit: two separate imports from the same package — can be merged into one.

workspaces/scorecard/plugins/scorecard-backend/src/service/aggregations/strategies/ValueAggregationStrategy.ts
imykhno: Updated, thank you!

workspaces/scorecard/plugins/scorecard-backend/src/utils/aggregation/buildAggregationConfigThresholds.ts
```diff
@@ -0,0 +1,35 @@
...
+export function buildAggregationConfigThresholds(
+  config: Config,
+): ThresholdConfig | undefined {
+  const thresholdsConfig = config.getOptionalConfig('options.thresholds');
+  if (thresholdsConfig) {
+    return {
+      rules: thresholdsConfig.getConfigArray('rules').map(rule => ({
+        key: rule.getString('key'),
+        expression: rule.getString('expression'),
+        color: rule.getString('color'),
```
AndrienkoAleksandr: Maybe:

```suggestion
        color: rule.getOptionalString('color'),
```

you defined 'color' like optional value here https://github.com/redhat-developer/rhdh-plugins/pull/3571/changes#diff-0dfb2adaa457d3968f237ff0d4fb56574d603531202cf19cee1fc8be7c2cb00dR28

imykhno: Yes you are right, updated to use `getOptionalString`. Thank you!

workspaces/scorecard/plugins/scorecard-backend/config.d.ts
```diff
@@ -27,30 +27,63 @@ export interface Config {
...
-        /** Metric ID for which the aggregation is calculated */
-        metricId: string;
-        /** Type-specific settings */
-        options?: {
-          /** Required under `options` when `type` is `weightedStatusScore` */
-          statusScores?: {
-            [thresholdRuleKey: string]: number;
+      [aggregationId: string]:
+        | {
+            /** Title of the aggregation */
```
dzemanov, line 32: Minor: would it be possible to have these as types?

dzemanov (commented):
I have only quickly went through the code and looks very nice! Haven't tested yet.

workspaces/scorecard/plugins/scorecard-backend/config.d.ts
imykhno: Thank you for your advice! 
However we can't do it as Backstage’s config schema loader only allows a single Config export. Extra type aliases in that file fail schema extraction (`additional symbol definitions are not allowed`). Importing a type from elsewhere also doesn’t work well with the current loader (`noResolve: true`) — the imported shape is dropped and validation becomes unconstrained.

AndrienkoAleksandr (approved):
Tested, looks good to me.

PR #3611, opened 2026-06-29
fix(theme): hover color of myGroup submenu items

rohitratannagar:
## Summary

  Replace class*= CSS attribute selectors with MUI styleOverrides.selected for
  BackstageSidebarItem and BackstageSidebarSubmenuItem.

  ### Problem
  The class*="BackstageSidebarItem-selected-" selectors relied on CSS class name substrings. In
  production, JSS can minify class names (e.g., jss4-123), breaking the match and losing selected
  styles in the cluster.

  ###Fix
  - Use styleOverrides.selected which merges at the JS level inside makeStyles — before class
  names are generated, so minification doesn't matter
  - Add BackstageSidebarSubmenuItem to the Components type (was missing)
  - Add !important to override hardcoded styles in the upstream components
## in cluster before changes
Screenshot 2026-06-29 at 1 53 28 PM

## after changes locally

https://github.com/user-attachments/assets/db5965a3-d1a4-4931-b31e-c28aa810032f

### DIFF
--- a/workspaces/theme/.changeset/fix-sidebar-selected-overrides.md
+++ b/workspaces/theme/.changeset/fix-sidebar-selected-overrides.md
@@ -1,0 +1,7 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+fix(theme): use MUI theme overrides for sidebar selected states instead of class name selectors
+
+Replaced `class*=` CSS selectors with proper `styleOverrides.selected` on `BackstageSidebarItem` and `BackstageSidebarSubmenuItem`. The previous `class*=` approach failed in production where JSS minifies class names.
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -630,9 +630,9 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
       styleOverrides: {
         drawer: {
           gap: '0.25rem',
-          borderRight: `0.5rem solid ${sidebarBackgroundColor}`,
           paddingBottom: '1.5rem',
           backgroundColor: sidebarBackgroundColor,
+          alignItems: 'stretch',
           '& hr': {
             backgroundColor: general.sidebarDividerColor,
           },
@@ -652,7 +652,7 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
       styleOverrides: {
         root: {
           borderRadius: '6px',
-          width: 'calc(100% - 0.5rem) !important',
+          width: 'calc(100% - 1rem) !important',
           marginLeft: '0.5rem !important',
           textDecorationLine: 'none',
           '&:hover, &:focus-visible': {
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -34,6 +34,7 @@ export type Components = UnifiedThemeOptions['components'] & {
   BackstageHeaderTabs?: Component;
   BackstageSidebar?: Component;
   BackstageSidebarItem?: Component;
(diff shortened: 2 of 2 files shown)


karthikjeeyar (commented):
There are a couple of observations:


1. Selected Submenu item does not have a background color in both RHDH light and dark themes. Where as backstage light/dark theme has selected background that clearly indicates the selected item. 



| Theme| Light | Dark |
| -------- | -------- | -------- |
| **Backstage**| Image| Screenshot 2026-06-30 at 10 12 38 AM |
| **RHDH**    |  Image   | Image  |


_Note: Backstage theme (Row 1) has the selected background color whereas RHDH doesn't_


---

2. Moving the mouse horizontally from the main menu (My Groups) to submenu item hides the submenu cotnainer, which makes is difficult for the user to interact with submenu. This happens only on the RHDH theme, and this issue can be also observed from the screenshots that is attached to this PR description.

https://github.com/user-attachments/assets/3ba9dffd-c067-4a72-a574-64af11a132e7

There is a padding in the main menu, that could cause the submenu to disappear when you move out of the main menu item. This padding isn't set in the default backstage theme, so it works seamlessly.

Image

---


@ShiranHi Please take a look at these issues, and let us know your thoughts on this?

rohitratannagar:
cc: @karthikjeeyar 

Selected submenu Items doesn't have background in RHDH,  because these values have been hardcoded in the backstage: [link](https://github.com/backstage/backstage/blob/master/packages/core-components/src/layout/Sidebar/SidebarSubmenuItem.tsx#L67C2-L67C29)
In RHDH one of the theme PR: [#2840 ](https://github.com/redhat-developer/rhdh-plugins/pull/2840), made the background color of selected sidebarItem and submenuItem same [here](https://github.com/redhat-developer/rhdh-plugins/pull/2840/changes#diff-8542adf66ed9b1012498d93dac98fe3e180676b6646854ba5e6669a0f835771cR637-R646). 

Since to preserve Backstage's inverted styling—where the sidebar and submenu use contrasting background colors—the selected submenu item intentionally does not have a background color, even when it is selected. This behavior is consistent with Backstage's implementation. 

And Since these changes [link](https://github.com/backstage/backstage/blob/master/packages/core-components/src/layout/Sidebar/SidebarSubmenuItem.tsx#L67C2-L67C29), does not reflect in the cluster, we do see that backstage hardcode color (#6f6f6f), it goes well with the backstage theme as their sidebar is almost same for both light and dark theme.

rohitratannagar:
And for that hardcoded color, I do have backstage pr=34741 up, and after it got merged we will have to override that color in rhdh theme palette

karthikjeeyar:
@rohitratannagar what is the status of this PR? are you looking into fixing the second issue I mentioned above?

rohitratannagar:
> @rohitratannagar what is the status of this PR? are you looking into fixing the second issue I mentioned above?

This issue does not occur in the RHDH, As the layout there has been set in a different way with it's some own additional css in [root.tsx](https://github.com/redhat-developer/rhdh/blob/main/packages/app/src/components/Root/Root.tsx) , so I doubt fixing it here may break the rhdh sidebar layout,  and we don't even add myGroup item in any of the workspace . But Still fixing it !!!

rohitratannagar:
cc: @karthikjeeyar 
for 2nd suggested fix:

Screenshot 2026-07-24 at 12 45 37 AM

Plz do verify changes in `rhdh-local | rhdh` as well

karthikjeeyar (approved):
Verified the changes in RHDH and rhdh-plugins

karthikjeeyar:
shall we merge this fix and have a follow up PR once your Backstage PR is merged ?

rohitratannagar:
> shall we merge this fix and have a follow up PR once your Backstage PR is merged ?

We can get it merged, as the upstream will be add on (with customizability of the sidebar submenu)

PR #3626, opened 2026-06-29
fix(lightspeed): stop hardcoding vector_store_ids to product docs only

rajin-kichannagari:
# Summary
The backend plugin was explicitly setting `vector_store_ids` to only the `rhdh-product-docs` vector store, which prevented BYOK RAG sources configured in `rag.inline` from being queried via inline RAG.

By removing this override, the lightspeed-stack uses its own `rag.inline` config to determine which stores to query, enabling multi-source RAG.

## Context
When BYOK knowledge sources are configured in `lightspeed-stack.yaml`, the `_fetch_byok_rag()` function in lightspeed-stack filters the request's `vector_store_ids` against `rag.inline`. Since the plugin was only sending the product docs store ID, BYOK stores were always filtered out and inline RAG was silently skipped.

- [x] A changeset describing the change and affected packages.

### DIFF
--- a/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
+++ b/workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
@@ -134,8 +134,6 @@ export async function createRouter(
     lightspeedCoreBaseUrl,
     logger,
   );
-  let lightspeed_vector_store_id: string = '';
-
   // Parse admin-configured MCP servers from app-config.
   // Only name is required; token is optional (users can provide their own via the UI).
   // URLs come from LCS (GET /v1/mcp-servers), not from app-config.
@@ -600,19 +598,6 @@ export async function createRouter(
         const { userEntityRef } = getIdentity(request);
 
         logger.info(`/v1/query receives call from user: ${userEntityRef}`);
-
-        // get the vector store id for the rhdh-product-docs vector store
-        if (lightspeed_vector_store_id === '') {
-          const vectorStores = await vectorStoresOperator.vectorStores.list();
-          lightspeed_vector_store_id =
-            vectorStores.data.find((v: any) =>
-              v.name.startsWith('rhdh-product-docs'),
-            )?.id || '';
-        }
-
-        if (lightspeed_vector_store_id !== '') {
-          request.body.vector_store_ids = [lightspeed_vector_store_id];
-        }
 
         const userQueryParam = `user_id=${encodeURIComponent(userEntityRef)}`;
         request.body.media_type = 'application/json'; // set media_type to receive start and end event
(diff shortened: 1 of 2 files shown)


workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -601,19 +599,6 @@ export async function createRouter(
...
-          lightspeed_vector_store_id =
-            vectorStores.data.find((v: any) =>
-              v.name.startsWith('rhdh-product-docs'),
-            )?.id || '';
-        }
-
-        if (lightspeed_vector_store_id !== '') {
-          request.body.vector_store_ids = [lightspeed_vector_store_id];
-        }
-
```
michael-valdron, line 616: @maysunfaisal @rajin-kichannagari Do we still need to get the vector store ID from the vector stores this way if the platform engineer does not use BYOK? If so, we might want to add some additional logic here to determine if this is a BYOK situation or not.

JslYoon: I'm pretty sure we still need this explicit vector_store setup or otherwise Developer LIghtspeed tool will also query from Notebooks vector_store.

JslYoon (commented):
I believe we deliberately disabled `rag.inline` and only enabled `tool.rag` because the two conflicted and caused an infinite loop.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
maysunfaisal: @JslYoon is that why it was introduced in the first place? We also need to remove this hardcoded vector store dependency because OKP also cannot function if we are calling LCORE with hardcoded vector store names 🤔

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -601,19 +599,6 @@ export async function createRouter(
 
         logger.info(`/v1/query receives call from user: ${userEntityRef}`);
 
-        // get the vector store id for the rhdh-product-docs vector store
```
maysunfaisal, line 604: According to Claude,

> The test fixture at __fixtures__/lcsHandlers.ts:281-282 mocks the vector stores endpoint returning rhdh-product-docs. Since the PR removes the code that calls that endpoint to look up the vector store
  ID, this mock handler will just go unused — it won't break any tests, but it's now dead fixture code.

maysunfaisal:
@rajin-kichannagari This PR is still valid and addresses a concern for both BYOK and OKP. I shared some insights on the Slack thread. Maybe we can chat more about this on the Parking Lot or we can discuss on Slack.

maysunfaisal (approved):
lgtm

PR #3629, opened 2026-06-29
Rhidp 14314/attachment type model validation

JslYoon:

- Add `/v1/validate-model-vision` endpoint that probes model vision
   capability by sending a minimal test JPEG to LCS `/v1/responses`
   - Add `validateAttachmentsForModel` middleware on `/v1/query` to reject
   image attachments for non-vision models
   - Add `ModelCapabilitiesCache` singleton to cache vision capability
   results per model
   - Remove stray `console.log` debug statement from notebooks router
   - Unit tests for the cache and integration tests for both the endpoint and
   query validation

https://redhat.atlassian.net/browse/RHIDP-14314
https://redhat.atlassian.net/browse/RHIDP-15365


workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
```diff
@@ -730,6 +736,74 @@ export async function createRouter(
     },
   );
 
+  router.post(
+    '/v1/validate-model-vision',
+    requirePermission(lightspeedChatReadPermission),
+    async (request, response) => {
+      const { model, provider } = request.body;
```
its-mitesh-kumar, line 866: @JslYoon Can you check if I can send array of models for supportsVision check ? Lets check about latency

JslYoon: I'm worried that since if we were to send in an array of models and providers, not only the latency will increase, but also since I'll be looping `/streaming_query` requests to lightspeed, which if interrupted, could cause data race problems

its-mitesh-kumar: Then lets lets keep each call scoped to a single model.

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -730,6 +757,74 @@ export async function createRouter(
...
+              ],
+              tool_choice: 'none',
+              temperature: 0,
+              store: false,
+              stream: false,
+            }),
+          },
+        );
+        const supportsVision = testResponse.ok;
+        ModelCapabilitiesCache.set(model, supportsVision);
```
its-mitesh-kumar: Shouldn't we also include `provider` in the  Key for cache.

its-mitesh-kumar (changes requested):

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
JslYoon: Updated!

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -730,6 +757,76 @@ export async function createRouter(
...
+        ModelCapabilitiesCache.set(cacheKey, supportsVision);
+
+        logger.info(
+          `Vision test for ${cacheKey}: ${supportsVision ? 'PASS' : 'FAIL'}`,
+        );
+
+        response.json({ model, provider, supportsVision });
+      } catch (error) {
+        logger.error(`Vision test error for ${cacheKey}:`, error);
+        ModelCapabilitiesCache.set(cacheKey, false);
```
its-mitesh-kumar: I think we should not cache the false in case of errors, it can happen due to transient failure which will result into permanently mark the model as non-vision-capable.

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
```diff
@@ -730,6 +757,76 @@ export async function createRouter(
...
+              ],
+              tool_choice: 'none',
+              temperature: 0,
+              store: false,
+              stream: false,
+            }),
+          },
+        );
+        const supportsVision = testResponse.ok;
+        ModelCapabilitiesCache.set(cacheKey, supportsVision);
```
its-mitesh-kumar: Should we consider adding TTL for it. I know once a model has vision support it might not change ?

its-mitesh-kumar (changes requested):

workspaces/lightspeed/plugins/lightspeed-backend/src/service/router.ts
```diff
@@ -728,6 +734,27 @@ export async function createRouter(
...
+            content: string;
+          }>) {
+            if (att.attachment_type !== 'image') {
+              contextParts.push(att.content);
+            }
+          }
+          if (contextParts.length > 0) {
+            request.body.query = `${contextParts.join('\n')}\n\n${request.body.query}`;
+          }
+          delete request.body.attachments;
```
Jdubrick: This writes all the non images to `contextParts` then deletes all the images? Why?

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
```diff
@@ -860,6 +887,75 @@ export async function createRouter(
     },
   );
 
+  router.post(
```
Jdubrick, line 861: We should make sure rate limiting is added to all endpoints for consistency

JslYoon:
Waiting on https://github.com/lightspeed-core/lightspeed-stack/pull/2108 to be merged first as image support for attachments will be supported on lightspeed core

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
johnmcollier: Hey @JslYoon did you address this? I don't see rate limiting added

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/validation.ts
```diff
@@ -103,3 +101,51 @@ export const validateLoadHistoryRequest = (
 
   return next();
 };
+
+export const validateAttachmentsForModel = (
+  req: Request,
+  res: Response,
+  next: NextFunction,
+) => {
```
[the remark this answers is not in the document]
johnmcollier: @JslYoon This might be worth addressing, could this cause problems when the Backstage pod restarts?

johnmcollier (changes requested):
Left a couple comments. 

I also don't see a changeset, so you'll want to add one.

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
```diff
@@ -848,6 +859,72 @@ export async function createRouter(
...
+    '/v1/validate-model-vision',
+    requirePermission(lightspeedChatReadPermission),
+    async (request, response) => {
+      const { model, provider } = request.body;
+      const cacheKey = `${provider}/${model}`;
```
Jdubrick, line 878: Should we check if these are set and return a 400 if they aren't?

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/router.ts
```diff
@@ -848,6 +859,72 @@ export async function createRouter(
...
+                ],
+              },
+            ],
+            tool_choice: 'none',
+            temperature: 0,
+            store: false,
+            stream: false,
+          }),
+        });
+        const supportsVision = testResponse.ok;
```
Jdubrick: What if the response we get back is a `5xx` because LCORE is down/struggling, and not because vision isn't supported?  We'd cache `false` in this case and block that model from being used

JslYoon: If model truely doesn't accept images, LCORE will return `422`. I can make it so that if we only set the cache either if it succeeds or LCORE returns `422`

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/validation.ts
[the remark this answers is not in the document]
JslYoon: If the pod restarts, cache will be wiped off and we requery, I believe UI/UX wants so that we validate all possible models before actually querying to display.

johnmcollier (approved):

its-mitesh-kumar (approved):
/lgtm

Jdubrick (approved):
lgtm

PR #3635, opened 2026-06-30
ci: add nightly CI workflow that runs all workspaces

christoph-jerolimov:

Adds a scheduled nightly workflow that runs the full CI suite across all workspaces with Node 22 and 24, plus a helper script to list workspaces with optional filtering via workflow_dispatch input.

Tested this in my fork.

### With one workspace defined

You can run specific tests for the main branch from the GitHub UI:

image

Running job when one workspace is passed

image

Finished

image

### No workspace passed

Running job when no workspace is passed:

image

Finished with errors:

image

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -1,0 +1,195 @@
+name: CI Nightly
+
+on:
+  # Run every night at 00:00 UTC
+  schedule:
+    - cron: '0 0 * * *'
+  # Allow manual runs, optionally scoped to a subset of workspaces
+  workflow_dispatch:
+    inputs:
+      workspaces:
+        description: 'Optional JSON string array of workspaces to run, e.g. ["global-header","theme"]. When empty all workspaces run.'
+        required: false
+        type: string
+
+jobs:
+  list-workspaces:
+    name: List workspaces
+    runs-on: ubuntu-latest
+    outputs:
+      workspaces: ${{ steps.list-workspaces.outputs.workspaces }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Node
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: 24
+          registry-url: https://registry.npmjs.org/ # Needed for auth
+
+      - name: List workspaces
+        id: list-workspaces
+        run: node ./scripts/ci/list-all-workspaces.js
+        env:
+          WORKSPACES: ${{ inputs.workspaces }}
+
+  ci:
+    name: Workspace ${{ matrix.workspace }}, CI step for node ${{ matrix.node-version }}
+    runs-on: ubuntu-latest
+    needs: list-workspaces
+    strategy:
+      matrix:
+        workspace: ${{ fromJSON(needs.list-workspaces.outputs.workspaces) }}
+        node-version: [22, 24]
+      fail-fast: false
+    defaults:
+      run:
+        working-directory: ./workspaces/${{ matrix.workspace }}
+
+    env:
+      CI: true
+      NODE_OPTIONS: --max-old-space-size=8192
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4
+
+      - name: Set up Node ${{ matrix.node-version }}
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: https://registry.npmjs.org/ # Needed for auth
+
+      - name: yarn install
+        run: yarn install --immutable
+
+      - name: check for missing repo fixes
+        run: yarn fix --check
+
+      - name: validate config
+        if: ${{ hashFiles(format('workspaces/{0}/app-config.yaml', matrix.workspace)) != '' }}
+        run: yarn backstage-cli config:check --lax
+
+      - name: type checking and declarations
+        run: yarn tsc:full
+
+      - name: prettier
+        run: yarn prettier:check
+
+      - name: read knip-reports flag from bcp.json
+        id: bcp
+        run: |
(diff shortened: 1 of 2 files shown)


scripts/ci/list-all-workspaces.js
```diff
@@ -0,0 +1,98 @@
...
+
+  return parsed;
+}
+
+function listAllWorkspaces(workspacesDir) {
+  return fs
+    .readdirSync(workspacesDir, { withFileTypes: true })
+    .filter(dirent => dirent.isDirectory())
+    .map(dirent => dirent.name)
+    .filter(name =>
```
gustavolira, line 59: This only filters on "has a package.json", but the repo already has `scripts/list-workspaces.js` exporting `listWorkspaces()`, which excludes `EXCLUDED_WORKSPACES = ['noop', 'repo-tools']`. Both of those do have a package.json, so the nightly matrix will run full CI (x Node 22 and 24) on the `noop` placeholder and on `repo-tools`. Should we reuse `listWorkspaces()` here, or at least copy its exclusion list, so the two listers don't diverge?

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,195 @@
...
+        run: yarn build:knip-reports --ci
+
+      - name: check api reports and generate API reference
+        run: yarn build:api-reports:only --ci
+
+      - name: build all packages
+        run: yarn backstage-cli repo build --all
+
+      - name: lint
+        run: yarn backstage-cli repo lint --since origin/main
```
gustavolira, line 105: Does this actually lint anything on the scheduled run? On a `schedule` trigger we check out the default branch, so `HEAD == origin/main` and `repo lint --since origin/main` finds no changed packages. Since the point of the nightly is to exercise everything, shouldn't this be `--all` (or otherwise unscoped) here?

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,195 @@
...
+        id: tests
+        env:
+          JEST_JUNIT_OUTPUT_DIR: ${{ runner.temp }}/test-results/${{ matrix.workspace }}
+          JEST_JUNIT_CLASSNAME: '{filepath}'
+          JEST_JUNIT_UNIQUE_OUTPUT_NAME: 'true'
+        run: yarn test:all --maxWorkers=3 --reporters=default --reporters=${{ runner.temp }}/jest-junit/node_modules/jest-junit
+
+      - name: Upload coverage to Codecov
+        if: ${{ !cancelled() && steps.tests.outcome != 'skipped' }}
+        uses: codecov/codecov-action@57e3a136b779b570ffcdbf80b3bdc90e7fab3de2 # v6.0.0
```
gustavolira, line 125: This pins `codecov-action` to `v6.0.0`, but `ci.yml` is already on `fb8b3582... # v6.0.2`. New file shouldn't land on an older pin — worth matching `ci.yml` so they don't drift from day one. (Good example of why duplicating the whole job is risky.)

scripts/ci/list-all-workspaces.js
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Lists the workspaces that should run in the nightly CI workflow.
+ *
+ * When the WORKSPACES environment variable is set it is parsed as a JSON
+ * string array and used as an explicit allow list. Otherwise every workspace
+ * that contains a package.json is returned.
+ */
+function parseRequestedWorkspaces() {
+  const raw = process.env.WORKSPACES?.trim();
+  if (!raw) {
```
gustavolira, line 29: The dispatch description says "when empty all workspaces run," which holds for an empty string. But an empty JSON array `[]` is truthy, so it flows through as an explicit empty allow-list: the matrix ends up empty, `ci`/`verify` get skipped, and the `result` job's `contains(needs.*.result, 'skipped')` then fails the run. Should we treat `[]` the same as unset (or reject it with a clear error)?

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,195 @@
...
+    - cron: '0 0 * * *'
+  # Allow manual runs, optionally scoped to a subset of workspaces
+  workflow_dispatch:
+    inputs:
+      workspaces:
+        description: 'Optional JSON string array of workspaces to run, e.g. ["global-header","theme"]. When empty all workspaces run.'
+        required: false
+        type: string
+
+jobs:
```
gustavolira, line 15: Two top-level keys worth adding here: an explicit `permissions:` block (`contents: read` baseline) like most sibling workflows already declare, and a `concurrency:` group so a manual `workflow_dispatch` during the scheduled run doesn't stack two full all-workspace matrices in parallel.

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,195 @@
...
+          WORKSPACES: ${{ inputs.workspaces }}
+
+  ci:
+    name: Workspace ${{ matrix.workspace }}, CI step for node ${{ matrix.node-version }}
+    runs-on: ubuntu-latest
+    needs: list-workspaces
+    strategy:
+      matrix:
+        workspace: ${{ fromJSON(needs.list-workspaces.outputs.workspaces) }}
+        node-version: [22, 24]
```
gustavolira, line 44: Everything from "Set up Node" down through `verify` and `result` is a near-verbatim copy of `ci.yml` (~150 lines). It'll rot — the codecov pin below is already an instance of the two drifting. Could we extract the per-workspace steps into a `workflow_call` reusable workflow that takes a `workspaces` JSON input, and have both `ci.yml` and this file call it? Bigger refactor since it touches the existing CI, but the two are basically identical now so it pays off fast.

.github/workflows/ci-nightly.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      - name: publish check
+        run: yarn backstage-cli repo fix --check --publish
+
+      - name: Install jest-junit reporter
+        run: |
+          npm install jest-junit@17.0.0 --ignore-scripts --prefix ${{ runner.temp }}/jest-junit
+          mkdir -p ${{ runner.temp }}/test-results/${{ matrix.workspace }}
+
+      - name: Test changed packages
```
gustavolira, line 115: nit: "Test changed packages" is inherited from `ci.yml`, but in a workflow whose whole job is running everything (`yarn test:all`) the name reads wrong. "Test all packages" would be clearer in the logs here.

PR #3636, opened 2026-06-30
ci: add reusable workflow to build dynamic plugins container images

christoph-jerolimov:

Introduce build-container-image.yml that uses rhdh-cli to package and push workspace images to quay.io.

Wire it into CI for a per-PR build, and the release workflow for next and latest tags.

The per-PR images has a 2w expiry, next and images a 4w expiry.

**The images are build per workspace, not separated per package. The workspace is also added as a prefix to the tag, so that we don't need a quay repo per workspace. Feedback is very welcome for this. Both is different to the overlay build process but simplifies this, because rhdh-cli just builds all packages.**

:warning: Untested, and it includes a high risk that this could break the CI or release pipelines.

### DIFF
--- a/.github/workflows/build-container-image.yml
+++ b/.github/workflows/build-container-image.yml
@@ -1,0 +1,124 @@
+name: Build container image
+
+# Reusable workflow that builds a Red Hat Developer Hub dynamic plugins
+# container image for a single workspace using the rhdh-cli.
+# Adds a quay expiration label and pushes the image to the container registry.
+
+on:
+  workflow_call:
+    inputs:
+      workspace:
+        description: Name of the workspace to build an image for
+        required: true
+        type: string
+      tags:
+        description: >-
+          Space separated list of image tags (the part after the colon).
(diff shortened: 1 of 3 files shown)


.github/workflows/build-container-image.yml
```diff
@@ -0,0 +1,124 @@
...
+            podman tag "${PRIMARY_IMAGE}" "${REGISTRY}:${tag}"
+          done
+
+          # Hand the resolved tags over to the push step.
+          echo "RESOLVED_TAGS=${RESOLVED_TAGS}" >> "${GITHUB_ENV}"
+
+      - name: Push image to registry
+        env:
+          REGISTRY: ${{ inputs.registry }}
+          QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
```
kadel, line 112: I think that this will not work.
This workflow is triggered from `ci.yaml` which is triggered by `on: pull_request`, this means that secrets won't be injected into this workflow.

PR #3641, opened 2026-06-30
test(orchestrator): add scaffolder backend module test coverage

rostalan:

- Unit test coverage for scaffolder-backend-module-orchestrator
- Adds unit tests for utils, getWorkflowParams, and runWorkflow actions. 
- Fixes undefined token fallback in utils and dry-run early return in runWorkflow.

Closes: https://redhat.atlassian.net/browse/RHIDP-9861


lholmquist (commented):
Overall this is looking really good.

I think you need to generate a changeset.

No need to change this, but i was wondering why the change here https://github.com/redhat-developer/rhdh-plugins/pull/3641/changes#diff-029a0f4e7d93973a0df6d5422798d5df8f82ff865a46151e6d6bfa762c3662feL76-R84

lholmquist (approved):

PR #3643, opened 2026-06-30
fix(boost): adjust specs for changes need for a working boost backend demo

gabemontero:

These are spec-only adjustments to cover gaps in the code originally produced by fullsend that I needed to make in order to successfully run the boost-backend against kagenti and llamastack

Once this merges, I will run fullsend code agent and see how close the generated code maps with the locally developed branch I curated to successfully pull data from both llamastack and kagenti

- [/] Screenshots attached (for UI changes)


gabemontero:
Good catch — fixed in e47f16310c.

`boost-responses-api-toolkit` depends on `boost-common`, which transitively pulls in `@backstage/plugin-permission-common`, so "zero Backstage dependencies" was inaccurate. Updated all three spots in `boost-context.md` and `platform-operations-deployment.md` to say "minimal Backstage dependencies via boost-common" instead. The claim for `boost-toolscope` is left as-is since that package truly has zero runtime dependencies.

PR #3645, opened 2026-06-30
feat(boost): scaffold AI Catalog frontend plugin and dev app

rohitkrai03:
## Fixes
- https://redhat.atlassian.net/browse/RHIDP-15165

## Summary
- Adds frontend architecture document.
- Adds specs for the Ai catalog work.
- Scaffolds the `@red-hat-developer-hub/backstage-plugin-boost` NFS-only frontend plugin with PageBlueprint, EntityCardBlueprint/EntityContentBlueprint stubs, `isAiAsset` condition filter, `useAiAssets` hook, i18n scaffold, and unit tests
- Adds dev app (`packages/app`) and dev backend (`packages/backend`) following the orchestrator/quickstart NFS pattern with `app.packages: all` auto-discovery
- Creates sample catalog fixtures for all 7 AI asset entity types and `app-config.yaml` with route redirect
- Adds `@red-hat-developer-hub/backstage-plugin-boost` to `pluginPackages` in all boost-family packages

## Test plan

- [x] `yarn tsc --noEmit --skipLibCheck` passes (clean compilation)
- [x] `yarn test` passes (436 tests, 45 suites including 2 new boost plugin tests)
- [x] `yarn start` launches dev app — frontend compiles, backend starts, catalog ingests fixtures
- [x] `/ai-catalog` page renders the placeholder AiCatalogPage
- [x] AI Catalog sidebar item visible with distinct icon
- [x] ESLint and Prettier pass (verified via pre-commit hook)


workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  return filterSets;
+}
+
+/**
+ * Wraps catalogApiRef to query AI asset entities with optional filtering.
+ * Uses client-side filtering for search, tags, and category (post-fetch).
+ */
+export function useAiAssets(filters: AiAssetFilters = {}): UseAiAssetsResult {
```
mareklibra, line 120: This will become important hook, worth of having comprehensive unit tests

workspaces/boost/plugins/boost/src/utils/isAiAsset.ts
```diff
@@ -0,0 +1,50 @@
...
+};
+
+/** Condition filter for NFS Blueprints — returns true for AI asset entities. */
+export function isAiAsset(entity: {
+  kind: string;
+  spec?: { type?: string };
+}): boolean {
+  const kind = entity.kind.toLowerCase();
+  const specType = entity.spec?.type;
+  if (!specType) {
```
mareklibra: This is inconsistent:

`isAiAsset` treats any `AiResource` without `spec.type` as an AI asset, but `buildCatalogFilter()` in `useAiAssets.ts` only ever requests: `['AiResource', ['skill', 'rule']]`

It should be unified and the architecture doc be updated based for that.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
```diff
@@ -0,0 +1,170 @@
...
+
+/**
+ * Builds catalog entity filter expressions for all AI asset kind/type
+ * combinations. Returns an OR query across all valid pairs.
+ */
+function buildCatalogFilter(
+  filters: AiAssetFilters,
+): Record<string, string | string[]>[] {
+  const kindTypeMap: [string, string[]][] = [
+    ['AiResource', ['skill', 'rule']],
```
mareklibra: so this will exclude what isAiAsset() would accept.

workspaces/boost/plugins/boost/src/utils/isAiAsset.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const AI_ASSET_SPEC_TYPES: Record<string, Set<string>> = {
+  airesource: new Set(['skill', 'rule']),
+  api: new Set(['mcp-server']),
+  component: new Set(['ai-agent']),
+  resource: new Set(['ai-model', 'ai-tool', 'vector-store']),
+};
+
+/** Condition filter for NFS Blueprints — returns true for AI asset entities. */
+export function isAiAsset(entity: {
```
mareklibra: For future stability, it would be better to use Backstage's `Entity` as the type

workspaces/boost/plugins/boost/package.json
```diff
@@ -0,0 +1,102 @@
...
+    "@testing-library/jest-dom": "^6.0.0",
+    "@testing-library/react": "^16.0.0",
+    "@types/react": "^18.2.58",
+    "@types/react-dom": "^18.2.19",
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0",
+    "react-router-dom": "^6.3.0"
+  },
+  "maintainers": [
+    "@rohitkrai03"
```
mareklibra, line 99: Shouldn't this be workspace-level?

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
```diff
@@ -0,0 +1,170 @@
...
+  const { search, owner } = filters;
+  const categoryKey = filters.category?.join(',') ?? '';
+  const lifecycleKey = filters.lifecycle?.join(',') ?? '';
+  const tagsKey = filters.tags?.join(',') ?? '';
+
+  // Only lifecycle and owner are passed to the catalog API filter.
+  // Other filter dimensions (search, category, tags) are applied client-side.
+  const catalogFilter = useMemo(
+    () => buildCatalogFilter(filters),
+    [lifecycleKey, owner], // eslint-disable-line react-hooks/exhaustive-deps
```
mareklibra: A comment properly explaining why just this subset is used, will be useful in the future.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
```diff
@@ -0,0 +1,170 @@
...
+
+  useEffect(() => {
+    let cancelled = false;
+
+    async function fetchEntities() {
+      setLoading(true);
+      setError(undefined);
+
+      try {
+        const response = await catalogApi.getEntities({
```
mareklibra: This is expensive operation which is triggered even for FE-only filtering.

search, categoryKey, and tagsKey are in the useEffect dependency array but are only used for client-side post-filtering - they don't change the catalogFilter sent to the API. Every search keystroke or tag toggle triggers a redundant catalogApi.getEntities() call returning identical data.

mareklibra (dismissed):
There are stale references to the `boost-frontend` instead of just `boost` (agents.md and maybe other files). This can cause confusion later. #3640 has this fixed.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Yes, agreed.

workspaces/boost/plugins/boost/package.json
rohitkrai03: Each plugin can have different sets of maintainers actually. Although this list here is not used anywhere. What we have in CODEOWNERS file is more important.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Yes, this is an issue which I identified during the follow AI catalog implementation. I've fixed it already but now I am thinking if we really need this hook here or we can just drop it from this PR and introduce it in the follow up PR.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Pushed a much simplified hook that also fixes your other comments.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Fixed.

workspaces/boost/plugins/boost/src/utils/isAiAsset.ts
rohitkrai03: Fixed.

workspaces/boost/plugins/boost/src/utils/isAiAsset.ts
rohitkrai03: Done.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Pushed the fix.

workspaces/boost/plugins/boost/src/hooks/useAiAssets.ts
rohitkrai03: Removed the disable line. Its not needed in the newer hook.

workspaces/boost/app-config.yaml
```diff
@@ -9,28 +16,49 @@ backend:
...
         - System
-        - Domain
+        - Group
+        - Resource
         - Location
         - Template
-        - Group
-        - User
+        - API
+        - AiResource
```
gabemontero, line 56: yeah @rohitkrai03 @mareklibra once https://github.com/redhat-developer/rhdh-plugins/pull/3661 merges I'll pivot toward updating catalog storage in the backend to use AIResource

this will include cross ref's with @johnmcollier @maysunfaisal on other 2.1 content ... see https://docs.google.com/document/d/1_HOQZ5f1nj38io7tGiNBibemRG3kRbTk9O0Z21O9MG8/edit?tab=t.0#heading=h.xkf2mswwisx8 for some of the initial discussion on this, though more discussion among the rhdh ai team will be occurring soon

gabemontero (approved):

gabemontero (approved):

PR #3627, opened 2026-06-29
fix(lightspeed): minor UI fix

ciiay:
## Hey, I just made a Pull Request!

Fix duplicate caret icons in the message bar model selector dropdown.

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ ] Added or Updated documentation
- [ ] Tests for new functionality and regression tests for bug fixes
- [x] Screenshots attached (for UI changes)

Before:
image

After:
image

PR #3672, opened 2026-07-03
feat(lightspeed): add Sources chip + popover for notebook context

its-mitesh-kumar:
## Description
Replace the inline SourcesCard in the Notebook view with a compact "N Sources" chip button that opens a PatternFly Popover displaying source documents. Each source is shown with a colored FileTypeIcon badge and optional description summary. This change is scoped exclusively to the Notebook context — the general chat view retains its original inline sources display.

Also consolidates source document handling during streaming to use the shared `transformDocumentsToSources` utility for consistency, and adds i18n translations for all 5 supported languages (Japanese, French, German, Spanish, Italian).

## Fixed
- https://redhat.atlassian.net/browse/RHIDP-14655

## UI after changes

https://github.com/user-attachments/assets/02013e4b-0b57-4681-8320-7037de973a28

https://github.com/user-attachments/assets/82a98908-cedf-4454-83df-dbcb9fd12287

https://github.com/user-attachments/assets/6d6edb25-1e5e-440a-9345-88a184ab0ac0

https://github.com/user-attachments/assets/42ee67c5-39f3-48d9-9306-001b8015b6b8

## ✔️ Checklist


its-mitesh-kumar:
@HusneShabbir Lets firstly wait for @ShiranHi feedback, then only go for go for testing it. There are differences wrt to prototype like we are using [PF popups](https://www.patternfly.org/components/popover/) and tried not overriding it much to look exactly as our prototype. Secondly in RHDH 2.1 we will not have the document summary, so we will be only showing documents name, that's why I added the lines between each doscuments.

ShiranHi:
Thank you @its-mitesh-kumar, looks good to me. I understand that we are unable to remove the border line between each result or change the 'i' icon from purple to gray. If that’s a constraint, I am completely fine with this implementation. Thank you!

its-mitesh-kumar:
@ShiranHi 
>  we are unable to remove the border line between each result or change 

We can remove the border line between each result. I did introduced it knowingly to look better. In 2.1 we will be only having only document name.  Let me know your preference.

https://github.com/user-attachments/assets/3d452563-60be-42ac-8eba-2f1967ec2489


> change the 'i' icon from purple to gray.

We can change the 'i' icon from purple to gray. I made it purple since to me its  look better particularly in light theme. Let me know your preference.

its-mitesh-kumar:
@ShiranHi Made both changes now
## Light Theme

S_ 2026-07-06 at 8 04 21 PM

## Dark Theme

S_ 2026-07-06 at 8 04 40 PM

Cc: @HusneShabbir

HusneShabbir (commented):
Implementation looks good.

Just one question: is there a reason we're using a different close button icon here? Wouldn't it be better to keep it consistent with the close icon used elsewhere?

Image

.

Image

its-mitesh-kumar:
> Implementation looks good.
> 
> Just one question: is there a reason we're using a different close button icon here? Wouldn't it be better to keep it consistent with the close icon used elsewhere?
> 
> Image
> .
> 
> Image

@HusneShabbir this close icon is coming from PF popover. So lets not try to override it.

HusneShabbir:
cc: @ShiranHi

ciiay:
>@HusneShabbir this close icon is coming from PF popover. So lets not try to override it.

@its-mitesh-kumar do we have a bug ticket for this issue? It affects MCP config modal icons too, the x icon and key icon. We need to find a universal fix for this.

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
```diff
@@ -0,0 +1,196 @@
...
+export const SourcesChipModal = ({ sources }: SourcesChipModalProps) => {
+  const classes = useStyles();
+  const { t } = useTranslation();
+
+  const count = sources.sources?.length ?? 0;
+  if (count === 0) return null;
+
+  const handleSourceClick = (link: string, isExternal?: boolean) => {
+    if (link) {
+      window.open(
```
ciiay: nit: Consider using `Button component="a" href={...} target="_blank" rel="noopener noreferrer"` to match PatternFly `SourcesCardBase` behavior, instead of `window.open` on a button. This gives better a11y and avoids javascript: URI risk from backend data.

workspaces/lightspeed/plugins/lightspeed/src/components/LightspeedChatBox.tsx
```diff
@@ -114,6 +115,8 @@ type LightspeedChatBoxProps = {
   conversationId: string;
   isStreaming: boolean;
   displayMode?: ChatbotDisplayMode;
+  /** When true, sources are shown as a compact chip + modal instead of the inline SourcesCard. */
+  useSourcesChipModal?: boolean;
```
ciiay: nit: `useSourcesChipModal` reads like a React hook. Consider renaming to `showSourcesChipPopover` since the UI is a Popover, not a Modal.

ciiay (commented):
Thanks for the pr 👏  Mostly looks good to me, only left a couple nit comments. Also it might be worth adding a `LightspeedChatBox` test with `useSourcesChipModal` and updating the Message mock to render `extraContent.endContent`.

its-mitesh-kumar:
> > @HusneShabbir this close icon is coming from PF popover. So lets not try to override it.
> 
> @its-mitesh-kumar do we have a bug ticket for this issue? It affects MCP config modal icons too, the x icon and key icon. We need to find a universal fix for this.

@ciiay Since its PF close icon only. So lets use this only. At other places if we are importing close icon from MUI or somewhere else we can change that and use PF close icon throughout lightspeed.

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
its-mitesh-kumar: Removing this as we don't need it at all. Notebook can only reference to the documents uploaded.

workspaces/lightspeed/plugins/lightspeed/src/components/LightspeedChatBox.tsx
its-mitesh-kumar: Renamed.

ciiay:
> @ciiay Since its PF close icon only. So lets use this only. At other places if we are importing close icon from MUI or somewhere else we can change that and use PF close icon throughout lightspeed.

Not only the close icon. It happens to key icon on MCP config modal as well. Before the issue gets resolved in upstream, what can we do from our level? Do you think we need a jira ticket to track this issue?

ciiay (approved):
Thanks for the pr 👏 

/lgtm

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+    '& .pf-v6-c-popover__title-text': {
+      margin: 0,
+    },
+    '& .pf-v6-c-popover__header': {
+      paddingBlockEnd: theme.spacing(2),
+    },
+    '& .pf-v6-c-popover__body': {
+      paddingBlockStart: 0,
+    },
+  },
```
karthikjeeyar, line 49: This is a fragile code could break on PF upgrade, fighting PF internals. 

 See if Patternfly `Popover` has its own spacing props for header to avoid these overrides.

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
```diff
@@ -0,0 +1,160 @@
...
+                  key={`${source.title}-${index}`}
+                  className={classes.sourceItem}
+                >
+                  <FileTypeIcon fileName={title} />
+                  <div className={classes.sourceContent}>
+                    <div className={classes.sourceTitle}>{title}</div>
+                    {source.body && (
+                      <div className={classes.sourceBody}>{source.body}</div>
+                    )}
+                  </div>
```
karthikjeeyar: This SourcesChipModal, uses raw HTML elements with custom styles over Patternfly standard components.


- Instead of `ul, li` , see if you can use PF `List` , `ListItem`

- Instead of using raw `div`, see if you can use PF `HelperText` or `Content` component or any alternative.

```
     <div className={classes.infoRow}>
        <InfoCircleIcon className={classes.infoIcon} />
        <span>{t('sources.modal.description')}</span>
      </div>

```


The value of using PatternFly is to get tested, maintained and reusable components that looks consistent across products and handling accessibility automatically.    Using PF's Popover for the container but raw HTML for everything inside it means you get the outer accessibility shell but lose it for the actual content users interact with. 


I understand you are trying to match with the design with custom components, but if using PF components means we get slightly different spacing, I think that is acceptable rather than missing accessibility and maintenance overhead. @ShiranHi let me know if this  is reasonable. Thanks

karthikjeeyar (commented):
Haven't tested yet, I have some code style comments.

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
ShiranHi: Yes, please stick to the standard PatternFly icons and components for this plugin. If we feel there's room for improvement, we can open an issue with the PatternFly team directly, but please avoid creating any custom icon, styles or components for now. Thanks!

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
its-mitesh-kumar: @karthikjeeyar, Checked the[ PF v6 Popover API ](https://www.patternfly.org/components/popover)— the only spacing-related prop is `hasNoPadding`, which removes **all** content padding (too
  aggressive). There are no `headerProps`, `bodyProps`, or individual padding props for header/body sections.

  Since there's no built-in way to control these individually, the CSS class overrides are the only option for this level of granularity. I've scoped them to the component's root class (`.sourcesPopover`) to minimize blast radius. If PF adds granular spacing props in a future version, we can drop these overrides then.

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
its-mitesh-kumar: @ShiranHi / @karthikjeeyar  After using `HelperText`, `List` , Its look like this

S_ 2026-07-08 at 2 59 23 PM
S_ 2026-07-08 at 2 59 40 PM

ShiranHi: Looks good to me! Can we truncate long names to a single line with an ellipsis (...)?

karthikjeeyar:
As far as the `X` icon issue, I can see there two icons rendered at the same time from PF, `pf-v6-icon-rh-ui` and `pf-v6-icon-default`, these two icons stacked on each other. Check why two icons are rendering at the same time or check with PF team to solve this issue in this lightspeed plugin.

image

image

its-mitesh-kumar:
> As far as the `X` icon issue, I can see there two icons rendered at the same time from PF, `pf-v6-icon-rh-ui` and `pf-v6-icon-default`, these two icons stacked on each other. Check why two icons are rendering at the same time or check with PF team to solve this issue in this lightspeed plugin.
> 

@karthikjeeyar I have commented my some findings here : https://github.com/redhat-developer/rhdh-plugins/pull/3627#pullrequestreview-4626873304

I will be checking the same with PF team, too.

its-mitesh-kumar:
@karthikjeeyar Updated the patternfly version. Its solving the `double close icon` issue as well as the double icon issue @ciiay  was trying to fix here https://github.com/redhat-developer/rhdh-plugins/pull/3627 . I will be discussing this PF team and will do needful.

S_ 2026-07-09 at 8 42 00 AM (2)

karthikjeeyar:
>@karthikjeeyar I have commented my some findings here : https://github.com/redhat-developer/rhdh-plugins/pull/3627#pullrequestreview-4626873304

> I will be checking the same with PF team, too.

Thanks. Since PatternFly is gradually transitioning to Red Hat brand icons, we should temporarily hide`icon-default` using `display:none` (and include a TODO comment) instead of hiding `.pf-v6-icon-rh-ui`

I see this `pf-v6-icon-rh-ui` is hidden in some other places already, can you update that as well to hide the default icon infavor of RH UI icons? cc: @ciiay

workspaces/lightspeed/plugins/lightspeed/src/components/SourcesChipModal.tsx
its-mitesh-kumar: @karthikjeeyar For Adding ellipsis truncation for long file names. We need to override this:

```tsx
'& .pf-v6-c-list__item-text': {
  minWidth: 0,
},
```
Since , PF's `ListItem` wraps content in an internal `<span class="pf-v6-c-list__item-text">` which defaults to `min-width: auto` — flex children refuse to shrink below their content width. There's no PF prop or CSS variable available for this. Should I go ahead with this override, or skip the truncation?

S_ 2026-07-09 at 9 18 52 AM


Cc: @ShiranHi

its-mitesh-kumar: Added truncation for longer file name, as well as added tooltip.

S_ 2026-07-09 at 9 57 20 AM

S_ 2026-07-09 at 9 57 45 AM

Cc: @ShiranHi

karthikjeeyar (approved):
/lgtm

PR #3682, opened 2026-07-03
fix(lightspeed): disable model selector during active chat session

its-mitesh-kumar:
## Description

Disables the model selector once a chat session is active (has messages) to prevent the user from inadvertently creating a new conversation by switching models mid-session. When disabled, hovering the selector shows a localized tooltip explaining: "Each chat session supports only one model. To switch models, open a new chat."

**Root cause:** The `MessageBarModelSelector`'s `disabled` prop was only set to `isSendButtonDisabled` (true only during streaming). Once streaming completed, the selector became clickable again. The `onSelect` handler always called `onNewChat()`, which created a new conversation instead of keeping the user in the current session.

## Fixed

## Checklist

### DIFF
--- a/workspaces/lightspeed/.changeset/quiet-lions-rest.md
+++ b/workspaces/lightspeed/.changeset/quiet-lions-rest.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed': patch
+---
+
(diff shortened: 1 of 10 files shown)


its-mitesh-kumar:
This is the test PR by bug-fix skill .

PR #3683, opened 2026-07-03
fix(lightspeed): disable model selector when chat session is active

its-mitesh-kumar:
## Description

Disable the model selector in the Lightspeed chat footer when a chat session already has messages. Previously, switching models mid-conversation would open a new chat, fragmenting the user experience. Now the selector is disabled once a message has been sent, and a tooltip explains: "Each chat session supports only one model. To switch models, open a new chat."

### Root cause
The `MessageBarModelSelector` component in the message bar footer was only disabled during streaming (`isSendButtonDisabled`). Its `onSelect` handler called `onNewChat()` unconditionally, creating a new conversation whenever the model was changed — even in an active chat session.

## Fixed

## Checklist

### DIFF
--- a/workspaces/lightspeed/.changeset/gentle-pens-lock.md
+++ b/workspaces/lightspeed/.changeset/gentle-pens-lock.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed': patch
+---
+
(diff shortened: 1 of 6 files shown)


its-mitesh-kumar:
Closing this as its Test PR by https://github.com/redhat-developer/rhdh-skill/pull/54

PR #3685, opened 2026-07-04
fix(lightspeed): disable model selector in active chat sessions

its-mitesh-kumar:
## Description

Disable the model selector once an active chat session has started (messages exist in the conversation). When disabled, hovering over the selector shows a tooltip: "Each chat session supports only one model. To switch models, open a new chat." This prevents the confusing behavior where switching models mid-conversation would open a new chat thread.

### Root cause
The `MessageBarModelSelector` component's `onSelect` handler unconditionally called `onNewChat()` when a model was chosen, creating a new conversation. The selector was only disabled during streaming (`isSendButtonDisabled`), but remained enabled for conversations that already had messages.

The fix disables the selector when `messages.length > 0` (conversation has content) and wraps it in a `Tooltip` explaining the restriction. The tooltip message is fully localized in all 6 supported languages.

## Fixed

## Checklist

### DIFF
--- a/workspaces/lightspeed/.changeset/gentle-models-lock.md
+++ b/workspaces/lightspeed/.changeset/gentle-models-lock.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-lightspeed': patch
+---
+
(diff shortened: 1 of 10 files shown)


its-mitesh-kumar:
Close as this is Test PR by bug-fix skill

PR #3689, opened 2026-07-06
fix(orchestrator): remove incorrect aria-label on Run again button

lokanandaprabhu:
## Summary

- Fix misleading accessibility labels on the workflow instance page **Abort** and **Run again** buttons
- Only wrap buttons in a permission `Tooltip` when the user is **not** authorized to execute/abort
- When authorized, buttons render without a tooltip so MUI does not set `aria-label="user not authorized to execute workflow"` on enabled buttons

image


karthikjeeyar (approved):
/approve
/lgtm

PR #3690, opened 2026-07-06
fix(lightspeed): disable model selector during active chat sessions

its-mitesh-kumar:
## Summary

Fixes [RHDHBUGS-3313](https://redhat.atlassian.net/browse/RHDHBUGS-3313)

**Bug**: Switching the model in an active conversation incorrectly opens a new chat instead of staying in the current session.

**Root cause**: The model selector's `onSelect` handler in `LightSpeedChat.tsx` unconditionally called `onNewChat()`, creating a new conversation every time the model was changed — even mid-conversation.

**Fix**:

* Disable the model selector once a chat session has messages (`messages.length > 0`)
* Show a localized tooltip explaining why: _"Each chat session supports only one model. To switch models, open a new chat."_
* Remove the spurious `onNewChat()` call from the model selector's `onSelect` handler
* Move the attach file (+) button to the left-most position in the message bar for better UX
* Remove redundant `isModelSelectorDisabled` prop from the header model selector (it is already hidden via `hideModelSelector`)

## UI after changes

https://github.com/user-attachments/assets/d07dcd2f-1d73-40ff-88a6-1986f0ab68e0

## Test plan

* `yarn tsc:full` passes with no errors
* `yarn build` passes
* Manual verification: open Intelligent Assistant, send a message, confirm model selector is disabled with tooltip
* Manual verification: start a new chat, confirm model selector is enabled again
* Manual verification: confirm attach (+) button appears at the left-most position in the message bar


its-mitesh-kumar:
Cc: @HusneShabbir if you get time please check if atleast functionality is working properly. As it is first pr by `bug-fix` skill.

HusneShabbir (commented):
@its-mitesh-kumar  I agree that this is an edge-case scenario, but I think it's worth addressing as well.

Image

When the model dropdown is open and a message is sent, it's still possible to change the model mid-conversation. Please refer to the recording below.

It would be great if we could handle this case too and prevent the model from being changed while a conversation is in progress.

cc: @karthikjeeyar / @rohitkrai03 

https://github.com/user-attachments/assets/3c22aae7-f24a-4a32-8498-3828eb3940cb

its-mitesh-kumar:
> @its-mitesh-kumar I agree that this is an edge-case scenario, but I think it's worth addressing as well.
> 
> Image
> When the model dropdown is open and a message is sent, it's still possible to change the model mid-conversation. Please refer to the recording below.
> 
> It would be great if we could handle this case too and prevent the model from being changed while a conversation is in progress.
> 
> cc: @karthikjeeyar / @rohitkrai03
> 
>  Screen.Recording.2026-07-06.at.8.37.29.PM.mov

@HusneShabbir Now you can see as soon as user press `Enter` or `Send` button, Dropdown is closed.

tobiastal:
LGTM
cc @its-mitesh-kumar

HusneShabbir (commented):
Still able to reproduce the issue, PTAL

https://github.com/user-attachments/assets/a9687ceb-eb4c-4f3b-9ee9-65727822a90e

its-mitesh-kumar:
> * prop from the header model selector (it is already hidden via `hideModelSelector`)

Fixed now, also added e2e for the same.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/MessageBarModelSelector.tsx
```diff
@@ -121,4 +130,14 @@ export const MessageBarModelSelector = ({
       </DropdownList>
     </Dropdown>
   );
```
ciiay, line 133: Thanks for addressing the open-dropdown edge case in d634de2 — closing on disabled via `useEffect` is a good step.

One remaining concern: `useEffect` runs after paint, so while `isOpen` is still `true` the menu items can remain clickable for a frame (or longer if the effect is delayed). Could we close the menu synchronously and also guard `onSelect`?

```suggestion
const dropdown = (
  <Dropdown
    className={classes.dropdown}
    isOpen={isOpen && !disabled}
    onSelect={(_e, value) => {
      if (disabled) return;
      onSelect(value as string);
      setIsOpen(false);
    }}
    onOpenChange={open => setIsOpen(open)}
    popperProps={{ position: 'left' }}
    shouldFocusToggleOnSelect
    shouldFocusFirstItemOnOpen={false}
    toggle={toggle}
    isScrollable={models.length > 10}
    maxMenuHeight={models.length > 10 ? '240px' : undefined}
  >
    <DropdownList>
      {models.map(model => (
        <DropdownItem
          value={model.value}
          key={model.value}
          isSelected={selectedModel === model.value}
        >
          {model.label}
        </DropdownItem>
      ))}
    </DropdownList>
  </Dropdown>
);
```

With `isOpen={isOpen && !disabled}`, the menu collapses on the same render that disables the selector (e.g. right when Send/Enter sets `isSendButtonDisabled`), so the race Shabbir hit should be closed more reliably. The `useEffect` can then be removed as redundant, or kept as a belt-and-suspenders cleanup.

Once the open-dropdown → send → select-model flow on the updated change is re-verified by @HusneShabbir , I'm good with merging it.

its-mitesh-kumar: @ciiay Good point. Addressed it.

HusneShabbir (approved):
/lgtm, Thanks for fixing this

PR #3697, opened 2026-07-07
feat(lightspeed): replace Context-based drawer state with global store

its-mitesh-kumar:
## Description

Remove legacy (OFS) component re-exports from the main `./` entry point — they are now exclusively available at the `./legacy` subpath. Refactors the drawer state management from a `globalThis` React Context singleton to a proper global store using `@backstage/version-bridge` + `useSyncExternalStore`, eliminating the Provider dependency for NFS consumers and resolving cross-MF-boundary state sharing issues cleanly.

### Key Changes
- Legacy exports removed from `./` entry point (breaking) — consumers must use `./legacy`
- New `lightspeedDrawerStore.ts` provides a singleton store via `@backstage/version-bridge`
- New `useLightspeedDrawer` hook replaces `useLightspeedDrawerContext` using `useSyncExternalStore`
- `LightspeedDrawerProvider` simplified to a router-bridge shell (no Context.Provider)
- Updated API report, README migration guide, and changeset (major bump)

## Fixed
- https://redhat.atlassian.net/browse/RHIDP-15153

### What globalThis was solving

In Module Federation (NFS), each dynamically loaded plugin remote executes in its **own JavaScript module scope**. When the Lightspeed plugin's code is split across multiple remotes (main plugin, FAB module, Translations module), each remote runs its own copy of `LightspeedDrawerContext.tsx`. This means `createContext()` is called multiple times, producing **duplicate React Context objects**:

- Remote A (FAB): `createContext()` → `Context_A`
- Remote B (Drawer): `createContext()` → `Context_B`

The FAB component uses `Context_A` to read state, but `LightspeedDrawerProvider` provides values on `Context_B`. They never see each other → "must be used within a LightspeedDrawerProvider" error.

**The globalThis fix** (PR #3513) solved this by storing the Context in a well-known global key:

    const CONTEXT_KEY = '__lightspeed_drawer_context__' as keyof typeof globalThis;
    
    function getOrCreateContext() {
      const existing = (globalThis as any)[CONTEXT_KEY];
      if (existing) return existing;
      const ctx = createContext<LightspeedDrawerContextType | undefined>(undefined);
      (globalThis as any)[CONTEXT_KEY] = ctx;
      return ctx;
    }
    
    export const LightspeedDrawerContext = getOrCreateContext();

This ensured all remotes shared the **same Context object**, so Provider and consumers could communicate.

### Problems with globalThis approach

1. **Still requires Provider nesting** — Even with a shared Context, a `<LightspeedDrawerProvider>` must wrap all consumers in the React tree. In NFS, the FAB mounts via `AppRootWrapperBlueprint` and the drawer content via `AppDrawerContentBlueprint` — they're siblings, not parent-child. Ensuring the Provider wraps both depends on extension load order, which is fragile.

2. **Race condition risk** — If any remote calls `getOrCreateContext()` before the "canonical" one, it wins. The Context is created without any guarantee about which remote established it first.

3. **No version safety** — If two different plugin versions are loaded (e.g., during a rolling upgrade), `globalThis['__lightspeed_drawer_context__']` silently collides with no type safety or version negotiation.

4. **Invisible to React DevTools** — A raw `globalThis` key doesn't appear in React's component tree or DevTools Context inspector, making debugging harder.

5. **Not aligned with Backstage conventions** — Backstage has a purpose-built utility (`@backstage/version-bridge`) for exactly this problem, but it wasn't being used.

### How version-bridge + useSyncExternalStore solves it

The new approach eliminates React Context entirely for state management:

    // store/lightspeedDrawerStore.ts
    import { getOrCreateGlobalSingleton } from '@backstage/version-bridge';
    
    export const lightspeedDrawerStore = getOrCreateGlobalSingleton(
      'rhdh-lightspeed-drawer',
      createLightspeedDrawerStore,
    );

    // hooks/useLightspeedDrawer.ts
    import { useSyncExternalStore } from 'react';
    
    export function useLightspeedDrawer() {
      const snapshot = useSyncExternalStore(
        lightspeedDrawerStore.subscribe,
        lightspeedDrawerStore.getSnapshot,
        getServerSnapshot,
      );
      return { isChatbotActive: snapshot.isOpen, toggleChatbot: lightspeedDrawerStore.toggle, ... };
    }

**Why this is better:**

| Aspect | globalThis Context | version-bridge + useSyncExternalStore |
|--------|-------------------|--------------------------------------|
| Provider needed | Yes — must wrap all consumers | No — works from anywhere in the tree |
| Version conflicts | Silent collision on same key | version-bridge handles multiple versions gracefully |
| Concurrent mode | Context may tear during concurrent renders | useSyncExternalStore guarantees tear-free reads |
| Extension ordering | Provider must mount before consumers | Irrelevant — store exists before any React renders |
| Testability | Tests need Provider wrapper | Tests call store directly, no wrapper needed |
| Backstage alignment | Ad-hoc pattern | Uses the official recommended utility |

**How the data flows:**

    ┌─────────────────────────────────────────────────────────┐
    │  @backstage/version-bridge singleton                     │
    │  getOrCreateGlobalSingleton('rhdh-lightspeed-drawer')   │
    │                                                          │
    │  ┌──────────────────────────────────────────────────┐   │
    │  │  lightspeedDrawerStore                            │   │
    │  │  • state: { isOpen, displayMode, ... }           │   │
    │  │  • listeners: Set<() => void>                    │   │
    │  │  • handlers: (registered by router-bridge shell) │   │
    │  └──────────────────────────────────────────────────┘   │
    └──────────────┬──────────────────────┬───────────────────┘
                   │                      │
        subscribe + getSnapshot    subscribe + getSnapshot
                   │                      │
        ┌──────────▼──────────┐  ┌───────▼─────────────┐
        │  FAB Remote          │  │  Drawer Remote       │
        │  useLightspeedDrawer │  │  useLightspeedDrawer │
        │  → reads isOpen      │  │  → reads displayMode │
        │  → calls toggle()    │  │  → renders chat UI   │
        └─────────────────────┘  └─────────────────────┘

No Provider in between. No Context. No ordering dependency. Just a shared store that any component in any remote can subscribe to.

### Prerequisites
- In `.env` of RHDH-local 

```
APP_CONFIG_app_packageName=app-next
ENABLE_STANDARD_MODULE_FEDERATION=true
```

- app-defaults plugin (provides `app-root-wrapper:app/drawer`)

### 1. Build the Lightspeed dynamic plugin

```bash
cd workspaces/lightspeed/plugins/lightspeed
yarn build
npx @red-hat-developer-hub/cli@latest plugin export
cp -r dist-dynamic /path/to/rhdh-local/local-plugins/red-hat-developer-hub-backstage-plugin-lightspeed
```

### 2. Configure dynamic plugins

In `dynamic-plugins.override.yaml`:

```yaml
includes:
  - dynamic-plugins.default.yaml
  - /dynamic-plugins-root/dynamic-plugins.extensions.yaml

plugins:
  # Disable OCI lightspeed frontend
  - package: 'oci://registry.access.redhat.com/rhdh/red-hat-developer-hub-backstage-plugin-lightspeed:{{inherit}}'
    disabled: true
  # Enable local build
  - package: ./local-plugins/red-hat-developer-hub-backstage-plugin-lightspeed
    disabled: false
  # Enable app-defaults (provides drawer shell)
  - package: ./local-plugins/red-hat-developer-hub-backstage-plugin-app-defaults
    disabled: false
  # Keep backend from OCI
  - package: 'oci://registry.access.redhat.com/rhdh/red-hat-developer-hub-backstage-plugin-lightspeed-backend:{{inherit}}'
    disabled: false
```

### 3. Configure NFS extensions

In `configs/app-config/app-config.local.yaml`, use either option:

```yaml
# Option A: Auto-discover all extensions (simplest — recommended)
app:
  packages: all

# Option B: Explicit list (if you need to limit what's enabled)
app:
  extensions:
    - app-root-wrapper:app/drawer
    - app-root-wrapper:app/lightspeed-fab
    - translation:app/lightspeed-translations
```

> **Note:** With the global store refactor, extension ordering no longer matters — the drawer state no longer relies on React Context nesting.

### 4. Start RHDH-local

```bash
cd /path/to/rhdh-local
podman compose down
podman compose up -d
```

### 5. Verify

- FAB button appears on all pages
- Click FAB → drawer opens in "docked to window" mode
- Toggle to overlay mode → modal opens correctly
- Chat history, new conversation, model selector all functional
- No "must be used within a Provider" errors in console

## ✔️ Checklist


its-mitesh-kumar:
Closing this in favor of https://github.com/redhat-developer/rhdh-plugins/pull/3708.

PR #3711, opened 2026-07-08
feat(extensions): add source metadata to entities

hopehadfield:

Resolves [RHIDP-13665](https://redhat.atlassian.net/browse/RHIDP-13665) / Epic [RHIDP-13606](https://redhat.atlassian.net/browse/RHIDP-13606)

### Problem

When RHDH is configured with multiple catalog index images (`CATALOG_INDEX_IMAGE` + `EXTRA_CATALOG_INDEX_IMAGES`), the `install-dynamic-plugins` init container extracts entities from each image into separate directories. However, once `BaseEntityProvider` ingests them into the Backstage catalog, all source information is lost with the current implementation.

### Solution

Add an `extensions.backstage.io/catalog-source` annotation that is **automatically derived** from the on-disk directory layout and set on every entity emitted by `BaseEntityProvider`.

The derivation logic:
- Entities from `extra/<name>/catalog-entities/…` get `catalog-source: "<name>"` (matching the name from `EXTRA_CATALOG_INDEX_IMAGES`)
- All other entities get `catalog-source: "primary"`

### Changes

| File | Change |
|------|--------|
| `extensions-common/src/annotations.ts` | Add `CATALOG_SOURCE` to `ExtensionsAnnotation` enum |
| `catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts` | Add `deriveCatalogSource()` static method; update `addProviderAnnotations()` to set the annotation using the entity's file path |
| `catalog-backend-module-extensions/src/providers/BaseEntityProvider.test.ts` | 12 new tests: source derivation (7) + annotation integration (5) |
| `docs/catalog/plugins.md` | Document the new annotation |
| `catalog-backend-module-extensions/README.md` | Document multi-source catalog layout and source derivation |

### How to test

**Unit tests:**
```bash
cd workspaces/extensions/plugins/catalog-backend-module-extensions
yarn backstage-cli package test --watchAll=false
```

**Local dev (manual):**
1. Create a multi-source directory:
   ```
   test-extensions/
     catalog-entities/plugins/primary-plugin.yaml
     extra/community/catalog-entities/plugins/community-plugin.yaml
   ```
2. Set `extensions.directory` in `app-config.yaml` to point at it
3. Start the backend and inspect the relevant plugin entities

### Design decisions

- **`"primary"` as default.** Vendor-neutral. The UI/config layer can map it to a display label as desired.
- **Freeform string values.** Not an enforced enum — operators choose source names via `EXTRA_CATALOG_INDEX_IMAGES` (`community=quay.io/...`).
- **Applies to all entity kinds.** Implemented in `BaseEntityProvider`, so Plugins, Packages, and Collections all get the annotation.
- **No change to duplicate handling.** Existing first-wins dedup policy is unchanged; source annotations don't affect it.

### DIFF
--- a/workspaces/extensions/docs/catalog/plugins.md
+++ b/workspaces/extensions/docs/catalog/plugins.md
@@ -229,6 +229,21 @@ spec:
 
 ## Annotations
 
+### Catalog source
+
+Automatically set by the `BaseEntityProvider` to identify which catalog image each entity came from. The value is derived from the on-disk directory layout produced by `install-dynamic-plugins`:
+
+- Entities under the primary `catalog-entities/` tree get `"primary"`
+- Entities under `extra/<name>/catalog-entities/` get `<name>` as the source (matching the name from `EXTRA_CATALOG_INDEX_IMAGES`)
+
+This annotation is always set and should not be manually specified in YAML files.
+
+```yaml
+metadata:
+  annotations:
+    extensions.backstage.io/catalog-source: 'primary'
+```
+
 ### Support type for Core and Community plugins
 
 ```yaml
--- a/workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
+++ b/workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
@@ -74,18 +74,24 @@ export abstract class BaseEntityProvider<T extends Entity>
    *
    * When `install-dynamic-plugins` extracts extra catalog index images
    * (via `EXTRA_CATALOG_INDEX_IMAGES`), entities are placed under
-   * `<extensionsDir>/extra/<name>/catalog-entities/…`. This method
-   * detects the `extra/<name>/` segment and returns `<name>` as the
+   * `<extensionsRoot>/extra/<name>/catalog-entities/…`. This method
+   * computes the relative path from `extensionsRoot` and checks whether
+   * the first segment is `extra/` — if so, the second segment is the
    * source identifier. All other paths return `"primary"`.
    */
-  static deriveCatalogSource(filePath: string): string {
-    // Normalise to forward slashes so the regex works on all platforms.
-    const normalised = filePath.replace(/\\/g, '/');
-    const match = normalised.match(/\/extra\/([^/]+)\//);
-    return match ? match[1] : BaseEntityProvider.DEFAULT_CATALOG_SOURCE;
+  static deriveCatalogSource(filePath: string, extensionsRoot: string): string {
+    const relative = path.relative(extensionsRoot, filePath);
+    const [first, second] = relative.split(path.sep);
+    return first === 'extra' && second
+      ? second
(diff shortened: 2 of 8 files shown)


workspaces/extensions/plugins/catalog-backend-module-extensions/src/providers/BaseEntityProvider.ts
```diff
@@ -61,7 +62,30 @@ export abstract class BaseEntityProvider<T extends Entity>
...
+   *
+   * When `install-dynamic-plugins` extracts extra catalog index images
+   * (via `EXTRA_CATALOG_INDEX_IMAGES`), entities are placed under
+   * `<extensionsDir>/extra/<name>/catalog-entities/…`. This method
+   * detects the `extra/<name>/` segment and returns `<name>` as the
+   * source identifier. All other paths return `"primary"`.
+   */
+  static deriveCatalogSource(filePath: string): string {
+    // Normalise to forward slashes so the regex works on all platforms.
+    const normalised = filePath.replace(/\\/g, '/');
```
jonkoops: Node.js also has utilities in the `node:path` module to normalize paths; you can import specific ones for specific plaforms (win32, unix) if you need them as well. Perhaps you could do something like (untested):

```ts
import path from 'node:path';

static deriveCatalogSource(filePath: string, extensionsRoot: string): string {
  const relative = path.relative(extensionsRoot, filePath);
  const [first, second] = relative.split(path.sep);
  return first === 'extra' && second ? second : BaseEntityProvider.DEFAULT_CATALOG_SOURCE;
}
```

This would also allow you to avoid regexes, which are challenging to parse at a glance.

hopehadfield: Implemented, thanks for the tip!

workspaces/extensions/docs/catalog/plugins.md
```diff
@@ -229,6 +229,21 @@ spec:
...
 ## Annotations
 
+### Catalog source
+
+Automatically set by the `BaseEntityProvider` to identify which catalog image each entity came from. The value is derived from the on-disk directory layout produced by `install-dynamic-plugins`:
+
+- Entities under the primary `catalog-entities/` tree get `"primary"`
+- Entities under `extra/<name>/catalog-entities/` get `<name>` as the source (matching the name from `EXTRA_CATALOG_INDEX_IMAGES`)
+
+This annotation is always set and should not be manually specified in YAML files.
```
christoph-jerolimov, line 239: I personally think this isn't a good pattern, derivate annotations from a path. But you have started this way and I don't want stop you having our timelines and other priorities in mind. At some point I would be happy to have an implementation IN the plugin instead of something dictated from outside.

christoph-jerolimov (approved):
One comment that I'm not 100% happy with the architecture of the integration, but for now, we can move forward to verify this e2e.

christoph-jerolimov:
The PR didn't get updated, then I wanted close and reopen it but this also didn't worked...

image

So I created a new one, let see when this gets green. Sorry.

Issue #3720, opened 2026-07-10
boost-common: add test coverage for permission array immutability


durandom:
## Problem

The `boostPermissions`, `boostAgentPermissions`, `boostToolPermissions`, and `boostFunctionalPermissions` arrays exported from `boost-common` are mutable. A consumer could accidentally `push()` or `splice()` them, corrupting the shared reference for other callers.

## Expected behavior

Add a test to `plugins/boost-common/src/index.test.ts` that verifies the exported permission arrays are frozen (or at minimum that mutating them throws or has no effect), so any future `Object.freeze()` addition is covered by tests.

## Scope

- **Workspace:** `workspaces/boost`
- **Package:** `@red-hat-developer-hub/backstage-plugin-boost-common`
- **File:** `plugins/boost-common/src/index.test.ts`

Add a new `describe('permission array immutability')` block with tests that attempt `.push()` on each exported array and assert it either throws or leaves the length unchanged.

---

_Smoke-test issue for fullsend v0.30.0 upgrade verification._

PR #3717, opened 2026-07-09
chore(fullsend): upgrade to v0.30.0 base composition model

durandom:
## Summary

- Replace full-file harness forks (`customized/harness/`) and inline `network_policies` with the v0.30.0 `base:` URL composition pattern and provider-backed profiles
- Harness files now inherit upstream via pinned commit SHA (`807037ff`) and only override RHDH-specific fields: custom image, toolchain env, pre-fix rebase script, rhdh-workspace skill
- Inline policies replaced by modular provider/profile YAML files from upstream scaffold, with `fullsend-package-registries` patched for `repo.yarnpkg.com`, `**/corepack`, and `allow_encoded_slash`
- Workflow ref bumped from pinned SHA (`e22a1eff`, v0.29.0) to `@v0` floating tag

Same migration pattern as [rhdh-agentic PR #82](https://github.com/redhat-developer/rhdh-agentic/pull/82), adapted for rhdh-plugins' customizations.

## Test plan

- [ ] Trigger `/fs-code` on a test issue to verify dispatch routes correctly
- [ ] Verify pre-flight connectivity check passes (api.github.com reachable)
- [ ] Verify sandbox can install yarn dependencies (scoped packages, corepack)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3723, opened 2026-07-10
fix(fullsend): clear agent field to use default claude agent type

durandom:
## Summary

- Add `agent: ""` to both `harness/code.yaml` and `harness/fix.yaml` to prevent inheriting `agent: agents/code.md` / `agents/fix.md` from the upstream base
- The custom agent types lack Bash access in the sandbox, causing the agent to waste ~15 minutes and $6 discovering workarounds via subagent spawning before actually working
- Matches the pattern used in rhdh-agentic (`agent: ""` on code.yaml)

**Root cause:** with `base:` composition, omitting a field means you inherit it from upstream. The upstream base sets `agent: agents/code.md`, which runs the sandbox as a custom "code" agent type. That type doesn't have Bash enabled. Setting `agent: ""` clears the inherited value and uses the default agent type (which has full tool access).

Discovered during smoke test of #3717 on issue #3720 — see run transcript analysis in the issue.

## Test plan

- [ ] Trigger `/fs-code` on a test issue — agent should have Bash at the top level (no subagent spawning)
- [ ] Run should complete in ~15 min instead of ~35 min

### DIFF
--- a/.fullsend/harness/code.yaml
+++ b/.fullsend/harness/code.yaml
@@ -1,3 +1,4 @@
+agent: ""
 base: https://raw.githubusercontent.com/fullsend-ai/fullsend/807037ff09a6b5d18fbc339d06e57752af9cd522/internal/scaffold/fullsend-repo/harness/code.yaml#sha256=adba903da2cbb861205a75e906dca341eb6dd67560224421c3ed5975c4e00ca1
 image: ghcr.io/redhat-developer/rhdh-fullsend-code:latest
 host_files:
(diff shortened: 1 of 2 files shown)


gabemontero (approved):

gabemontero:
merge at your discretion @durandom

durandom:
Superseded by #3740, which adopts the complete rhdh-agentic reference install. As documented in redhat-developer/rhdh-fullsend#23, an empty agent override inherits the base value and cannot resolve the Bash tool issue.

PR #3729, opened 2026-07-10
fix(orchestrator): drop Alpha module from translationResources

lokanandaprabhu:
## Summary
- Remove the explicit `module: Alpha` from Orchestrator `translationResources` in the plugin `app-config.yaml`
- RHDH defaults to `PluginRoot` when `module` is omitted, which avoids loading the Alpha Scalprum bundle and the related module initialization race

## Context
Loading translations from the Alpha module can cause intermittent page load failures (`Cannot access 'U' before initialization`) because the host loader initializes the Alpha bundle, which also exports the NFS plugin as its `default` export. This can happen even with NFS disabled.

## Test plan
- [ ] Upgrade to RHDH/IDH 1.10 with Orchestrator dynamic plugin enabled and verify the app loads consistently (no stuck spinner)
- [ ] Confirm Orchestrator translations still load correctly in legacy/dynamic plugin mode
- [ ] Confirm NFS apps importing from `/alpha` are unaffected

### DIFF
--- a/workspaces/orchestrator/.changeset/fix-orchestrator-translation-module.md
+++ b/workspaces/orchestrator/.changeset/fix-orchestrator-translation-module.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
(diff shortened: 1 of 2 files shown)


karthikjeeyar (approved):
/approve
/lgtm

PR #3738, opened 2026-07-13
fix(x2a): copy gitignored files in job script instead of only warning

hardengl:
## Summary

`copy_changed_files()` in the x2a job script detects files created by x2a that match the source repo's `.gitignore` but only logs a warning — it never actually copies them to the project output directory.

When `generated-project-metadata.json` (or any other x2a output) is matched by the source repo's `.gitignore`, the file is silently excluded from the copy. This causes the init phase to crash at:

```bash
METADATA=$(cat ${PROJECT_PATH}/generated-project-metadata.json)
```

Under `set -eo pipefail`, this exits the script with code 1. The K8s Job retries up to `backoffLimit=3` (4 pods total), all fail, and the project state becomes `failed`.

## Root cause

Commit f2b9c97f added detection of gitignored files and a WARNING log, but did not merge them into the `changed_files` array that drives the actual copy loop.

## Fix

Append each ignored file to `changed_files` inside the warning loop, so they are copied alongside other x2a-generated output. The warning is preserved (with updated wording) for visibility.

## How to reproduce

1. Create an x2a project targeting a Chef cookbook repo whose `.gitignore` matches `*.json` or `generated-project-metadata.json`
2. Run the init phase
3. Observe: all K8s pods for the init job end in Error state; project state = failed
4. Pod logs show `cat: .../generated-project-metadata.json: No such file or directory`

## Test plan

- [ ] Run init phase against a cookbook with `.gitignore` matching `*.json` — verify `generated-project-metadata.json` is copied and init succeeds
- [ ] Run init phase against a cookbook with no matching `.gitignore` — verify behavior is unchanged
- [ ] Check pod logs: WARNING line should now say "including them anyway" and the file should appear in the copied count

### DIFF
--- a/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
+++ b/workspaces/x2a/plugins/x2a-backend/templates/x2a-job-script.sh
@@ -151,9 +151,10 @@ copy_changed_files() {
   done < <(git ls-files --others --ignored --exclude-standard)
 
   if [[ ${#ignored_files[@]} -gt 0 ]]; then
-    echo "  WARNING: ${#ignored_files[@]} file(s) created by x2a but blocked by .gitignore:"
+    echo "  WARNING: ${#ignored_files[@]} file(s) created by x2a matched .gitignore — including them anyway:"
     for file in "${ignored_files[@]}"; do
       echo "    - $file"
+      changed_files+=("$file")
     done
   fi
 


eloycoto:
umm, but if it's part of the gitignore, it shoudn't be commited, I remember that @elai-shalev worked on that.

hardengl:
@eloycoto I am just seeing if it fixes it, I will report back .

hardengl:
## CI Triage Findings (2026-07-14)

After deep investigation of the 6-day CI failure streak (builds #1192–#1252, all with `state=failed` for init), we confirmed that **this PR's fix (the gitignore copy bug) is NOT the cause of the current CI failures.**

The current failures are caused by an **expired/invalid AWS Bedrock bearer token** (`AWS_BEARER_TOKEN_BEDROCK`) — the x2a-convertor crashes at LLM authentication before it ever produces output files, so the `copy_changed_files()` code path is never reached.

Pod logs from build #1249 show:
```
Error: AWS Bedrock access denied.
Check your AWS credentials and IAM permissions.
Details: Authentication failed: Please make sure your API Key is valid.
```

This means:
1. **This PR's fix is still valid and needed** — it fixes a real bug where gitignored convertor output files are silently skipped by the job script
2. **This PR cannot be validated in CI right now** because the Bedrock token issue (tracked in [FLPATH-4582](https://redhat.atlassian.net/browse/FLPATH-4582)) blocks all init jobs before the copy step
3. **The CI source repos** (`puppet-examples`, `chef-examples`, `chef-examples-metadata`) **do not have `*.json` in their `.gitignore`**, so this PR's fix would not change CI behavior even when Bedrock is working — it fixes a scenario that affects customer repos with restrictive `.gitignore` patterns
4. Once the Bedrock token is rotated (FLPATH-4582), this PR should be mergeable and testable

Tracked in:
- [FLPATH-4575](https://redhat.atlassian.net/browse/FLPATH-4575) — this PR's bug (gitignore copy)
- [FLPATH-4582](https://redhat.atlassian.net/browse/FLPATH-4582) — the Bedrock token issue blocking CI

hardengl:
Closing — this is no longer needed. The upstream fix in PR #3436 (FLPATH-4383) added gitignore detection and warnings, and the underlying issue doesn't manifest with the repos used in CI testing. The x2a team's position is that gitignored files should remain ignored rather than force-copied, which is reasonable. If the edge case resurfaces with a specific source repo, we can revisit.

PR #3741, opened 2026-07-13
fix(orchestrator): hide Workflows tab from User profile entity pages

its-mitesh-kumar:
## Description
Fixes unintended 'Workflows' tab appearing on User profile entity pages. The `IsOrchestratorCatalogTabAvailable` condition function was implemented as a React component using `useEntity()` hook, which is incompatible with the RHDH dynamic-plugin framework's expected `(e: Entity) => boolean` signature. This caused the condition to not properly filter out User entities, making the Workflows tab appear on all entity pages including the My Profile page.

### Root cause
`IsOrchestratorCatalogTabAvailable` in `CatalogTab.tsx` was a React component that used the `useEntity()` hook instead of accepting the entity as a parameter. The RHDH dynamic-plugin framework expects custom `if` condition functions to follow the `(e: Entity) => boolean` signature and calls them as plain functions with the entity — the hook-based implementation did not work correctly in this context.

Additionally, the `app-config.yaml` mount point condition referenced this broken custom component instead of using the framework's built-in `hasAnnotation` checker. The combination caused the Workflows tab to appear unconditionally on all entity pages, including User profile pages where it should not be shown.

### Changes
1. **`CatalogTab.tsx`**: Changed `IsOrchestratorCatalogTabAvailable` from a React hook-based component to a pure predicate function `(entity: Entity) => boolean`. Compatible with both the legacy `EntityLayout.Route` `if` prop and the RHDH dynamic-plugin `if.anyOf` condition system.
2. **`app-config.yaml`**: Replaced the custom `IsOrchestratorCatalogTabAvailable` reference with the built-in `hasAnnotation: orchestrator.io/workflows` checker for reliable entity filtering in the dynamic plugin system.

## Fixed
- [RHDHBUGS-3245](https://redhat.atlassian.net/browse/RHDHBUGS-3245) — [Profile] Unintended 'Workflows' tab displayed under 'My profile' section

## Test Plan
- [ ] Navigate to My Profile page (click avatar → My Profile in global header)
- [ ] Verify the Workflows tab is NOT visible on the User entity page
- [ ] Navigate to a Component entity that HAS the `orchestrator.io/workflows` annotation
- [ ] Verify the Workflows tab IS visible and shows workflow content
- [ ] Navigate to a Component entity WITHOUT the `orchestrator.io/workflows` annotation
- [ ] Verify the Workflows tab is NOT visible
- [ ] Verify no regressions on other entity pages (API, System, Group, Domain)

## Checklist

### DIFF
--- a/workspaces/orchestrator/.changeset/gentle-foxes-march.md
+++ b/workspaces/orchestrator/.changeset/gentle-foxes-march.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Fix unintended Workflows tab on User profile pages by converting the entity tab condition from a React hook-based component to a plain predicate function and using the built-in hasAnnotation checker in dynamic plugin config
--- a/workspaces/orchestrator/plugins/orchestrator/app-config.yaml
+++ b/workspaces/orchestrator/plugins/orchestrator/app-config.yaml
@@ -30,4 +30,4 @@ dynamicPlugins:
               gridColumn: '1 / -1'
             if:
               anyOf:
-                - IsOrchestratorCatalogTabAvailable
+                - hasAnnotation: orchestrator.io/workflows
--- a/workspaces/orchestrator/plugins/orchestrator/report.api.md
+++ b/workspaces/orchestrator/plugins/orchestrator/report.api.md
@@ -4,14 +4,15 @@
 
 ```ts
 import { BackstagePlugin } from '@backstage/core-plugin-api';
+import { Entity } from '@backstage/catalog-model';
 import { JSX as JSX_2 } from 'react/jsx-runtime';
 import { RouteRef } from '@backstage/core-plugin-api';
 import { SvgIconProps } from '@mui/material/SvgIcon';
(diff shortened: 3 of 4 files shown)


lokanandaprabhu:
**Note:** After this PR merges and the plugin is published, we also need a follow-up PR on [`rhdh-plugin-export-overlays`](https://github.com/redhat-developer/rhdh-plugin-export-overlays) to update:

1. The commit/version reference to pick up this fix
2. The condition at [line 68 of `rhdh-bsp-orchestrator.yaml`](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/orchestrator/metadata/rhdh-bsp-orchestrator.yaml#L68) from:
   ```yaml
   if:
     anyOf:
       - IsOrchestratorCatalogTabAvailable
   ```
   to:
   ```yaml
   if:
     anyOf:
       - hasAnnotation: orchestrator.io/workflows
   ```

Without this overlay update, the dynamic plugin deployment in RHDH will still use the old broken condition and the Workflows tab will continue appearing on User profile pages.

lokanandaprabhu (approved):
/lgtm

PR #3814, opened 2026-07-17
fix(theme): inline config.d.ts types to fix matchMedia schema crash

its-mitesh-kumar:
## Description

Inlines all `ThemeConfig` types directly in `config.d.ts` with zero imports, replacing the previous `import { ThemeConfig } from './'` that pulled in the entire `@mui/material` type graph via `dist/index.d.ts`.

### Root cause

The theme plugin's `config.d.ts` imports `ThemeConfig` from the package entry point (`'./'`), which resolves to `dist/index.d.ts`. That file contains `import * as _mui_material from '@mui/material'`, which transitively reaches `@mui/system/useMediaQuery/useMediaQuery.d.ts` containing:

```typescript
matchMedia?: typeof window.matchMedia;
```

`typeof window.matchMedia` is a browser-only interface method with no **value declaration** — only a type signature. When `ts-json-schema-generator` (used by `@backstage/config-loader` v1.11.0) encounters this, its `TypeofNodeParser` throws:

```
Error: No value declaration found for symbol "matchMedia"
```

### Why it was working before

`@backstage/config-loader` versions <=1.10.x used `noResolve: true` in its TypeScript compiler options, meaning all `import` statements in `config.d.ts` files were **silently ignored**. `ThemeConfig` was treated as `any`, and the MUI type graph was never followed.

Additionally, workspaces on Backstage 1.49.x used the **legacy Scalprum export path** in the CLI, which didn't use `ts-json-schema-generator` at all.

### Why it's breaking now

`@backstage/config-loader` v1.11.0 (resolved by `@red-hat-developer-hub/cli@1.10.7`) switched to:
- `ts-json-schema-generator` instead of `typescript-json-schema`
- `noResolve: false` instead of `true` — imports are now **followed**
- `skipLibCheck: false` — strict type checking enabled

The `intelligent-assistant` workspace is the first `rhdh-plugins` workspace on Backstage 1.52.x, which triggers the **standard module federation export path** with the new config-loader — exposing this bug.

The import chain: `config.d.ts` → `'./'` → `dist/index.d.ts` → `@mui/material` → `@mui/system/useMediaQuery` → `typeof window.matchMedia` → crash.

### Fix

Replace the import with fully inlined types. All fields from `ThemeConfig`, `ThemeConfigPalette`, `RHDHThemePalette`, `ThemeConfigPageTheme`, `ThemeConfigOptions`, and `BackstageTypography` are spelled out directly with primitive types — no imports, no external type references.

The `@deepVisibility frontend` annotation is preserved. MUI's standard palette keys use `[key: string]: unknown` since they are extensible/freeform.

A sync-note comment references `src/types.ts` as the source of truth for the runtime types.

### How to test locally

```bash
# 1. Go to the intelligent-assistant frontend plugin
cd workspaces/intelligent-assistant/plugins/intelligent-assistant

# 2. Install dependencies
yarn install

# 3. Copy the fixed config.d.ts into node_modules
cp ../../node_modules/@red-hat-developer-hub/backstage-plugin-theme/config.d.ts ../../node_modules/@red-hat-developer-hub/backstage-plugin-theme/config.d.ts.bak
cp /path/to/this/pr/workspaces/theme/plugins/theme/config.d.ts ../../node_modules/@red-hat-developer-hub/backstage-plugin-theme/config.d.ts

# 4. Run the export — should succeed with no errors
npx @red-hat-developer-hub/cli@1.10.7 plugin export
```

CI failure log (before fix):
-  https://github.com/redhat-developer/rhdh-plugin-export-overlays/actions/runs/29517726978/job/87687030869
- https://github.com/redhat-developer/rhdh-plugin-export-overlays/actions/runs/29558493912/job/87815791586
## Fixed
- Unblocks [redhat-developer/rhdh-plugin-export-overlays#2785](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2785) (intelligent-assistant rename)
- Will unblock all future 1.52.x+ workspace exports that depend on the theme package

## ✔️ Checklist


lokanandaprabhu (approved):
Verified locally — applied the inlined `config.d.ts` to `node_modules` in the orchestrator workspace and `npx @red-hat-developer-hub/cli@1.10.7 plugin export` succeeds with no errors (previously hit both the `TS2307: Cannot find module './src'` and `No value declaration found for symbol "matchMedia"` crashes).

Fix approach is clean — `config.d.ts` only serves config schema validation, so inlining primitive types avoids the entire MUI type graph without any runtime impact. Types look complete and match the `ThemeConfig` interfaces in `src/types.ts`.

PR #3820, opened 2026-07-17
fix(theme): prevent overlapping scrollbars on TechDocs pages

karthikjeeyar:
## Fixes: 

https://redhat.atlassian.net/browse/RHDHBUGS-3461

## Summary

The page scrollbar and the TechDocs Table of Contents scrollbar overlap because both end up at the same horizontal position (~24px from the right edge).
    
  ## Root cause
The page-inset styles set `max-height` on `<main>`, which gives it its own scrollbar. Meanwhile, the TechDocs ToC sidebar inside the shadow DOM is fixed at `right: 24px`. Since the page-inset margin is also `1.5rem`(24px), the two scrollbars land right on top of each other.

  ## Fix

The fix removes the height constraint on `<main>` only when TechDocs content is present, using `main:has([data-testid='techdocs-native-shadowroot'])`. This lets `<main>` grow with its content so it no longer scrolls on its own, the scrolling moves to the outer page container instead, whose scrollbar sits at the viewport edge, away from the ToC scrollbar.

## Screenshots

https://github.com/user-attachments/assets/823fba1b-456c-4310-9b92-e02c2bc37fe3

 
 ## How to test
 
1. Import the sample techdocs from https://github.com/hiroyha-r/sample-catalog/blob/main/catalog-info.yaml 
2. Visit `sample-techdocs-1`  techdocs page.
3. Scrollbars should not overlap in the techdocs page.

 
 

### DIFF
--- a/workspaces/theme/.changeset/fix-techdocs-double-scrollbar.md
+++ b/workspaces/theme/.changeset/fix-techdocs-double-scrollbar.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+prevent overlapping scrollbars on TechDocs pages
--- a/workspaces/theme/packages/app/package.json
+++ b/workspaces/theme/packages/app/package.json
@@ -54,7 +54,9 @@
     "react-router-dom": "^6.30.2"
   },
   "devDependencies": {
+    "@axe-core/playwright": "^4.11.0",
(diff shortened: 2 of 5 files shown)


lokanandaprabhu (approved):
Tested locally on RHDH, working fine


https://github.com/user-attachments/assets/3d5b5d28-63b1-4eed-b5e4-639a48f3d5e9

PR #3826, opened 2026-07-17
fix(deps): bump @red-hat-developer-hub/backstage-plugin-theme to ^0.14.11

jonkoops:
Versions of `@red-hat-developer-hub/backstage-plugin-theme` prior to 0.14.10 shipped a `config.d.ts` that imported from `./src`, a path that does not exist in the published package. This caused TypeScript compilation errors (`TS2307: Cannot find module './src'`) in all workspaces that depend on the theme package.

This bumps the minimum version to `^0.14.11` (which inlines the types instead of importing them) in the 7 affected workspaces: adoption-insights, bulk-import, extensions, global-header, homepage, orchestrator, and quickstart.


nickboldt (approved):
/lgtm
/seems-legit
/make-it-so

PR #3830, opened 2026-07-18
fix(quickstart): correct scalprum QuickstartPlugin module path

jonkoops:
The `QuickstartPlugin` entry in the scalprum `exposedModules` config referenced `./src/index.ts` but the actual file is `./src/index.tsx`. This went unnoticed because webpack resolves the extension automatically during local development, but the stricter packaging step fails on the exact path.

Introduced in #3530.


its-mitesh-kumar (approved):
/lgtm

jonkoops:
Thanks @its-mitesh-kumar :)

PR #3834, opened 2026-07-19
feat(intelligent-assistant): add inline rename for notebook resources

its-mitesh-kumar:
## Description

Adds inline rename capability for notebook documents (resources) in the DocumentSidebar with full backend support.

## Fixed 
- https://redhat.atlassian.net/browse/RHIDP-14854

### Backend
- Consolidated `renameDocument` logic into `upsertDocument` using an options object (`{ fileType?, fileId?, newTitle? }`) to eliminate code duplication
- `upsertDocument` includes rollback protection: if the vector store re-creation fails after delete, the original entry is restored
- Added title length validation (max 255 characters) in the PATCH endpoint with a 400 response for violations
- Defensive fallback for `resolvedFileType` to prevent `undefined` values in attributes

### Frontend
- Users can double-click a filename or select "Rename" from the kebab menu to edit inline
- File extension shown as a static suffix (cannot be accidentally modified)
- Client-side validation with real-time error feedback:
  - Duplicate name detection ("Name already exists.")
  - Title length enforcement ("Name too long (max 255).")
  - Legacy long titles (pre-existing > 255 chars) don't trigger error until the user changes the name
- Hover effects on document rows and file names for discoverability
- Save on blur (clicking away commits the rename; Escape cancels)
- Only error toast on failure (no success toast — the inline name change provides visual confirmation)
- Error icon vertically centered in the input field (PF6 padding override)

### UI after fix 

https://github.com/user-attachments/assets/5af15fcd-5c89-4cf9-a1e5-0a0501e1fd51

### Translations
- Added `notebook.document.rename.tooLong` key across all 6 languages
- Shortened conflict message to "Name already exists."
- Removed orphaned `notebook.document.rename.success` key

## Fixed

* [RHIDP-14854
](https://redhat.atlassian.net/browse/RHIDP-14854)
## ✔️ Checklist


its-mitesh-kumar:
@JslYoon Please review the backend changes.

HusneShabbir (commented):
I noticed a slight delay in the file renaming process. It seems the non-atomic rename relies on a delete-then-create flow, which introduces a brief inconsistency window. I'd encourage @JslYoon  to take a closer look and suggest any improvements if applicable.

Also, could you please address comments **1** and **2** from the Fullsend review? Specifically:

* **1:** The non-atomic rename (`delete` → `create`) introduces a brief inconsistency window.
* **2:** Very long title names don't add much value; it would be better to consider adding a length cap.

https://github.com/redhat-developer/rhdh-plugins/pull/3834#issuecomment-5015363269

https://github.com/user-attachments/assets/ffa4679c-42a5-4632-8c2c-fd6243a78214

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/notebooks/documents/documentService.ts
```diff
@@ -259,6 +259,74 @@ export class DocumentService {
...
+   * Rename a document by updating its title attribute in the vector store.
+   * Deletes the old vector store file entry and re-creates it with the
+   * same underlying file but an updated title attribute.
+   * @param sessionId - Vector store ID
+   * @param currentTitle - Current document title
+   * @param newTitle - New document title
+   * @throws NotFoundError if document not found
+   * @throws ConflictError if newTitle conflicts with existing document
+   */
+  async renameDocument(
```
JslYoon: There already exists a function **upsertDocument** that handles document updates, which follows the same logic as this function, please use it instead.

its-mitesh-kumar: @JslYoon Done — removed `renameDocument` and consolidated into `upsertDocument` with an options object (`{ fileType?, fileId?, newTitle? }`). Added rollback protection, attribute preservation, defensive `fileType` fallback, and a 255-char title length cap at the router level.

One concern: rename latency on local itself is **1–2 min** due to the delete → re-create flow on the vector store (no in-place attribute update API). Can we do something here ?

its-mitesh-kumar:
> I noticed a slight delay in the file renaming process. It seems the non-atomic rename relies on a delete-then-create flow, which introduces a brief inconsistency window. I'd encourage @JslYoon to take a closer look and suggest any improvements if applicable.
> 
> Also, could you please address comments **1** and **2** from the Fullsend review? Specifically:
> 
> * **1:** The non-atomic rename (`delete` → `create`) introduces a brief inconsistency window.
> * **2:** Very long title names don't add much value; it would be better to consider adding a length cap.
> 
> [#3834 (comment)](https://github.com/redhat-developer/rhdh-plugins/pull/3834#issuecomment-5015363269)
> 
>  Screen.Recording.2026-07-20.at.1.24.16.PM.mov

@HusneShabbir 

Addressed both points:

1. **Non-atomic rename (delete → create):** This is an inherent limitation of the vector store API — there's no in-place attribute update endpoint. The rollback protection in `upsertDocument` mitigates the most likely failure (create throws after delete), but a brief inconsistency window remains unavoidable with the current API surface. On latency @JslYoon will check if we can do something here ?

2. **Title length cap:** Added a 255-character limit enforced at both the backend (PATCH endpoint returns 400) and frontend (inline validation with error feedback: "Name too long (max 255)").

its-mitesh-kumar:
@ShiranHi Handled all your below feedbacks and have updated the screen recording please have a look.

- Could we add a hover state to the field so users know it's editable, as in the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/)?
- We don't need the floating confirmation toast message on successful rename. Changes should auto-save as soon as the user clicks outside the field or presses Enter
- When selecting "Rename" from the 3-dots menu, the text field should immediately become active and focused.
- For the duplicate name error, let's center the error icon inside the field and shorten the copy to "Name already exists.

ShiranHi:
> @ShiranHi Handled all your below feedbacks and have updated the screen recording please have a look.
> 
> * Could we add a hover state to the field so users know it's editable, as in the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/)?
> * We don't need the floating confirmation toast message on successful rename. Changes should auto-save as soon as the user clicks outside the field or presses Enter
> * When selecting "Rename" from the 3-dots menu, the text field should immediately become active and focused.
> * For the duplicate name error, let's center the error icon inside the field and shorten the copy to "Name already exists.

LGTM, thank you!

JslYoon:
> > I noticed a slight delay in the file renaming process. It seems the non-atomic rename relies on a delete-then-create flow, which introduces a brief inconsistency window. I'd encourage @JslYoon to take a closer look and suggest any improvements if applicable.
> > Also, could you please address comments **1** and **2** from the Fullsend review? Specifically:
> > 
> > * **1:** The non-atomic rename (`delete` → `create`) introduces a brief inconsistency window.
> > * **2:** Very long title names don't add much value; it would be better to consider adding a length cap.
> > 
> > [#3834 (comment)](https://github.com/redhat-developer/rhdh-plugins/pull/3834#issuecomment-5015363269)
> > Screen.Recording.2026-07-20.at.1.24.16.PM.mov
> 
> @HusneShabbir
> 
> Addressed both points:
> 
> 1. **Non-atomic rename (delete → create):** This is an inherent limitation of the vector store API — there's no in-place attribute update endpoint. The rollback protection in `upsertDocument` mitigates the most likely failure (create throws after delete), but a brief inconsistency window remains unavoidable with the current API surface. On latency @JslYoon will check if we can do something here ?
> 2. **Title length cap:** Added a 255-character limit enforced at both the backend (PATCH endpoint returns 400) and frontend (inline validation with error feedback: "Name too long (max 255)").

Yes unfortunately this is a problem with the nature of how vector stores are configured in the upstream, and would require bigger code changes upstream for delete/insert to be faster. However, I'm addressing this issue right now through https://redhat.atlassian.net/browse/RHDHBUGS-3302

JslYoon (approved):
Backend changes look good

its-mitesh-kumar:
@ShiranHi @HusneShabbir @tobiastal
I made the changes to align document rename UX with notebook title rename: 

1. **single-click to edit** (was double-click), 
2. **optimistic cache update with rollback on API failure** : Now user will see the updated name instantly, if API failures happen it will get roll-backed.

**Happy flow**

https://github.com/user-attachments/assets/822dd724-d489-4ab3-b6b7-a135adc842a4

**API failure**

https://github.com/user-attachments/assets/63b22d18-161e-40be-963e-ff00916d929c

its-mitesh-kumar:
The document rename file was jumping to the top of the list. I made changes to preserve positions of documents in the sidebar

@JslYoon : Added `created_at` ascending sort in `listDocuments` to stabilize document ordering after rename. The delete → re-create flow changes the vector store's internal ordering, but since `upsertDocument` preserves the original `created_at` attribute, sorting by it keeps documents in their original upload order. Hope you are good with this change.

## UI before changes

https://github.com/user-attachments/assets/991d6dfe-625f-4a32-9cb6-b00e8faa42e2

## UI after changes

https://github.com/user-attachments/assets/60b808ec-9c14-42be-942f-65324fe37f92

Cc: @ShiranHi @HusneShabbir @tobiastal

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/notebooks/notebooksRouters.ts
```diff
@@ -439,6 +439,43 @@ export async function createNotebooksRouter(
     }),
   );
 
+  notebooksRouter.patch(
```
ciiay, line 442: **Requested change:** Please document this new PATCH endpoint in `plugins/intelligent-assistant-backend/README.md` under the Documents API Endpoints list (currently only PUT / GET / GET status / DELETE are listed).

Suggested entry:
```
PATCH /intelligent-assistant/notebooks/v1/sessions/:sessionId/documents/:documentId — Rename a document (JSON body: { "title": "new name" })
```

The route itself looks good; this is a docs-only gap for the public API change.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/DocumentSidebar.tsx
```diff
@@ -169,11 +234,87 @@ export const DocumentSidebar = ({
...
+      const newFullName = trimmedBase + extension;
+      if (newFullName.length > NOTEBOOK_MAX_TITLE_LENGTH) {
+        cancelEditing();
+        return;
+      }
+      if (
+        documents.some(d => d.document_id !== docId && d.title === newFullName)
+      ) {
+        cancelEditing();
+        return;
```
ciiay, line 298: **Requested change:** When the name is too long or conflicts, `saveRename` currently calls `cancelEditing()` and exits edit mode (on both Enter and blur).

Please keep the user in the inline editor when validation fails so the error helper text remains visible and they can correct the name. For example, early-return without `cancelEditing()` when `getValidationError(docId, originalTitle)` is non-null (empty / unchanged can still cancel as today).

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/translations/ref.ts
```diff
@@ -101,6 +101,11 @@ export const intelligentAssistantMessages = {
   'notebook.overwrite.modal.description':
     'The following files already exist in this notebook. Do you want to overwrite them with the new versions?',
   'notebook.overwrite.modal.action': 'Overwrite',
+  'notebook.document.rename': 'Rename',
+  'notebook.document.rename.tooltip': 'Click to rename',
+  'notebook.document.rename.error': 'Failed to rename "{{documentName}}".',
+  'notebook.document.rename.conflict': 'Name already exists.',
+  'notebook.document.rename.tooLong': 'Name too long (max 255)',
```
ciiay: **Nit:** Please add a trailing period for consistency with `notebook.document.rename.conflict` (`Name already exists.`) and the non-EN locales, e.g. `Name too long (max 255).`.

ciiay (changes requested):
Thanks for the work on this — the rename flow looks solid overall (optimistic update, validation, rollback, and the created_at ordering fix are appreciated).

Please address **#1** and **#3** before merge. **#2** is a nit only.

### Requested changes
1. **Document the new PATCH endpoint** in the backend README Documents API list (implementation is already in place; docs are missing).
3. **Keep the inline editor open when the name is invalid** (too long / conflict) on Enter or blur, instead of calling `cancelEditing()`.

### Nit
2. Add a trailing period to the EN `notebook.document.rename.tooLong` string for consistency with the conflict message and other locales.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/DocumentSidebar.tsx
its-mitesh-kumar: We intentionally cancel on blur/Enter when validation fails. The error is visible in real-time while the user types, giving them a chance to correct it before submitting. If they press Enter or click away with an invalid name, we treat it as an implicit cancel (same as Escape) and revert to the original name. This avoids trapping the user in the editor and matches common file-rename UX patterns (e.g. macOS Finder).
Cc: @ShiranHi @tobiastal

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/translations/ref.ts
its-mitesh-kumar: Done.

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/notebooks/notebooksRouters.ts
its-mitesh-kumar: Done.

HusneShabbir (commented):
Behaviour looks good to me, tests locally

/lgtm

ciiay (approved):
/lgtm

Thanks for the pr 👏

PR #3837, opened 2026-07-19
feat(intelligent-assistant): replace modal rename with inline edit on notebook cards

its-mitesh-kumar:
## Description
Replace the modal-based notebook rename workflow with an inline double-click-to-edit rename on notebook cards and the document sidebar. This provides a faster, more intuitive UX for renaming notebooks without a disruptive modal dialog. Also removes redundant PatternFly CSS custom property overrides in NotebookCard (11 → 8) by dropping values that match PF v6 defaults.

## Fixed
- [RHIDP-14854](https://redhat.atlassian.net/browse/RHIDP-14854)

## UI after changes

https://github.com/user-attachments/assets/eb685808-4040-4e15-894d-40929c1da54e

## ✔️ Checklist


HusneShabbir (commented):
working as expected

https://github.com/user-attachments/assets/1569f2af-0532-4717-8b11-d3e10c0e6e95

HusneShabbir (commented):
/lgtm

HusneShabbir (changes requested):
Wouldn't it be better if we prevented duplicating notebook renames? It would help maintain consistent application behavior and avoid confusion caused by users ending up with duplicate notebook tabs later on.

https://github.com/user-attachments/assets/da80fd87-93c3-4ead-a459-4cd2fb9c44da

cc: @ShiranHi / @karthikjeeyar

ShiranHi:
> Wouldn't it be better if we prevented duplicating notebook renames? It would help maintain consistent application behavior and avoid confusion caused by users ending up with duplicate notebook tabs later on.

@HusneShabbir I recommend we skip duplicate name validation for notebooks. Since users can also edit notebook names directly within the notebook view itself, adding restriction logic here could create a confusing experience. Allowing users the flexibility to rename notebooks freely, even if duplicate names occur, is standard practice and keeps the editing flow frictionless and easier.

HusneShabbir (approved):
fine then,
/lgtm

its-mitesh-kumar:
@ShiranHi Handled your feedback
1. Added hover Effect
2. On Blur(moving out) : Rename 
3. Toast message only on error


https://github.com/user-attachments/assets/c0a7375e-2aea-4c18-a85d-3a3a92012564

Cc: @HusneShabbir @karthikjeeyar

ShiranHi:
> @ShiranHi Handled your feedback
> 
> 1. Added hover Effect
> 2. On Blur(moving out) : Rename
> 3. Toast message only on error
> 
>  S_.2026-07-24.at.11.31.44.AM.mov 
> Cc: @HusneShabbir @karthikjeeyar

LGTM, thank you!

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -406,6 +407,31 @@ export const NotebookView = ({
...
+              notebookName: notebookName,
+            }) as string,
+            variant: 'danger',
+          },
+          ...prev,
+        ]);
+      }
+    },
+    [renameNotebook, sessionId, notebookName, t, queryClient],
+  );
```
karthikjeeyar: This function is nearly identical with a function in `LightSpeedChat.tsx`, with the same try/catch, the same `setXxxAlerts(prev => [...])` pattern

I would extract this into a new hook that wraps `useRenameNotebook` and accepts the differences as arguments from the caller function and returns a rename function already wired to produce toast alerts. This will keep the code DRY.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookCard.tsx
```diff
@@ -136,9 +174,31 @@ export const NotebookCard = ({
...
+          ) : (
+            <Typography
+              component="span"
+              className={classes.notebookTitleText}
+              title={t('notebook.rename.inline.tooltip')}
+              onClick={e => e.stopPropagation()}
+              onDoubleClick={e => {
+                e.stopPropagation();
+                startEditing();
+              }}
```
karthikjeeyar: Users see the hover highlight and cursor change, which signals `click me` but a single click does nothing (or navigates). That is confusing.


Honestly, As a user I would expect to hover and just a click on the title should enter into edit mode just like how @ShiranHi designed it in the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/intelligent-assistant/notebooks). FYI: NotebookLLM also has a single click edit mode for quick discoverability.

karthikjeeyar (commented):
In Inline editing, we should utilise the optimistic updates from react-query to immediately show the edited value in the card rather than waiting for the API to come back with a success. 

In current approach, when the network is slow or the backend service is slow for some reason, then user will not be able to see the edited value immediately:

https://github.com/user-attachments/assets/5436217f-8dcc-44d4-bbe5-d3dc6d282e22


Suggestion is to show the edited value immediately, incase the API failed and when toast appears we can rollback to the original value, This way the edit will seem very fast and intutive.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
karthikjeeyar, line 1: Can you please add test coverage for this `useInlineEdit` hook?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/translations/ref.ts
```diff
@@ -38,12 +38,8 @@ export const intelligentAssistantMessages = {
...
   'notebooks.actions.rename': 'Rename',
   'notebooks.actions.delete': 'Delete',
-  'notebooks.rename.title': 'Rename {{name}}?',
-  'notebooks.rename.description':
-    'Please input the new name for this notebook and click submit to proceed.',
-  'notebooks.rename.label': 'New name',
-  'notebooks.rename.placeholder': 'New name',
-  'notebooks.rename.action': 'Submit',
+  'notebook.rename.inline.tooltip': 'Double-click to rename',
+  'notebook.rename.inline.error': 'Failed to rename "{{notebookName}}".',
```
karthikjeeyar: Earlier keys were `notebooks` but these new keys are just `notebook`, update this for consistency.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
```diff
@@ -0,0 +1,86 @@
...
+  const savingRef = useRef(false);
+
+  const startEditing = useCallback(() => {
+    setIsEditing(true);
+    setEditValue(currentName);
+    onStart?.();
+    setTimeout(() => {
+      inputRef.current?.focus();
+      inputRef.current?.select();
+    }, 0);
```
karthikjeeyar:
```suggestion

```

Using `setTimeout(0)` is workaround and timeout pushes the focus call to the next event loop hoping the input UI has finished rendering.

However a better approach would be using effect to focus when isEditing is actually true.

```
 useEffect(() => {
    if (isEditing) {
      inputRef.current?.focus();
      inputRef.current?.select();
    }
  }, [isEditing]);

```

karthikjeeyar (commented):
A few more comments

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
karthikjeeyar: Also add a unit test for this new hook.

its-mitesh-kumar:
@karthikjeeyar @ShiranHi @HusneShabbir 
Now I have made changes to enable editing on single click itself. Also now in UI optimistic update ie as soon user click outside the updated notebook name will be shown, we are not waiting for api to get completed for updating it. If API calls fail we will show the toast message and show the previous name.

https://github.com/user-attachments/assets/cf035c3e-99f4-4615-84d0-93c062726111

*API failure*

https://github.com/user-attachments/assets/ef5f65b6-49bd-4091-b287-bcb4079f4b2a

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
its-mitesh-kumar: Added.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookCard.tsx
its-mitesh-kumar: Added edit flow on single click.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
its-mitesh-kumar: Done

ShiranHi:
> @karthikjeeyar @ShiranHi @HusneShabbir
> Now I have made changes to enable editing on single click itself. Also now in UI optimistic update ie as soon user click outside the updated notebook name will be shown, we are not waiting for api to get completed for updating it. If API calls fail we will show the toast message and show the previous name.

@its-mitesh-kumar it looks great! 
One small thing (probably not related to this PR), when hovering over the notebook cards in the main "Notebooks" tab, there's a slight jump between the card rows. Have you noticed this too? I have a feeling it's caused by the hover border on the card, it's subtle, but caught my eye.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
```diff
@@ -0,0 +1,89 @@
...
+    if (savingRef.current) return;
+    const trimmed = editValue.trim();
+    if (!trimmed || trimmed === currentName) {
+      cancelEditing();
+      return;
+    }
+    savingRef.current = true;
+    onSave(trimmed);
+    cancelEditing();
+    savingRef.current = false;
```
karthikjeeyar: `savingRef` does nothing here. It is set to `true` and reset to `false` synchronously within the same save() call. I understand this is mean't to prevent double-save, but `cancelEditing()` already handles this by setting the `editValue` to  empty string.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
```diff
@@ -0,0 +1,89 @@
...
+
+export const useInlineEdit = ({
+  currentName,
+  onSave,
+  onStart,
+}: UseInlineEditOptions) => {
+  const [isEditing, setIsEditing] = useState(false);
+  const [editValue, setEditValue] = useState('');
+  const inputRef = useRef<HTMLInputElement>(null);
+  const savingRef = useRef(false);
```
karthikjeeyar: Not required based on my previous comment.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/__tests__/useInlineEdit.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      });
+      act(() => {
+        result.current.save();
+      });
+
+      expect(onSave).toHaveBeenCalledTimes(1);
+      expect(onSave).toHaveBeenCalledWith('New Name');
+      expect(result.current.isEditing).toBe(false);
+    });
+  });
```
karthikjeeyar: This test doesn't acutally test the guard, it calls `result.current.save();` only once and even if you remove the `savingRef` from the `useInlineEdit` hook, this test will still pass. You can remove this safely.

its-mitesh-kumar:
> > @karthikjeeyar @ShiranHi @HusneShabbir
> > Now I have made changes to enable editing on single click itself. Also now in UI optimistic update ie as soon user click outside the updated notebook name will be shown, we are not waiting for api to get completed for updating it. If API calls fail we will show the toast message and show the previous name.
> 
> @its-mitesh-kumar it looks great! One small thing (probably not related to this PR), when hovering over the notebook cards in the main "Notebooks" tab, there's a slight jump between the card rows. Have you noticed this too? I have a feeling it's caused by the hover border on the card, it's subtle, but caught my eye.

@ShiranHi I have raised a bug for tracking it https://redhat.atlassian.net/browse/RHDHBUGS-3612 .

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/__tests__/useInlineEdit.test.ts
its-mitesh-kumar: Good catch, removed.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
its-mitesh-kumar: Good catch, removed.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/hooks/notebooks/useInlineEdit.ts
its-mitesh-kumar: Good catch, removed.

karthikjeeyar (approved):
/lgtm

PR #3842, opened 2026-07-19
fix(adoption-insights): resolve four accessibility violations in e2e tests

### LINKED ISSUE #3835: Accessibility Violations Detected in Adoption Insights
## Description

Accessibility tests for Adoption Insights (axe with WCAG 2.x tags) are currently passing only because certain violations are **filtered out** in e2e. The real issues still need to be fixed in the app.

### Known issues (currently suppressed in e2e)

1. **`aria-input-field-name`** – Form inputs missing accessible names (e.g. missing `aria-label`, `aria-labelledby`, or associated `<label>`).
2. **`button-name`** – Buttons without discernible text (e.g. icon-only buttons with no `aria-label` or visible text).
3. **`color-contrast`** – MUI Tooltip default styling does not meet contrast requirements; tooltip text/background fails axe's contrast rule.
4. **`svg-img-alt`** – Recharts pie/donut chart segments are `<path role="img">` with a `name` attribute but no proper accessible name (e.g. no `aria-label`, `title`, or `<title>`). Chart data is partly conveyed by legend/labels, but the SVG elements themselves are not fully accessible.

Image
Image

Image

## Steps to Reproduce

1. Navigate to the adoption-insights workspace:
   ```bash
   cd workspaces/adoption-insights
   ```
2. Install dependencies:
   ```bash
   yarn install
   ```
3. Install Playwright browsers (if not already installed):
   ```bash
   yarn playwright install chromium
   ```
4. Remove the suppressed rule IDs from the filter list in `e2e-tests/utils/accessibility.ts`
5. Run the e2e tests:
   ```bash
   yarn ui-test
   ```
6. Observe the `aria-input-field-name`, `button-name`, `color-contrast`, and `svg-img-alt` violations that are now no longer suppressed.

## Expected results

- All of the above are addressed in the Adoption Insights UI (ARIA labels, button names, tooltip styling, and chart accessibility) so that they pass axe and meet WCAG 2.0/2.1 Level A and AA accessibility standards.
- After verification, the **temporary filter logic** in `workspaces/adoption-insights/e2e-tests/utils/accessibility.ts` is reverted (the rule IDs removed from the filter list) so that the e2e suite enforces full accessibility and does not hide real violations.

## Reproducibility

Always

its-mitesh-kumar:
## Description

Fix four accessibility violations that were suppressed by an exclusion filter in `e2e-tests/utils/accessibility.ts` instead of being addressed at the component level. After fixing all violations, the filter has been removed so the e2e suite now enforces full WCAG 2.x accessibility compliance.

### Root cause
Four accessibility violations were suppressed in the e2e accessibility filter (`e2e-tests/utils/accessibility.ts`) instead of being fixed:
1. **`aria-input-field-name`** — The header date-range `<Select>` and `TablePagination` select rendered `<div role="combobox">` elements without accessible names. MUI's `aria-label` prop goes on the outer wrapper div (which has no valid role), not the inner combobox. Fixed by using `inputProps={{ 'aria-label': ... }}` which targets the correct element.
2. **`button-name`** — The info `<IconButton>` in the Users card had no `aria-label`, making it invisible to screen readers. Fixed by adding `aria-label={t('users.tooltip')}`.
3. **`color-contrast`** — MUI Tooltip default styling uses a semi-transparent background (`rgba(97,97,97,0.92)`) that fails WCAG AA contrast ratio (4.09:1 vs required 4.5:1). Fixed by applying high-contrast styling (`bgcolor: '#151515'`, `color: '#fff'`) via `slotProps` on all data table Tooltips (CatalogEntities, Templates, Techdocs). Also added `disableInteractive` to prevent tooltip stacking issues.
4. **`svg-img-alt`** — Recharts' `<Pie>` component renders `<path role="img" name="...">` for pie segments, but the `name` attribute is not a valid accessible name for elements with `role="img"`. Fixed by passing `aria-label={entry.name}` through `<Cell>` components.

After fixing all violations, the filter was removed from `accessibility.ts` so the e2e suite now enforces full WCAG 2.x accessibility. Also fixed a pre-existing flaky tooltip e2e test that used `page.getByRole('tooltip')` without name filtering, causing strict-mode violations when multiple tooltips coexisted in the DOM.

## Fixed
- Fixes https://github.com/redhat-developer/rhdh-plugins/issues/3835

## UI before changes
Before fix

## UI after changes
After fix

## Test Plan
- [ ] Run `yarn tsc:full` — type check passes
- [ ] Run `yarn test --watchAll=false` — all 341 unit tests pass
- [ ] Run `APP_MODE=legacy npx playwright test e2e-tests/insights.test.ts --project=en` — all 10 e2e tests pass including unfiltered accessibility checks
- [ ] Verify axe scan reports zero violations for `aria-input-field-name`, `button-name`, `color-contrast`, and `svg-img-alt`
- [ ] Verify the header date-range Select has an accessible name via screen reader or axe
- [ ] Verify the Users info IconButton has an accessible name
- [ ] Verify MUI Tooltip text has sufficient contrast (white on #151515)
- [ ] Verify pie chart segments have `aria-label` attributes

## Checklist


HusneShabbir (approved):
/lgtm

teknaS47 (commented):
Running the test locally is giving me one accessibility violation: Ensure the contrast between foreground and background colors meets WCAG 2 AA minimum contrast ratio thresholds

teknaS47 (approved):
All tests including accessibility tests passed.

Image

PR #3843, opened 2026-07-19
fix(extensions): fix accessibility violations in Extensions plugin

### LINKED ISSUE #3838: Accessibility violations for Extensions plugin
## Description

Accessibility tests for the Extensions (Marketplace) plugin revealed several issues, including the document not having a non-empty title value. The e2e accessibility tests currently pass only because violations are suppressed via a `skipFailures` flag in `workspaces/extensions/e2e-tests/utils/accessibility.ts`. The real issues still need to be fixed in the app.

Once all accessibility issues have been addressed and verified, the temporary accessibility logic adjustments applied during testing should be reverted to their normal state to restore standard functionality.

## Steps to Reproduce

1. Navigate to the extensions workspace:
   ```bash
   cd workspaces/extensions
   ```
2. Install dependencies:
   ```bash
   yarn install
   ```
3. Remove the `skipFailures` flag from the `runAccessibilityTests` options in `e2e-tests/utils/accessibility.ts`
4. Run the Playwright e2e tests:
   ```bash
   CI=1 yarn test:e2e
   ```
5. Observe the WCAG violations that are now no longer suppressed.

## Expected results

- All accessibility violations are fixed in the Extensions plugin UI so that they pass axe-core with WCAG tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`.
- The **temporary filter logic** in `e2e-tests/utils/accessibility.ts` is reverted (the `skipFailures` flag and suppressed rule IDs removed) so that the e2e suite enforces full accessibility and does not hide real violations.

## Reproducibility

Always

its-mitesh-kumar:
## Description

Fix three accessibility violations in the Extensions plugin that were suppressed by the `skipFailures` flag in e2e accessibility tests:

1. **role-img-alt** — MUI `CardMedia` elements with `role="img"` lacked an accessible name. Added `title` attribute using the plugin display name to `PluginIcon`.
2. **aria-prohibited-attr** — Badge triangle tooltip wrappers (`BadgeTriange`) had `aria-label` forwarded to a plain `<div>` with no role. Added `role="img"` and explicit `aria-label` to the tooltip child element.
3. **label** — Filter combobox inputs (Category, Author, Support type) had no associated labels. Added `id` prop to `Autocomplete` and `htmlFor` to `InputLabel` in `CustomSelectFilter` to create proper label associations.

Also removed the `skipFailures` flag from `e2e-tests/utils/accessibility.ts` so the e2e suite now enforces full WCAG 2.x accessibility.

## Fixed
- Fixes https://github.com/redhat-developer/rhdh-plugins/issues/3838

## UI before changes
Before fix

## UI after changes
After fix

## Checklist

### DIFF
--- a/workspaces/extensions/e2e-tests/utils/accessibility.ts
+++ b/workspaces/extensions/e2e-tests/utils/accessibility.ts
@@ -21,21 +21,18 @@ export async function runAccessibilityTests(
   page: Page,
   testInfo: TestInfo,
   attachName = 'accessibility-scan-results.json',
-  options: { skipFailures?: boolean } = { skipFailures: true },
 ) {
   const accessibilityScanResults = await new AxeBuilder({ page })
     .withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa'])
     .analyze();
 
   await testInfo.attach(attachName, {
-    body: JSON.stringify(accessibilityScanResults.violations, null, 2),
+    body: JSON.stringify(accessibilityScanResults, null, 2),
     contentType: 'application/json',
   });
 
-  if (!options?.skipFailures) {
-    expect(
-      accessibilityScanResults.violations,
-      'Accessibility violations found',
-    ).toEqual([]);
-  }
+  expect(
+    accessibilityScanResults.violations,
+    'Accessibility violations found',
+  ).toEqual([]);
 }
--- a/workspaces/extensions/.changeset/bright-waves-glow.md
+++ b/workspaces/extensions/.changeset/bright-waves-glow.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-extensions': patch
+---
+
+Fix accessibility violations (role-img-alt, aria-prohibited-attr, label) and remove suppression flag from e2e accessibility tests.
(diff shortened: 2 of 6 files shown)


workspaces/extensions/e2e-tests/utils/accessibility.ts
```diff
@@ -21,21 +21,18 @@ export async function runAccessibilityTests(
...
 
-  if (!options?.skipFailures) {
-    expect(
-      accessibilityScanResults.violations,
-      'Accessibility violations found',
-    ).toEqual([]);
-  }
+  expect(
+    accessibilityScanResults.violations,
+    'Accessibility violations found',
```
sanketpathak, line 36: This message is confusing even when there are no accessibility issues found. Change the message to something like this:
  `Accessibility check (${accessibilityScanResults.violations.length} violations)`,
so it will show the check o for no violations found and the number of violations recorded for failures.

sanketpathak (changes requested):

teknaS47 (approved):
All tests including accessibility tests passed.

Image

workspaces/extensions/e2e-tests/utils/accessibility.ts
sanketpathak: Created a bug to handle this for all plugins separately https://redhat.atlassian.net/browse/RHDHBUGS-3490

teknaS47 (changes requested):
The `.gif` files are still present

HusneShabbir (approved):

teknaS47 (approved):
/lgtm

sanketpathak (approved):

PR #3844, opened 2026-07-19
fix(quickstart): fix accessibility violations in Quick Start plugin

### LINKED ISSUE #3839: Accessibility violations for Quick Start plugin
## Description

Accessibility tests for the Quick Start plugin revealed several issues, including missing ARIA labels and buttons without discernible text. The e2e accessibility tests currently pass only because violations are suppressed via a `skipViolationsAssert` flag in `e2e-tests/utils/accessibility.ts`. The real issues still need to be fixed in the app.

Once all accessibility issues have been addressed and verified, the temporary **accessibility logic adjustments** applied during testing should be **reverted to their normal state** to restore standard functionality.

[quick-start-accessibility.json](https://github.com/user-attachments/files/30169169/quick-start-accessibility.json)

## Steps to Reproduce

1. Navigate to the quickstart workspace:
   ```bash
   cd workspaces/quickstart
   ```
2. Install dependencies:
   ```bash
   yarn install
   ```
3. Remove the `skipViolationsAssert` attribute from the `options` parameter of `runAccessibilityTests` in `e2e-tests/utils/accessibility.ts`
4. Run the Playwright e2e tests:
   ```bash
   CI=1 yarn test:e2e
   ```
5. Observe the WCAG violations that are now no longer suppressed.

## Expected results

- All accessibility violations are fixed in the Quick Start plugin UI so that they pass axe-core with WCAG tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`.
- The **temporary filter logic** in `e2e-tests/utils/accessibility.ts` is reverted (the `skipViolationsAssert` flag removed) so that the e2e suite enforces full accessibility and does not hide real violations.

## Reproducibility

Always

its-mitesh-kumar:
## Description

Fix three accessibility violations in the Quick Start plugin detected by axe-core:
- **aria-progressbar-name**: `LinearProgress` in `QuickstartFooter` had no accessible name — added `aria-label` using the existing progress/not-started translation text.
- **list**: `<ul>` (MUI `List`) in `QuickstartContent` had direct `<div>` children from the `Box` wrapper in `QuickstartItem` — changed `Box` to render as `<li>` and `ListItem` to render as `<div>`.
- **nested-interactive**: `ListItem` with `role="button"` contained a focusable `IconButton` — replaced `IconButton` with a non-interactive `<span>` since the parent already handles all interaction.

Also removed the `skipViolationsAssert` suppression flag from e2e accessibility tests, replacing it with an `exclude` option to skip the catalog table (which has Backstage-core `nested-interactive` violations outside this plugin's scope).

### Root cause

The Quick Start drawer's accessibility tests were set to `skipViolationsAssert: true`, masking real WCAG 2.x violations. The violations stemmed from MUI components not having proper ARIA attributes, incorrect HTML list structure (div inside ul), and a button nested inside another interactive element.

## Fixed
- Fixes https://github.com/redhat-developer/rhdh-plugins/issues/3839

## Test Plan
- [x] axe-core e2e scan passes with violations assert enabled (no more `skipViolationsAssert`)
- [x] Admin-guest e2e test passes (no regression)
- [x] All 76 unit tests pass
- [x] TypeScript compiles cleanly
- [x] Build completes successfully

## Checklist

### DIFF
--- a/workspaces/quickstart/e2e-tests/utils/accessibility.ts
+++ b/workspaces/quickstart/e2e-tests/utils/accessibility.ts
@@ -20,23 +20,30 @@ export async function runAccessibilityTests(
   page: Page,
   testInfo: TestInfo,
   attachName = 'accessibility-scan-results.json',
-  options: { skipViolationsAssert?: boolean; attachName?: string } = {
-    skipViolationsAssert: false,
-  },
+  options: { exclude?: string[] } = {},
 ) {
-  const accessibilityScanResults = await new AxeBuilder({ page })
-    .withTags(['wcag2a', 'wcag2aa', 'wcag21a', 'wcag21aa'])
-    .analyze();
+  let builder = new AxeBuilder({ page }).withTags([
+    'wcag2a',
+    'wcag2aa',
+    'wcag21a',
+    'wcag21aa',
+  ]);
+
+  if (options.exclude) {
+    for (const selector of options.exclude) {
+      builder = builder.exclude(selector);
+    }
+  }
+
+  const accessibilityScanResults = await builder.analyze();
 
   await testInfo.attach(attachName, {
     body: JSON.stringify(accessibilityScanResults, null, 2),
     contentType: 'application/json',
   });
 
-  if (!options?.skipViolationsAssert) {
-    expect(
-      accessibilityScanResults.violations,
-      'Accessibility violations found',
-    ).toEqual([]);
-  }
+  expect(
+    accessibilityScanResults.violations,
+    'Accessibility violations found',
+  ).toEqual([]);
 }
--- a/workspaces/quickstart/.changeset/gentle-rivers-flow.md
+++ b/workspaces/quickstart/.changeset/gentle-rivers-flow.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-quickstart': patch
+---
+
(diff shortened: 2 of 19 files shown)


workspaces/quickstart/e2e-tests/utils/accessibility.ts
```diff
@@ -20,23 +20,30 @@ export async function runAccessibilityTests(
...
 
-  if (!options?.skipViolationsAssert) {
-    expect(
-      accessibilityScanResults.violations,
-      'Accessibility violations found',
-    ).toEqual([]);
-  }
+  expect(
+    accessibilityScanResults.violations,
+    'Accessibility violations found',
```
sanketpathak, line 47: This message is confusing even when no accessibility issues are found. Change the message to something like this:
  `Accessibility check (${accessibilityScanResults.violations.length} violations)`,
so it will show the check o for no violations found and the number of violations recorded for failures.

sanketpathak: Created a bug to handle this for all plugins separately https://redhat.atlassian.net/browse/RHDHBUGS-3490

sanketpathak:
Accessibility violations are found when testing locally, 
[quick-start-accessibility.json](https://github.com/user-attachments/files/30220362/quick-start-accessibility.json)

teknaS47 (commented):
Running the test locally is giving me one accessibility violation: `Ensure buttons have discernible text`

teknaS47 (approved):
/lgtm

Image

PR #3846, opened 2026-07-19
fix(orchestrator): fix accessibility violations in WorkflowSuccessRatioCell

### LINKED ISSUE #3845: Accessibility violations for Orchestrator plugin
## Description

Accessibility tests for the Orchestrator plugin revealed several issues, including the document not having a non-empty title value. These need to be reviewed and fixed to ensure compliance with accessibility standards.

Once all accessibility issues have been addressed and verified, the temporary accessibility logic adjustments applied during testing should be reverted to their normal state to restore standard functionality.

[accessibility-violation-all-runs-orchestrator.docx](https://github.com/user-attachments/files/30169539/accessibility-violation-all-runs-orchestrator.docx)
[aceesibility-violation-workflow-page-orchestrator.docx](https://github.com/user-attachments/files/30169540/aceesibility-violation-workflow-page-orchestrator.docx)

## Steps to Reproduce

1. Navigate to the orchestrator workspace:
   ```bash
   cd workspaces/orchestrator
   ```
2. Install dependencies:
   ```bash
   yarn install
   ```
3. Install Playwright browsers (if not already installed):
   ```bash
   yarn playwright install chromium
   ```
4. Remove any `skipViolationsAssert` or `skipFailures` flags from the accessibility test utility
5. Run the Playwright e2e tests:
   ```bash
   CI=1 yarn test:e2e
   ```
6. Observe the WCAG violations that are now no longer suppressed.

## Expected results

- All accessibility violations are fixed in the Orchestrator plugin UI so that they pass axe-core with WCAG tags `wcag2a`, `wcag2aa`, `wcag21a`, `wcag21aa`.
- Any temporary filter logic in the e2e accessibility utils is reverted so that the e2e suite enforces full accessibility and does not hide real violations.

## Reproducibility

Always

its-mitesh-kumar:
## Description

Fix accessibility violations in the `WorkflowSuccessRatioCell` component of the Orchestrator plugin. The `MuiCircularProgress` elements used to display success ratios in the workflows table were missing accessible names (`aria-progressbar-name` violation), and a wrapper `<div>` had an `aria-label` attribute without a valid ARIA role (`aria-prohibited-attr` violation).

### Root cause

`WorkflowSuccessRatioCell` renders two overlapping `CircularProgress` components (a background track and a foreground fill) inside a `Box` wrapper. The wrapper `<div>` had `aria-label` set directly, which violates ARIA rules since plain `<div>` elements without a role cannot have `aria-label`. Neither `CircularProgress` element had an accessible name, triggering `aria-progressbar-name` violations. The fix removes `aria-label` from the wrapper, marks the decorative track progress as `aria-hidden="true"`, and adds a descriptive `aria-label` to the foreground progress indicator.

Note: The `nested-interactive` violations reported from `MuiTableSortLabel` are from Backstage core's table component and cannot be fixed from the orchestrator plugin.

## Fixed
- Fixes https://github.com/redhat-developer/rhdh-plugins/issues/3845

## Test Plan
- [ ] Navigate to the Orchestrator workflows page
- [ ] Verify the success ratio column renders circular progress indicators correctly
- [ ] Run axe-core accessibility scan and verify no `aria-progressbar-name` or `aria-prohibited-attr` violations
- [ ] Verify `nested-interactive` violations (from Backstage core table) are the only remaining violations

## Checklist

### DIFF
--- a/workspaces/orchestrator/.changeset/bright-clouds-shine.md
+++ b/workspaces/orchestrator/.changeset/bright-clouds-shine.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Fix accessibility violations (aria-progressbar-name, aria-prohibited-attr) in WorkflowSuccessRatioCell component.
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/ui/WorkflowSuccessRatioCell.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/ui/WorkflowSuccessRatioCell.tsx
@@ -59,14 +59,12 @@ export const WorkflowSuccessRatioCell = ({
 
   return (
     <Box sx={{ display: 'flex', alignItems: 'center', gap: 1 }}>
-      <Box
-        sx={{ position: 'relative', display: 'inline-flex' }}
-        aria-label={`${percent}%`}
-      >
+      <Box sx={{ position: 'relative', display: 'inline-flex' }}>
         <CircularProgress
           variant="determinate"
           value={100}
           size={PROGRESS_SIZE}
+          aria-hidden="true"
           sx={{ color: trackColor }}
         />
         <CircularProgress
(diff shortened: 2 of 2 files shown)


lokanandaprabhu (approved):
/lgtm

PR #3847, opened 2026-07-19
fix(global-header): restore keyboard navigation for menu items without links

its-mitesh-kumar:
## Summary

- **[RHDHBUGS-1931]** Fix keyboard accessibility in Help dropdown menu — arrow keys now navigate all items correctly
- `MenuItem` with `component={Fragment}` silently dropped `role="menuitem"` and `tabIndex`, making items without a `link` prop (e.g. `SupportButton`) invisible to MUI's keyboard navigation
- Section headers without links now render `<ListSubheader>` directly instead of wrapping in a no-op `<MenuItem component={Fragment}>`

## Root Cause

In `MenuSection.tsx`, when a menu item had no `link` prop, `MenuItem` was rendered with `component={Fragment}`. React's `Fragment` drops all props passed to it, so MUI's injected `role="menuitem"` and `tabIndex` attributes were never emitted to the DOM. MUI's `MenuList` uses sibling traversal with `nextElementSibling` to handle ArrowUp/ArrowDown, and elements without these attributes are skipped — making those items unreachable by keyboard.

The Help dropdown's `SupportButton` (which manages its own navigation internally) has no `link` in its mount point config, so it was always rendered via `Fragment`.

## Test plan

- [ ] Open Help dropdown — verify ArrowUp/ArrowDown navigates between all items
- [ ] Open Application Launcher dropdown — verify arrow key navigation still works (regression check)
- [ ] Verify section headers with links (e.g. App Launcher "Documentation" header) remain clickable
- [ ] Verify section headers without links are not keyboard-focusable (correct a11y behavior)
- [ ] Verify mouse clicks on all menu items still work as before

### DIFF
--- a/workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/MenuSection.tsx
+++ b/workspaces/global-header/plugins/global-header/src/components/HeaderDropdownComponent/MenuSection.tsx
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import { Fragment } from 'react';
 import type { ComponentType, FC } from 'react';
 import Divider from '@mui/material/Divider';
 import Box from '@mui/material/Box';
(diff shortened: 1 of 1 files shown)


its-mitesh-kumar:
Closing — will recreate using the raise-pr skill for proper changeset and PR template.

PR #3848, opened 2026-07-19
fix(global-header): restore keyboard navigation for menu items without links

its-mitesh-kumar:
## Description

`MenuSection` used `component={Fragment}` on MUI `MenuItem` when items had no `link` prop. React's `Fragment` silently drops all props, so MUI's `role="menuitem"` and `tabIndex` attributes were never rendered to the DOM. This made those items invisible to MUI `MenuList` arrow-key traversal. The Help dropdown's `SupportButton` (which manages navigation internally and has no `link` in its mount point config) was the primary victim.

The fix conditionally spreads `component`/`to` only when a link exists, letting `MenuItem` render its default `<li>` with proper ARIA attributes. Non-clickable section headers now render `<ListSubheader>` directly instead of wrapping in a no-op `<MenuItem component={Fragment}>`.

## Fixed
- [RHDHBUGS-1931](https://redhat.atlassian.net/browse/RHDHBUGS-1931) — [global-header] Menu items in Help menu can't be accessed with a keyboard

## Checklist

### DIFF
--- a/workspaces/global-header/.changeset/gentle-menus-focus.md
+++ b/workspaces/global-header/.changeset/gentle-menus-focus.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-header': patch
+---
+
+Fix keyboard navigation for menu items without links in Help dropdown by removing Fragment-based MenuItem rendering that silently dropped role and tabIndex attributes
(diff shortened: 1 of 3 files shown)


its-mitesh-kumar:
/hold
Lets wait for accessibility fix of global-header by @rohitratannagar then we will revisit and check ifwe still need this PR.
Cc: @sanketpathak @teknaS47

HusneShabbir (commented):
Pushed a small e2e fix in `a84198f` — the Help test was hitting Playwright strict mode because `support-button` matched both the catalog Support control and the Help menu item. Scoped the assertion to the open menu (and kept the translated label check). CI should be green with this.

PR #3892, opened 2026-07-21
orchestrator - version:bump to v1.52.0

lokanandaprabhu:
Backstage release v1.52.0 has been published, this Pull Request contains the changes to upgrade orchestrator to this new release.

Please review the changelog before approving, there may be manual changes needed:

- Changelog: [v1.52.0](https://github.com/backstage/backstage/blob/master/docs/releases/v1.52.0-changelog.md)
- Upgrade Helper: [From 1.51.1 to 1.52.0](https://backstage.github.io/upgrade-helper/?from=1.51.1&to=1.52.0)

### 1. `config:check` failure — `@backstage/plugin-kubernetes-react` config.d.ts

The `validate config` CI step fails with:
```
Error: Invalid TypeScript configuration schema:
node_modules/@backstage/plugin-kubernetes-react/config.d.ts(1,33):
  error TS2307: Cannot find module './src/components/PodExecTerminal/PodExecTerminal'
```

**Root cause:** `config.d.ts` has a stale, unused `import { PodExecTerminal } from './src/...'` — the `./src` directory isn't shipped in the published package. Same class of bug as the theme plugin ([#3792](https://github.com/redhat-developer/rhdh-plugins/pull/3792), [#3814](https://github.com/redhat-developer/rhdh-plugins/pull/3814)).

**Upstream fix:** [backstage/backstage#34721](https://github.com/backstage/backstage/pull/34721) (merged, shipped in `0.5.x`). However, the transitive dependency `@janus-idp/shared-react` pins `^0.4.4`, so our lockfile resolves to `0.4.4` which is still broken.

**Workaround:** Added a `sed` patch in `scripts/postinstall` to strip the stale import after `yarn install`. Safe to remove once `@janus-idp/shared-react` updates its range to pick up `>=0.5.0`.

### 2. Oversized Run ID copy button

The copy button next to the Run ID field renders oversized after the bump.

**Root cause:** Backstage 1.52.0's `CopyTextButton` (from `@backstage/core-components`) switched from a MUI `IconButton` to `ButtonIcon` from `@backstage/ui`, which uses scoped CSS (`width: 2rem; height: 2rem` via `data-size` attributes) that cannot be overridden from consumer code. The previous CSS overrides targeting `& > button` no longer apply.

**Fix:** Replaced `CopyTextButton` with a MUI `IconButton` + `ContentCopyIcon` + `Tooltip` — the same pattern used across other workspaces (augment, boost, dcm). Added `workflow.fields.workflowIdCopy` translation key across all 6 languages.


lokanandaprabhu:
Screenshot 2026-07-21 at 10 53 28 AM

Application is working fine.

workspaces/orchestrator/plugins/orchestrator-backend/package.json
```diff
@@ -94,8 +94,8 @@
   },
   "devDependencies": {
     "@backstage-community/plugin-rbac-common": "^1.12.1",
-    "@backstage/backend-test-utils": "^1.11.3",
-    "@backstage/cli": "^0.36.2",
+    "@backstage/backend-test-utils": "^1.11.4",
+    "@backstage/cli": "^0.36.3",
     "@janus-idp/backstage-plugin-audit-log-node": "^1.7.1",
     "@janus-idp/cli": "3.7.0",
```
christoph-jerolimov, line 100: Can you please create follow up stories that get rid of `@janus-idp/backstage-plugin-audit-log-node` and `@janus-idp/cli`

workspaces/orchestrator/scripts/postinstall
```diff
@@ -2,5 +2,14 @@
...
 cd ../../ && yarn install
 cd $mydir
+
+# Fix @backstage/plugin-kubernetes-react config.d.ts: stale import from ./src
+# that doesn't exist in the published package, breaking config:check.
+# Fixed upstream in backstage/backstage#34721; remove once shared-react updates its range.
+k8s_config="node_modules/@backstage/plugin-kubernetes-react/config.d.ts"
+if [ -f "$k8s_config" ] && grep -q "from './src/" "$k8s_config"; then
+  sed -i.bak "/from '\.\/src\//d" "$k8s_config" && rm -f "${k8s_config}.bak"
+fi
```
christoph-jerolimov: Instead of this workaround it should be possible that you change the `@backstage/config-loader` back to 0.10 in the yarn.lock or via a resolution in the package.json.

christoph-jerolimov (commented):
I didn't request a change on GH here, but maybe you can try the right `@backstage/config-loader` from Backstage 1.52 instead of the post-install workaround. When someone else wants merge it this way, its fine as interim solution.

workspaces/orchestrator/plugins/orchestrator-backend/package.json
lokanandaprabhu: https://redhat.atlassian.net/browse/RHIDP-15676 created

workspaces/orchestrator/scripts/postinstall
lokanandaprabhu: Updated the PR

lholmquist (approved):
Ran a workflow and the changes worked fine

PR #3908, opened 2026-07-21
fix(ci): allow fork PR checkout in changeset feedback workflow

rohitkrai03:
## Summary

- GitHub Actions now blocks checking out fork PR code from `pull_request_target` workflows by default
- The `automate_changeset_feedback.yml` workflow fails for all fork PRs with: _"Refusing to check out fork pull request code from a 'pull_request_target' workflow"_
- Adds `allow-unsafe-pr-checkout: true` to the `actions/checkout` step

See: https://gh.io/securely-using-pull_request_target

### DIFF
--- a/.github/workflows/automate_changeset_feedback.yml
+++ b/.github/workflows/automate_changeset_feedback.yml
@@ -33,6 +33,8 @@ jobs:
           # Fetch the commit that's merged into the base rather than the target ref
           # This will let us diff only the contents of the PR, without fetching more history
           ref: 'refs/pull/${{ github.event.pull_request.number }}/merge'
+          allow-unsafe-pr-checkout: true
+          persist-credentials: false
       - name: fetch base
         run: git fetch --depth 1 origin "$GITHUB_BASE_REF"
 


.github/workflows/automate_changeset_feedback.yml
```diff
@@ -33,6 +33,7 @@ jobs:
           # Fetch the commit that's merged into the base rather than the target ref
           # This will let us diff only the contents of the PR, without fetching more history
           ref: 'refs/pull/${{ github.event.pull_request.number }}/merge'
+          allow-unsafe-pr-checkout: true
```
kim-tsao, line 36:
```suggestion
          allow-unsafe-pr-checkout: true
          persist-credentials: false
```

kim-tsao (dismissed):
ran a cursor review and unintentionally created a [PR](https://github.com/redhat-developer/rhdh-plugins/pull/3910) but this is generally safe to add combined with the config to not store creds

.github/workflows/automate_changeset_feedback.yml
christoph-jerolimov: @rohitkrai03 addressed your comment @kim-tsao, as its a blocker I like to merge this now. I open a thread so that we don't miss this.

christoph-jerolimov (approved):

PR #3909, opened 2026-07-21
chore: remove broken postinstall scripts

jonkoops:
Every workspace inherited a `"postinstall": "cd ../../ && yarn install"` script from the community-plugins repo convention. This was a workaround to ensure root-level dependencies were installed when CI ran `yarn install` inside a workspace directory.

This workaround was superseded upstream by backstage/community-plugins#1903, which added an explicit root install step to the CI workflows. Our release workflow already has an equivalent step, but the CI workflow was missing one. Add it now so the postinstall can be safely removed.

Workspaces in this repo are independent projects and should not depend on the parent package or other workspaces. This postinstall violated that boundary.

### DIFF
--- a/workspaces/orchestrator/scripts/postinstall
+++ b/workspaces/orchestrator/scripts/postinstall
@@ -1,6 +1,0 @@
-#!/bin/bash
-mydir=`pwd`
-cd ../../ && yarn install
-cd $mydir
-cd plugins/orchestrator-common
-yarn openapi:check
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,10 @@ jobs:
           node-version: ${{ matrix.node-version }}
           registry-url: https://registry.npmjs.org/ # Needed for auth
 
+      - name: Install root dependencies
+        run: yarn install --immutable
+        working-directory: ${{ github.workspace }}
+
       - name: yarn install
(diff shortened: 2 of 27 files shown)


gabemontero (approved):

christoph-jerolimov (approved):

christoph-jerolimov:
> Workspaces in this repo are independent projects and should not depend on the parent package or other workspaces. This postinstall violated that boundary.

But it still depend on the fact that the root deps are installed, or?

workspaces/orchestrator/scripts/postinstall
```diff
@@ -1,6 +0,0 @@
-#!/bin/bash
-mydir=`pwd`
-cd ../../ && yarn install
-cd $mydir
-cd plugins/orchestrator-common
-yarn openapi:check
```
christoph-jerolimov, line 6: Don't know if orchestrator people would depend on this. I guess this openapi:check could be also part of a build step?!

What's your thought here @lokanandaprabhu @lholmquist  @rostalan

jonkoops:
Too busy to keep working on this, so dropping for now

jonkoops:
> But it still depend on the fact that the root deps are installed, or?

Unfortunately, yes. But ideally the root should not have any dependencies.

PR #3413, opened 2026-06-16
feat(global-header): Promote global-header NFS plugin from /alpha to …

rohitratannagar:
## Summary

**Jira: https://redhat.atlassian.net/browse/RHIDP-14169**

### after changes

https://github.com/user-attachments/assets/4a9f416e-e150-4692-8835-42c1bd5332e0

Promotes the global-header NFS (New Frontend System) plugin from /alpha to stable exports.

Restructures the Global Header plugin so the **New Frontend System (NFS)** becomes the default entry point. **Old Frontend System (OFS)** APIs move to a dedicated `./legacy` export (deprecated). Translations remain at `/alpha`.

Previously, NFS lived under `src/alpha` and was imported via:

```ts
@red-hat-developer-hub/backstage-plugin-global-header/alpha
```

while OFS was the default package export.

This PR flips that relationship:

- NFS is now the default package entry point.
- Legacy/OFS components move to the new `./legacy` subpath (deprecated).
- `/alpha` retains only translation exports; NFS APIs are no longer available from `/alpha`.
- All NFS APIs are promoted from `@alpha` to `@public`.
- Adds dedicated Module Federation entry points so NFS can discover Global Header extensions through `dynamicFrontendFeaturesLoader()`.

---

### Source layout restructuring

Physically moves files to match the new export structure (per review feedback):

| Before | After | Rationale |
|--------|-------|-----------|
| `src/alpha/plugin.ts`, `src/alpha/types.ts` | `src/plugin.ts`, `src/types.ts` | NFS is now the default — code lives at the top level |
| `src/alpha/extensions/` | `src/extensions/` | NFS extensions promoted to top level |
| `src/alpha/defaults/` | `src/defaults/` | NFS default extensions promoted to top level |
| `src/alpha/components/` (NFS) | `src/components/` | NFS components promoted to top level |
| `src/alpha/utils/` | `src/utils/` | NFS utils promoted to top level |
| `src/plugin.ts` (OFS) | `src/legacy/plugin.ts` | OFS plugin moved to legacy |
| `src/components/GlobalHeader.tsx`, `GlobalHeaderComponent.tsx`, `index.ts` (OFS) | `src/legacy/components/` | OFS-only components moved to legacy |
| `src/components/HeaderDropdownComponent/{ApplicationLauncher,Create,Help,Profile}Dropdown*` | `src/legacy/components/HeaderDropdownComponent/` | OFS-only dropdown components moved to legacy |
[...]

PR #4006, opened 2026-07-27
chore(global-header): Global header optimization

debsmita1:

**Resolves:**
https://redhat.atlassian.net/browse/RHIDP-15555

**Solution description:**
- Cut NFS root Module Federation sync from ~972 KB → ~~733 KB (−24.6%) by keeping the package entry thin and moving AppBar + default toolbar UI behind blueprint loader dynamic import()s.
- Added `loaders.ts` with a shared `loadCriticalHeaderBundle()` lazy singleton so every default toolbar loader resolves the same promise. Rspack emits one critical async chunk (~39 KB) for all first-paint header UI instead of one request per widget, keeping MUI off the sync graph.
- Consolidated first-paint UI behind `criticalHeaderBundle.ts` , a barrel that re-exports the canonical component implementations (no duplicated onMount/ tree or second source of truth). Dropdown trigger wrappers (Help, Profile, Starred, App Launcher) live in this critical chunk because they render on mount. Menu contents (GlobalHeaderDropdownContent, StarredDropdownMenu, search result rows, etc.) stay on separate interaction-only import() split points inside each dropdown.
- No header chunk on the sign-in page. `loaders.ts` is statically imported from blueprint registration but does not call import() at module scope. The critical chunk is fetched only when a toolbar extension’s loader runs after the app shell mounts.
- Kept building blocks (GlobalHeaderMenuItem, GlobalHeaderDropdown, HeaderIconButton, CompanyLogo, etc.) on `./components` so third-party plugins pull MUI-heavy UI only from blueprint-loader async chunks, not the root sync graph. `useBrandingFullLogo` is internal; branding goes through CompanyLogo.
- Used blueprint loaders for default extensions; registered outlined system icons via IconBundleBlueprint instead of Material Icons ligatures on the sync path.
- lazy menu items settle before showing empty state (trackValidity uses an off-menu probe so <MenuList> keeps direct MenuItem children for keyboard nav). `useRetainMenuContent` prevents empty-menu flash during MUI close transitions. `DropdownMenuLoading` shows a horizontal Progress bar while menu chunks load on slow networks.
- Remaining sync is mostly MF shared framework/vendor (Zod, @emotion, @mui/material/styles, frontend-plugin-api) plus blueprint registration, not the full AppBar bundle.
- Updated NFS integration guide, icon docs, README, and changeset with the new loading model and migration path
 
**Expose summary**

Metric | BEFORE | AFTER | Δ
-- | -- | -- | --
Sync chunks | 5 | 5 | 0
Sync size | 972 KB | 733 KB | -239 KB (-24.6%)
Async chunks | 198 | 216 | +18 (+9.1%)
Async size | 963 KB | 1269 KB | +306 KB (+31.8%)

**GIF:**

Tested on RHDH with Quickstarts plugin

https://github.com/user-attachments/assets/aa12ad08-a1d8-415a-8f2b-a82b73a0fcbe

**Test setup:**

Update the Quickstarts plugin

1. 
```
import { GlobalHeaderMenuItem } from '@red-hat-developer-hub/backstage-plugin-global-header/components';
```

2. Use blueprint loader — src/index.tsx:

```
const quickstartHelpMenuItem = GlobalHeaderMenuItemBlueprint.make({
  name: 'quickstart',
  params: {
    target: 'help',
    priority: 50,
    loader: async () => {
      const { QuickstartHelpMenuItem } = await import(
        './QuickstartHelpMenuItem'
      );
      return QuickstartHelpMenuItem;
    },
  },
});
(GlobalHeaderMenuItemBlueprint still comes from /alpha.)
```

3. Local portal (so TS + runtime see /components before publish)

In workspaces/quickstart:

File	                                                       Dependency
package.json (resolutions)             portal:../global-header/plugins/global-header
plugins/quickstart/package.json.   portal:../../../global-header/plugins/global-header
packages/app/package.json.          portal:../../../global-header/plugins/global-header

4. Build and export global-header and then quick start plugins into RHDH
```
npx @red-hat-developer-hub/cli plugin export --clean --dev \
  --dynamic-plugins-root <path-to>/rhdh/dynamic-plugins-root
```

5. check-out PR https://github.com/redhat-developer/rhdh/pull/5115 and run the following commands:

```
EXPERIMENTAL_MODULE_FEDERATION=true yarn workspace app-next build
yarn workspace backend start:next 
```

     

workspaces/global-header/plugins/global-header/src/defaults/menuItemExtensions.tsx
```diff
@@ -78,8 +78,11 @@ export const supportButtonMenuItemExtension =
     name: 'support-button',
     params: {
       target: 'help',
-      component: SupportButton,
       priority: 10,
+      loader: () =>
+        import('../../components/SupportButton/SupportButton').then(
+          m => m.SupportButton,
+        ),
```
ciiay, line 85: `HelpDropdown` sets `trackValidity`. This PR makes `support-button` async via `loader`. `GlobalHeaderDropdown` flips to `emptyState` (and unmounts menu content) when no `[role="menuitem"]` is found, and never resets `hasVisibleItems`.
This could cause a race condition: open Help before the lazy SupportButton paints → no menuitem → `hasVisibleItems=false` → content unmounted → items never mount for that session (and reopen stays empty). Qodo also flagged this which I agree is high severity.

ciiay (changes requested):
The following files still importing GlobalHeaderMenuItem from `/alpha`. Let's also update them in this pr.

1. workspaces/quickstart/plugins/quickstart/src/QuickstartHelpMenuItem.tsx
2. workspaces/app-defaults/packages/app/src/modules/drawer-demo/DrawerDemoContent.tsx

debsmita1:
> The following files still importing GlobalHeaderMenuItem from `/alpha`. Let's also update them in this pr.
> 
> 1. workspaces/quickstart/plugins/quickstart/src/QuickstartHelpMenuItem.tsx
> 2. workspaces/app-defaults/packages/app/src/modules/drawer-demo/DrawerDemoContent.tsx

These are different workspaces importing from the global-header package . I will update them once the new version of the global-header with these changes are released

debsmita1:
Let's merge this one https://github.com/redhat-developer/rhdh-plugins/pull/3413 first

workspaces/global-header/.changeset/components-entry-lazy-appbar.md
```diff
@@ -0,0 +1,9 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-header': minor
+---
+
+**BREAKING**: Import building blocks (`GlobalHeaderIconButton`, `GlobalHeaderMenuItem`, `GlobalHeaderDropdown`) only from `@red-hat-developer-hub/backstage-plugin-global-header/components` — they are no longer re-exported from the root entry or deprecated `/alpha`. Prefer dynamic `import()` inside blueprint loaders.
```
ciiay: This is labeled **BREAKING** (building blocks leave the root and `/alpha`) but the bump is `minor` on an already-stable `2.0.0`.

`workspaces/quickstart` depends on `^2.0.0` and still does:

```ts
import { GlobalHeaderMenuItem } from '@red-hat-developer-hub/backstage-plugin-global-header';
```

A 2.1.0 publish will be picked up automatically and that import will fail. Same for app-defaults importing from `/alpha`.

“Update consumers after release” only works with a **major** (`3.0.0`). With `minor`, please either:

- change this changeset to `major`, or
- keep a deprecated root re-export for one release, or
- update quickstart / app-defaults in this PR (or a same-train follow-up that lands before publish).

workspaces/global-header/packages/app/src/App.tsx
```diff
@@ -57,24 +56,35 @@ const signInModule = createFrontendModule({
...
+          return function CustomHelpMenuItem({
+            handleClose,
+          }: {
+            handleClose?: () => void;
+          }) {
+            return (
+              <GlobalHeaderMenuItem
+                to="https://backstage.io/docs"
+                title="Backstage Docs"
+                icon="menu_book"
```
ciiay: Two nits in the sample loader:

1. Comment still says this stays off the `/alpha` NFS sync chunk — NFS is on the **root** now.
2. `icon="menu_book"` is a Material Icons ligature id. After this PR `HeaderIcon` returns `null` for unregistered ids, so this example will render without an icon. Use a registered system-icon id (see `globalHeaderSystemIcons`), inline SVG, or a URL.

workspaces/global-header/docs/new-frontend-system.md
```diff
@@ -160,26 +174,38 @@ export const docsItem = GlobalHeaderMenuItemBlueprint.make({
...
+      return function MyDocsLink({
+        handleClose,
+      }: {
+        handleClose?: () => void;
+      }) {
+        return (
+          <GlobalHeaderMenuItem
+            to="https://docs.example.com"
+            title="Documentation"
+            icon="menu_book"
```
ciiay: Same ligature issue as the sample app: `menu_book` is not in `globalHeaderSystemIcons`, so this documented example will render no icon. Please switch to a registered id from the system-icons table (or SVG/URL) so the docs match the new resolution rules.

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
```diff
@@ -82,14 +76,5 @@ export const HeaderIcon = ({
...
-  return (
-    <MuiIcon
-      fontSize={size}
-      baseClassName="material-icons-outlined"
-      sx={layout}
-    >
-      {ligature}
-    </MuiIcon>
-  );
+  return null;
```
ciiay, line 79: Dropping the ligature fallback is the right call for Module Federation (no host icon font), but it is a silent behavior change for existing config.

Ids that used to work via the font — `menu_book`, `waving_hand`, `hub`, `manage_accounts`, `addCircleOutline`, `account_circle` — now render nothing unless the host or `IconBundleBlueprint` registers them.

Default extensions in this PR were updated; sample app, NFS docs, and in-repo consumers (quickstart `waving_hand`) were not. Please align those, and call out the config migration in the changeset so deployers know `globalHeader.menuItems[].icon` / `globalHeader.components[].icon` values may need to change.

workspaces/global-header/plugins/global-header/src/extensions/globalHeaderModule.tsx
```diff
@@ -70,7 +80,9 @@ function GlobalHeaderWrapper({
   );
   return (
     <GlobalHeaderProvider components={allComponents} menuItems={allMenuItems}>
-      <GlobalHeader />
+      <Suspense fallback={null}>
+        <LazyGlobalHeader />
+      </Suspense>
```
ciiay, line 88: Non-blocking: a rejected dynamic import throws past `Suspense`, and the `ErrorBoundary` inside `GlobalHeader` only runs after the shell loads. Wrapping this in `@backstage/core-components` `ErrorBoundary` (even with a null fallback) would keep a chunk-load failure from taking out the wrapper subtree.

`fallback={null}` also means a header-sized layout shift on first paint — worth a tiny placeholder if that’s noticeable in RHDH.

workspaces/global-header/plugins/global-header/src/icons/globalHeaderSystemIcons.ts
```diff
@@ -0,0 +1,48 @@
...
+  add: AddCircleOutline,
+  article: ArticleOutlined,
+  bug_report: BugReportOutlined,
+  dashboard: DashboardOutlined,
+  developerHub: HubOutlined,
+  forum: ForumOutlined,
+  logout: LogoutOutlined,
+  manageAccounts: ManageAccountsOutlined,
+  quiz: QuizOutlined,
+  support: Support,
```
ciiay: Non-blocking / measurement ask: this file is imported from `globalHeaderModule`, which is exported from the **root** entry, so these `@mui/icons-material` modules land back on the NFS sync chunk this PR is trying to empty.

Twelve outlined icons are much smaller than AppBar, so the 88% win can still be real — but the PR/changeset text says MUI stays off the root sync path, which is no longer strictly true. Worth confirming in the expose summary, or lazy-registering if `IconBundleBlueprint` allows it.

Also a small consistency nit: ids mix snake_case (`bug_report`) and camelCase (`manageAccounts`, `developerHub`), and `support` uses `Support` rather than an outlined variant.

ciiay (changes requested):
Thanks for the NFS sync work — the loader / `/components` split and the `trackValidity` settle logic look right, and the new dropdown tests cover the empty-state race.

The architecture matches RHIDP-15555 and the ~88% sync-size drop is a real win. A few API/consumer issues should be fixed before merge:

1. **Breaking export move vs `minor` bump.** Building blocks are removed from the root and `/alpha` on a 2.x package, but the changeset is `minor`. `workspaces/quickstart` is on `^2.0.0` and still imports `GlobalHeaderMenuItem` from the root, so 2.1.0 would break it (same for app-defaults `/alpha`). Please use a **major**, keep a deprecated root re-export for one release, or update in-repo consumers in this PR.
2. **`HeaderIcon` now returns null for unregistered ids.** Sample app + NFS docs still use `menu_book`; quickstart uses `waving_hand`. Those icons will disappear unless registered or switched to a system-icon id / SVG / URL.
3. Leftover “`/alpha` sync chunk” comment in `packages/app/src/App.tsx`.

Non-blocking: `globalHeaderSystemIcons` static MUI imports land back on the root module; consider wrapping `LazyGlobalHeader` in an `ErrorBoundary`.

CI looks good. Happy to re-review after the versioning + icon-id follow-ups.

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
debsmita1: I have addressed this in the changeset as a breaking change

workspaces/global-header/plugins/global-header/src/icons/globalHeaderSystemIcons.ts
debsmita1: Yes, the PR changes doesnot eliminating all MUI from root and the changeset reflects that. Lazy icon registration is not possible for IconBundleBlueprint 

On naming ids, I have fixed it in the code and captured it in the doc and the changeset

support now uses SupportAgentOutlined

workspaces/global-header/docs/new-frontend-system.md
```diff
@@ -160,26 +174,38 @@ export const docsItem = GlobalHeaderMenuItemBlueprint.make({
...
+          <GlobalHeaderMenuItem
+            to="https://docs.example.com"
+            title="Documentation"
+            icon="article"
+            onClick={handleClose}
+          />
+        );
+      };
+    },
+  },
```
rohitkrai03: This looks like a super complex component example for docs. Would be better to break it down into simple chunks and compose properly.

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
rohitkrai03: @debsmita1 I am thinking if we should have a default icon instead. Returning null just renders that awful text like incomplete icon which we already have as a bug. It just looks like the UI is broken because of that. Maybe we can have a default icon with a tooltip that gets rendered when none of the above options are available, WDYT?

workspaces/global-header/plugins/global-header/src/utils/readConfigComponents.tsx
```diff
@@ -42,13 +50,15 @@ export function readConfigComponents(
...
 
     const ConfigComponent = () => (
-      <HeaderIconButton
-        title={title}
-        titleKey={titleKey}
-        icon={icon}
-        tooltip={tooltip}
-        to={link}
-      />
+      <Suspense fallback={null}>
```
rohitkrai03, line 54: Should we have some fallback component here? Maybe a loader or something?

workspaces/global-header/plugins/global-header/src/componentsExport.ts
```diff
@@ -0,0 +1,51 @@
...
+export type { HeaderIconButtonProps } from './components/HeaderIconButton/HeaderIconButton';
+
+export { GlobalHeaderMenuItem } from './components/GlobalHeaderMenuItem';
+export type { GlobalHeaderMenuItemProps } from './components/GlobalHeaderMenuItem';
+
+export { GlobalHeaderDropdown } from './components/GlobalHeaderDropdown';
+export type { GlobalHeaderDropdownProps } from './components/GlobalHeaderDropdown';
+
+/**
+ * Empty module so this package export is published as a Module Federation
```
rohitkrai03: Do we actually need this to be available at module federation level? These are all the components that will be imported by another plugin to build there own components which in turn should be lazy loaded. But the base components could just be exported at `/component` without a remote in module federation.

workspaces/global-header/plugins/global-header/package.json
```diff
@@ -123,6 +127,7 @@
       "PluginRoot": "./src/index.ts",
       "Alpha": "./src/alpha/index.ts",
       "Legacy": "./src/legacy.ts",
+      "Components": "./src/componentsExport.ts",
```
rohitkrai03: Why do we need these in scalprum? This should only have modules that are needed in OFS.

rohitkrai03 (changes requested):

workspaces/global-header/plugins/global-header/src/components/GlobalHeaderDropdown.tsx
```diff
@@ -100,14 +152,26 @@ export const GlobalHeaderDropdown = ({
...
-          handleClose={handleClose}
-        />
+        <>
+          {/*
+           * Keep the menu content mounted while showing the empty state so
+           * lazy ExtensionBoundary items can still render and recover.
+           * This prevents permanently latching into the empty state if a
+           * menu item appears after the initial validity check.
+           */}
+          <div hidden={trackValidity && showEmptyState}>
```
rohitkrai03: Why are we wrapping the content in a div here? This puts a div directly inside the MUI `<ul role="menu">` which isn't valid, and it also breaks keyboard navigation since MUI only wires up arrow keys/type-ahead for direct children of the list. Can we keep the menu items as direct children instead? Also would be good to add a keyboard-nav test, the current tests won't catch this.

workspaces/global-header/plugins/global-header/src/extensions/blueprints.tsx
```diff
@@ -92,43 +101,111 @@ export interface MenuItemParams {
 // Data-driven component factories
 // ---------------------------------------------------------------------------
 
-function createDataDrivenToolbarComponent(
+function resolveLazyComponent(
+  node: AppNode,
+  loader: () => Promise<ComponentType<any>>,
+): ComponentType<any> {
+  return ExtensionBoundary.lazyComponent(node, async () => {
```
rohitkrai03, line 113: Do we have a Suspense boundary closer to these lazy items? Right now the nearest fallback is the top-level `<Suspense fallback={null}>` around LazyGlobalHeader, so if lazyComponent doesn't bring its own Suspense, a suspending menu item would blank out the whole header while it loads. The tests only use sync components so we aren't really covering this path. Can we confirm lazyComponent wraps its own Suspense, or should we add one closer to the item?

workspaces/global-header/plugins/global-header/src/defaults/toolbarExtensions.tsx
```diff
@@ -17,98 +17,127 @@
 /**
  * Default toolbar component extensions (`gh-component`) for the global header.
  *
+ * Heavy UI uses blueprint `loader` (ExtensionBoundary.lazyComponent).
```
rohitkrai03: The async size went up by ~984 KB across 43 chunks and all these loaders fire when the header mounts. I know cutting the sync chunk is the goal, but did we check the actual first-paint/render time of the header? Are we now waterfalling a bunch of small chunks on mount? Just want to make sure that's acceptable, WDYT?

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
debsmita1: Updated. Unknown icon ids now render HelpOutlineOutlined with an unknown id warning in the browser's console. Unknown id tooltip will otherwise overlap with the menu item tooltip
Screenshot 2026-08-25 at 9 01 58 PM

workspaces/global-header/plugins/global-header/src/utils/readConfigComponents.tsx
debsmita1: This is intentional.The toolbar buttons are tiny chunks, a spinner would flash in the header and cause a layout shift. The button appears as soon as the chunk resolves.

workspaces/global-header/plugins/global-header/src/componentsExport.ts
debsmita1: This is required for quickstart’s help menu item . Without a federated `./components` expose, that import would either fail at runtime or get resolved through the root remote, defeating the sync-chunk split.

workspaces/global-header/plugins/global-header/src/defaults/toolbarExtensions.tsx
debsmita1: The goal is moving MUI/AppBar UI off the root MF sync chunk. The async growth is expected as code splits into per-component chunks

workspaces/global-header/plugins/global-header/src/extensions/blueprints.tsx
debsmita1: ExtensionBoundary.lazyComponent self-wraps each resolved component in ExtensionBoundary + Suspense with the default <Progress /> fallback. A suspending toolbar/menu item shows progress in that slot it does not blank the whole header via the app-root Suspense fallback={null}. Added a comment in resolveLazyComponent pointing to this.

https://github.com/backstage/backstage/blob/1a8a90c3675c1aab02abb272eebc819e630e2b09/packages/frontend-plugin-api/src/components/ExtensionBoundary.tsx#L152

workspaces/global-header/plugins/global-header/src/components/GlobalHeaderDropdown.tsx
debsmita1: Fixed. The wrapper now uses `display: contents` when items are visible so MenuItems stay direct children of the MenuList for keyboard/type-ahead. It only switches to `display: none` while the empty state is shown . Also added an arrow-key navigation test

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
rohitkrai03: Not sure if thats the best option for default icon since we also use the same for help dropdown.

workspaces/global-header/plugins/global-header/src/componentsExport.ts
rohitkrai03: Wouldn't QuickStart import this component at build time? The component then built based on that will be lazy loaded in the blueprint registration, right?

workspaces/global-header/plugins/global-header/src/defaults/toolbarExtensions.tsx
rohitkrai03: Understood on shrinking the MF sync chunk but this is more about the overall approach than this PR.

If first paint stays the same because one large chunk is now many smaller ones all loading on mount, we have not really gained anything. Multiple HTTP requests can be slower than one, especially when everything is needed immediately — request overhead, weaker compression, and waterfalls add up even with HTTP/2.

Async growth from splitting is expected, but it only pays off if those chunks are deferred or first paint is actually faster. Have we measured header first-paint before vs after?

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
debsmita1: @ShiranHi could you help here with a fallback icon for global header menu items when the configured icon is not found

ShiranHi: What do you think about using a generic [shapes icon](https://fonts.google.com/icons?icon.query=shapes&utm_source=chatgpt.com) outlined as the default?

debsmita1: Done

Screenshot 2026-09-03 at 12 23 24 PM

cc @rohitkrai03 @ShiranHi

workspaces/global-header/plugins/global-header/src/componentsExport.ts
debsmita1: removed the empty module export

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
ShiranHi: Thanks Debsmita! Are all these icons set to the exact same size? The square and question mark icons feel slightly bolder and larger than the rest

workspaces/global-header/plugins/global-header/src/defaults/toolbarExtensions.tsx
debsmita1: @rohitkrai03  I have not run a formal browser first-paint benchmark before vs after. What I have measured is the MF graph via package bundle + mf-manifest.json

I have updated the approach to address many small chunks all loading on mount. The root expose stays a thin registration surface (blueprints, plugin, defaults metadata). All on-mount header UI goes through one shared async module, **onMountHeaderBundle**, so the AppBar shell and every default toolbar loader hit the same import(), producing one ~148 KB async chunk, not one request per widget. Now on-mount header widgets (AppBar, logo, search input, spacers, icon buttons, dropdown _triggers_) share a single async chunk via `src/components/onMountHeaderBundle.ts`. Dropdown menus, starred lists, and search-result rows stay on separate `import()` split points so they are not required for first paint.

Note: switching to the newer API with a top-level zod import as the `config.schema` option for `createExtensionBlueprint` is now deprecated in favor of a new top-level `configSchema` option, which increased the sync chunk again

Updated the PR description with the new expose summary

workspaces/global-header/plugins/global-header/src/components/HeaderIcon/HeaderIcon.tsx
debsmita1: @ShiranHi  Fixed it 

Screenshot 2026-09-03 at 9 59 05 PM

ciiay (approved):
LGTM — thanks for the follow-ups.

The items from my earlier review are addressed:

- Changeset is now **major**, so `^2.0.0` consumers (quickstart, app-defaults) will not pick up the breaking `/components` move automatically.
- `HeaderIcon` uses the shapes outlined fallback + console warning, and the changeset documents the ligature → system-icon migration.
- Sample app comment no longer mentions `/alpha`; the example icon is a registered id (`article`).
- `LazyGlobalHeader` is wrapped in `ErrorBoundary`.
- Icon ids are camelCase; `support` uses `SupportAgentOutlined`.
- On-mount UI sharing `onMountHeaderBundle` answers the “many small chunks on first paint” concern without putting AppBar back on the sync graph.

Follow-ups (not blocking this PR):

- Quickstart / app-defaults still import `GlobalHeaderMenuItem` from root / `/alpha` — please land those on 3.x in a same-train follow-up before RHDH consumes the new major.
- The PR description still has the old 117 KB / −88% table above the updated 757 KB / −22% numbers; worth aligning so reviewers don’t mix the two.

CI is green. Approving.

ciiay (approved):
/lgtm

Thanks for the updates 👏

workspaces/global-header/plugins/global-header/src/components/onMountHeaderBundle.ts
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * Do not statically import this file from the package root (`index.ts`,
+ * `plugin.ts`, `blueprints.tsx`) or it lands on the Module Federation sync
+ * graph. Dropdown menus, search results, and highlighter langs stay on their
+ * own `import()` / `React.lazy` split points.
+ *
+ * @internal
+ */
+
+export { GlobalHeader } from './GlobalHeader';
```
rohitkrai03: I don't understand this approach. What is the difference between this bundle file and a basic index.ts bucket export file?

rohitkrai03 (changes requested):

workspaces/global-header/plugins/global-header/src/components/onMountHeaderBundle.ts
debsmita1: `onMountHeaderBundle.ts` is a the single import() target in `loaders.ts`, and every critical loader (loadGlobalHeader, loadCompanyLogo, loadSearchComponent, etc.) shares that same promise. That way webpack emits one async chunk (~148 KB) for all critical header UI, instead of one chunk per toolbar extension as in multiple network requests on mount.


I have made the following updates to the approach:

- Added loaders.ts with a shared lazy singleton (**loadHeaderBundle()**), so that all critical header UI loads from one async chunk
- Split critical first-paint components into onMount/ (GlobalHeader, CompanyLogo, SearchComponent, Spacer, HeaderIconButton, etc.)
- Slimmed onMountHeaderBundle.ts to a webpack chunk boundary that re-exports only onMount/ (dropdowns removed)
- Moved dropdown loaders (loadStarredDropdown, loadHelpDropdown, etc.) to separate import() calls for interaction-only chunks
- Updated toolbarExtensions.tsx, blueprints.tsx, readConfigComponents.tsx, and globalHeaderModule.tsx to use the new loaders
- Deferred header bundle preload to useEffect in GlobalHeaderWrapper so it doesn’t fetch on the sign-in page

As a result, We now have smaller sync bundle, one network request for the visible header, interaction UI loaded only when needed, and no header chunk on the sign-in page

rohitkrai03: I understand the need for one shared `import()` target, but the dynamic import itself creates that boundary. We can keep one canonical component tree, export the first-paint components from a single `headerBundle.ts` barrel, and cache `import('./headerBundle')` in `loaders.ts`. The current `onMountHeaderBundle.ts` → `onMount/index.ts` → duplicated component tree adds two layers and copies roughly 25 implementation/test files without changing the loading model. It is already causing the Sonar duplication failure and creates two sources of truth. If Rspack genuinely requires physical copies rather than re-exporting the canonical implementations, could we include the bundle/manifest comparison demonstrating that constraint?

workspaces/global-header/plugins/global-header/src/defaults/toolbarExtensions.tsx
rohitkrai03: I think the waterfall is still doing more work than the description suggests.

`GlobalHeader` renders every contributed toolbar component as soon as the shared header bundle resolves. That means these loaders all run during the initial mount:

- `loadStarredDropdown`
- `loadApplicationLauncherDropdown`
- `loadHelpDropdown`
- `loadProfileDropdown`

The menu contents are interaction-deferred, but the dropdown wrappers themselves are still loaded immediately. In practice, that creates four additional mount-time requests, so the current implementation does not appear to support the claim that the visible header is delivered in a single request.

Could we include the dropdown trigger wrappers in `criticalHeaderBundle` and keep only their menu contents behind the interaction loaders? That would make the loading model clearer: one request for the visible header, followed by additional requests only when a menu is opened.

If the current waterfall is expected to behave differently, could we add a browser waterfall showing that these imports are not fetched during the initial mount?

workspaces/global-header/plugins/global-header/src/components/GlobalHeaderDropdown.tsx
rohitkrai03: I don't think `display: contents` fixes the original issue. It changes layout, but not the DOM tree—the `<div>` is still the direct child of the `<ul>`, with the menu items nested inside it.

MUI's `MenuList` navigates using `firstChild` / `lastChild` and sibling elements. That means Home/End and wrapping past the first or last item can focus the wrapper instead of a menu item.

The new test only covers ArrowDown between two items inside the wrapper, so it misses those cases. Can we remove the wrapper and add tests for boundary navigation plus Home/End?

workspaces/global-header/plugins/global-header/src/extensions/globalHeaderModule.tsx
```diff
@@ -68,9 +78,20 @@ function GlobalHeaderWrapper({
       ),
     [extensionMenuItems, configMenuItems],
   );
+
+  // Start the shared critical header chunk only after the app shell mounts
+  // (post sign-in), not during plugin module evaluation on the sign-in page.
+  useEffect(() => {
```
rohitkrai03, line 84: I don't think this effect is doing anything for the initial load.

`<LazyGlobalHeader />` calls `loadGlobalHeader()` during render, before effects run, and both paths use the same cached promise. If this wrapper renders on the sign-in page, the lazy component will still start the request; if it doesn't render there, this preload isn't needed.

Can we remove the effect? If auth gating is required, I think we should gate the lazy render explicitly instead.

workspaces/global-header/plugins/global-header/src/componentsExport.ts
```diff
@@ -0,0 +1,46 @@
...
+ * as an npm dependency when they are exported; they do not fetch a
+ * global-header `/components` remote at runtime. There is no default
+ * `FrontendModule` here so `./components` is not listed in `backstage.features`
+ * and is not downloaded at app startup.
+ *
+ * @public
+ * @packageDocumentation
+ */
+
+import './configureMuiClassName';
```
rohitkrai03, line 37: I think this can break in an optimized build.

This entry relies on `configureMuiClassName.ts` running for its side effect, but the package declares `"sideEffects": false`. That allows Webpack/Rspack to drop this import, which would leave the MUI class-name prefix unconfigured in `/components` (and potentially the lazy/legacy entries as well).

Could we either allow-list `configureMuiClassName.ts` in `sideEffects`, or call an explicit initializer from a module the bundler must retain? A production-build check would be useful here.

workspaces/global-header/.changeset/components-entry-lazy-appbar.md
```diff
@@ -0,0 +1,13 @@
+---
+'@red-hat-developer-hub/backstage-plugin-global-header': major
+---
+
+**BREAKING**: Import building blocks (`GlobalHeaderIconButton`, `GlobalHeaderMenuItem`, `GlobalHeaderDropdown`) only from `@red-hat-developer-hub/backstage-plugin-global-header/components` — they are no longer re-exported from the root entry or deprecated `/alpha` (which now exports translations only). Prefer dynamic `import()` inside blueprint loaders.
```
rohitkrai03, line 5: Looks like the changeset is missing a few other public API removals.

`report.api.md` also removes `useBrandingFullLogo`, `LogoURLs`, and `CompanyLogoProps` from the root entry, but the migration notes only cover the three building blocks and `HeaderIcon`. Existing consumers of those exports will get compile errors without knowing the replacement.

Should these move to `/components` too, or should we document their removal and the new import path here?

rohitkrai03 (changes requested):
I think we can keep the single async chunk without introducing a second component tree.

`import()` already provides the split point, so the structure could be simplified to:

```text
src/components/
  criticalHeaderBundle.ts   # barrel for first-paint components
  loaders.ts                # cached dynamic import
  AppBar/
  Logo/
  Search/
  ...
```

`criticalHeaderBundle.ts` would explicitly re-export the components needed to render the initial header, while `loaders.ts` would own the cached import:

```ts
let criticalHeaderBundlePromise: Promise<
  typeof import('./criticalHeaderBundle')
> | undefined;

export const loadCriticalHeaderBundle = () =>
  (criticalHeaderBundlePromise ??= import('./criticalHeaderBundle'));
```

The components would remain in one canonical tree, making it clear which components belong to the critical bundle. The dropdown trigger wrappers can be included in that bundle if they are rendered during the initial header mount, while their menu contents keep their own interaction-based lazy loaders.

With this structure, we could remove `onMountHeaderBundle.ts`, `onMount/index.ts`, and the duplicated `onMount/` tree. The current indirection adds two layers and copies roughly 25 implementation and test files without changing the loading model. It also creates two sources of truth and is already contributing to the Sonar duplication failure.

`criticalHeaderBundle` also describes the purpose more clearly than `onMount`: it explains what belongs in the bundle, rather than only when it is loaded.

If Rspack genuinely requires physical copies instead of a barrel that re-exports the canonical implementations, could we include the bundle or manifest comparison that demonstrates that constraint?

workspaces/global-header/plugins/global-header/src/defaults/menuItemExtensions.tsx
```diff
@@ -63,23 +60,42 @@ export const logoutMenuItemExtension = GlobalHeaderMenuItemBlueprint.make({
...
+      ),
   },
 });
 
 // ---------------------------------------------------------------------------
 // Help dropdown items
 // ---------------------------------------------------------------------------
 
+/** @public */
+export const quickStartMenuItemExtension = GlobalHeaderMenuItemBlueprint.make({
```
rohitkrai03: Is there a reason to contribute this extension from `global-header` when the Quickstart plugin is already contributing it?

debsmita1: This was added 'cause `defaultHelpDropdownMountPoints` always included a Quickstart docs link in the legacy app; the NFS defaults only had Support, so the Help E2E failed. 

I remembered that I can wire it in the app-config extensions. So, I have updated the config YAML and dropped the QS from the extensions

workspaces/global-header/plugins/global-header/src/components/loaders.ts
```diff
@@ -0,0 +1,68 @@
...
+ * Lazy singleton — do not use a top-level `import()` here. This module is
+ * statically imported from blueprint registration (`blueprints.tsx`,
+ * `toolbarExtensions.tsx`), which runs before sign-in. A module-scope
+ * `import()` would fetch the header chunk on the sign-in page.
+ */
+let criticalHeaderBundlePromise: Promise<CriticalHeaderBundle> | undefined;
+
+export const loadCriticalHeaderBundle = (): Promise<CriticalHeaderBundle> => {
+  if (!criticalHeaderBundlePromise) {
+    criticalHeaderBundlePromise = import('./criticalHeaderBundle');
```
lokanandaprabhu:
```suggestion
   criticalHeaderBundlePromise ??= import('./criticalHeaderBundle').catch(error => {
  criticalHeaderBundlePromise = undefined;
  throw error;
});
```

If import('./criticalHeaderBundle') rejects, the rejected promise is cached permanently. Retries won’t happen until full page reload. 

Can we clear criticalHeaderBundlePromise on rejection

workspaces/global-header/plugins/global-header/src/components/GlobalHeaderDropdown.tsx
```diff
@@ -73,42 +86,138 @@ export const GlobalHeaderDropdown = ({
...
+    <>
+      <HeaderDropdownComponent
+        buttonContent={buttonContent}
+        buttonProps={buttonProps}
+        isIconButton={isIconButton}
+        tooltip={tooltip}
+        onOpen={handleOpen}
+        onClose={handleClose}
+        anchorEl={anchorEl}
+        menuListRef={trackValidity ? menuListRef : undefined}
```
lokanandaprabhu: `menuListRef` is created and passed to `HeaderDropdownComponent` when `trackValidity` is on, but validity logic uses `validityProbeRef` on an external div. `menuListRef` is never read for validity anymore — leftover from the earlier approach.

lokanandaprabhu (approved):

rohitkrai03 (approved):
/lgtm

PR #4021, opened 2026-07-28
fix(quickstart): make Quick start help menu item clickable

ciiay:
## Description
The Quick start item in the global-header help dropdown was visible but not clickable, so the drawer never opened. This change renders a real MUI `MenuItem` via `forwardRef` (instead of `GlobalHeaderMenuItem` without a `to` prop, which used `component={Fragment}` in global-header 1.21.0 and dropped click handlers), bumps the global-header dependency, and adds unit coverage.

### Root cause
`GlobalHeaderMenuItem` without a `to` prop rendered `MenuItem component={Fragment}` in `@red-hat-developer-hub/backstage-plugin-global-header@1.21.0`, which discarded `onClick` and `role="menuitem"`.

## Fixed
- [RHDHBUGS-3489](https://redhat.atlassian.net/browse/RHDHBUGS-3489) — Quick start menu item in global-header help dropdown is not clickable

## UI before changes
Before fix

## UI after changes
After fix

## Test Plan
- [ ] Start the quickstart NFS app (`yarn start` in `workspaces/quickstart`)
- [ ] Open the Help (?) dropdown in the global header
- [ ] Verify \"Quick start\" appears as a menu item
- [ ] Click \"Quick start\" and verify the quickstart drawer opens on the right
- [ ] Verify no console warning about `ref` on `React.Fragment`
- [ ] Verify the sidebar Quick start control still toggles the drawer

## Checklist

### DIFF
--- a/workspaces/quickstart/.changeset/clickable-menu-refix.md
+++ b/workspaces/quickstart/.changeset/clickable-menu-refix.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-quickstart': patch
+---
+
+Fix Quick start help menu item not responding to clicks in the global-header dropdown
--- a/workspaces/quickstart/.changeset/clickable-menu-refix.md
+++ b/workspaces/quickstart/.changeset/clickable-menu-refix.md
@@ -2,4 +2,4 @@
 '@red-hat-developer-hub/backstage-plugin-quickstart': patch
 ---
 
-Fix Quick start help menu item not responding to clicks in the global-header dropdown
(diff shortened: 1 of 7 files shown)


HusneShabbir (commented):
/lgtm 

Tested Locally, working as expected

rohitkrai03 (changes requested):
@ciiay shouldnt the fix be in global header component if it's not using on click handler properly?

debsmita1 (approved):
Tested this on RHDH locally in app-next

Before the changes

Image

With the changes

Image

/lgtm

rohitkrai03 (approved):
/approve

PR #4036, opened 2026-07-28
chore(RHIDP-14706): update archive script in rhdh-plugins

AdamBrez:
Jira ticket: https://redhat.atlassian.net/browse/RHIDP-14706
References of the archived workspace are now automatically removed from the files: codecov.yml, .github/labeler.yml and .github/pr-labeler.yml


04kash (commented):
Nice! Just address the SonarCloud warning to "Use `.indexOf()` instead of `.findIndex()` when looking for the index of an item." and this should be good to merge.

04kash (approved):
Thanks @AdamBrez!

04kash (changes requested):
Actually could you also give the bug reported by qodo a look and see if it is something that actually needs fixing?

AdamBrez:
Qodo was right. The previously used argument in the indexOf() would silently fail the whole function. All should be good now.

04kash (approved):
Nice! Merging this in

PR #3849, opened 2026-07-19
feat(intelligent-assistant): Lightspeed Notebook enhancements

its-mitesh-kumar:
## Description

UI/UX improvements for the Lightspeed Notebook, aligned to the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/intelligent-assistant/notebooks) and acceptance criteria in [RHIDP-14896](https://redhat.atlassian.net/browse/RHIDP-14896).

## Fixed 
- https://redhat.atlassian.net/browse/RHIDP-14896
## UI after changes

https://github.com/user-attachments/assets/a2386b0c-b3d2-4386-a536-a2eb25e61f80

### Changes

**Notebook card list**
- Humanized document count: "No resources" (zero), "1 document" (singular), "N resources" (plural)
- Removed extra left padding on card body text for proper alignment

**Notebook action menu**
- Added PenIcon to Rename and TrashIcon to Delete dropdown items

**Add button icon**
- Changed sidebar "Add" button icon from PlusCircleIcon (filled) to AddCircleOIcon (outline)
- Moved "Upload a resource" button icon to the left side

**Upload modal**
- Title changed to "Add resources"
- Drag-and-drop text updated to "Drag and drop files here, or click to browse"
- Replaced plain-text file types with styled chips derived from NOTEBOOK_EXTENSION_TO_FILE_TYPE
- Added "Supported formats:" label and "Maximum file size is 25 MB." text
- Upload icon styled with brand/blue color and increased size
- "Add" button shows count only when files are selected (no more "Add (0)")
- Entire dropzone area is clickable to browse files (custom DropzoneClickArea component)

**Upload modal X button**
- Removed red hover color from file remove button
- Replaced MUI CloseIcon with PF TimesIcon (12x12px)
- Changed file item container from solid background to border-only style

**Disabled dropzone at file limit**
- Dropzone is now visually disabled (opacity + pointer-events) instead of hidden when 10 files are reached
- Tooltip on hover shows the max-reached message

**Prevent stacked modals**
- Upload modal closes before overwrite modal opens (no more modal stacking)
- Redesigned overwrite modal: "File already exists" title, warning alert with counts, "Replace existing files" / "Ignore duplicated files" radio options, file list with warning icons on duplicates, "Upload (N)" / "Back" buttons
- "Back" returns to upload modal with previously selected files preserved

**Auto-save on close**
- Empty untitled notebooks are automatically deleted on close
[...]

PR #4093, opened 2026-07-30
feat(intelligent-assistant): implement docked and overlay display modes for Notebook

rohitratannagar:
## Summary

Bring full notebook functionality to the **docked** and **overlay** display modes (previously fullscreen-only), and make notebook streaming survive switching between modes. Also fixes the source-citation chip so it appears live during a response instead of only after a page refresh.

- Enable notebooks (tabs, notebook management, document sidebar, dialogs) in docked and overlay modes, not just fullscreen
- Manage notebook selection via context state instead of URL navigation in compact modes (URL routing kept for fullscreen)
- Add compact header actions (close, add document, toggle sidebar) for overlay/docked
- Preserve live notebook streaming across display-mode switches (overlay ↔ dock ↔ fullscreen) via an external stream store mounted above the remount boundary
- Prune empty untitled ("scratch") notebooks on leave, without deleting them on a mode switch
- Stream notebook source citations during the response so the sources chip appears live (no refresh needed)

### Display-mode support
- **LightSpeedChat.tsx**: chat/notebook panel now shows in all modes (`showChatPanel = activeTab === 0`, previously fullscreen-only); tabs + divider render in compact too; notebook select/create/close set `activeNotebookId` and only `navigate()` in fullscreen; compact layout styles (`bodyCompact`, `compactDrawerPanel`, `notebooksGridCompact`); `maybeAutoDeleteScratchNotebook` prunes empty untitled notebooks on leave.
- **LightspeedDrawerContext.tsx / useLightspeedProviderState.ts**: added persisted `activeNotebookId` / `setActiveNotebookId`; stopped resetting `shellViewTab` when leaving embedded; added `isOnNotebooksPathRef` so a stale conversation-route update no longer clobbers the `/notebooks` URL (keeps the notebook open across mode switches).
- **LightspeedDrawerProvider.tsx**: mounts `NotebookStreamProvider` wrapping both the children and the overlay modal, so it is a common ancestor of every `LightSpeedChat` mount point. Also changed `overflowX: hidden` → `overflow: hidden` on the modal to prevent vertical scroll.
- **index.tsx**: drawer `resizable: true`.

### Notebook streaming architecture (survives mode switches)
- **notebookStreamStore.ts** (new): external store that owns notebook streaming, keyed by notebook `sessionId`. Reads the SSE stream, accumulates each chunk into an in-memory transcript, and notifies subscribers on every update. Exposes `send / stop / clear / clearAll`, using an `AbortController` + a `generation` counter to guard rapid re-sends and remounts. Handles `start / token / tool_call / tool_result / interrupted / end` events, temp→real conversation-id migration, and react-query invalidation on completion. The stream is aborted only on explicit intent (stop/close/delete/provider-unmount), never on view unmount.
- **NotebookStreamProvider.tsx** (new): owns the store above the display-mode remount boundary and exposes `useNotebookStream(sessionId)` (subscribes via `useSyncExternalStore`); calls `clearAll()` on unmount.
- **NotebookView.tsx**: reduced to a pure subscriber — reads the snapshot (`streamMessages`, `status`, `requestId`) and paints it; removed the old local streaming logic (`requestId` / `isSendButtonDisabled` / `autoDeleteRef`). Falls back to persisted server messages when the store is idle.
- **stream-event-helpers.ts** (new): extracted shared SSE/tool-call helpers (`parseSSEBuffer`, `parseToolCallFromEvent`, `parseToolResultFromEvent`, `applyToolResultToToolCalls`, `normalizeToolCalls`, `toolCallIdKey`, temp-cache-prefix helpers). **useConversationMessages.ts** deletes ~230 lines of inline duplicates and imports these instead; `onRequestIdReady` now also passes `conversation_id`.
- **NotebooksApiClient.ts / notebooksApi.ts / useCreateNotebookMessage.ts**: thread an optional `signal` (AbortSignal) through the create-message path.

### Source citations during streaming (backend)
- **intelligent-assistant-backend / notebooksRouters.ts**: the Responses API SSE transform now handles `response.completed` — new `extractReferencedDocuments()` collects `file_search_call` results, prefers files actually cited via `file_citation` annotations (falls back to all results), maps them to `{ doc_title, doc_url, doc_description }` (de-duped by file id), and emits a legacy `end` event carrying `referenced_documents`. Previously this push was commented out, so streamed notebook messages had no sources and the chip only appeared after a refresh (which fell back to server-loaded messages).

### Dialogs / modals (render inside the docked/overlay panel)
- **AddDocumentModal / DeleteDocumentModal / DeleteNotebookModal / OverwriteConfirmModal**: accept an `isCompact` prop and apply `getScopedDialogProps` so dialogs render inside the compact panel instead of the full window.
- **scoped-dialog-utils.ts** (new): `getScopedDialogProps` (`disablePortal`, absolute positioning, inset backdrop).
- **notebookDialogStyles.ts** (new): shared dialog styles with `*Compact` variants.

### Header actions, icons & toasts
- **NotebookHeaderActions.tsx** (new): close / add-document / sidebar-toggle buttons for compact mode.
- **SidebarCollapseIcon.tsx**: adds a `size` prop. **NotebooksTab.tsx**: `PlusCircleIcon` → `AddCircleOIcon`. **UploadResourceScreen.tsx**: `iconPosition="start"`.
- **ToastAlertGroup.tsx** (new): reusable PatternFly toast `AlertGroup` (auto-timeout + close) replacing duplicated inline alert JSX in `LightSpeedChat` and `NotebookView`.

### Tests
- **notebookStreamStore.test.ts** (new): remount survival, abort/generation races, temp-id migration, error handling, `clear`/`clearAll`.
- **lightspeed.notebooks-compact.test.ts** (new e2e): notebooks in overlay/docked — tabs, compact editor, header actions, upload modal, state preserved across tab/mode switches.
- Updated component and e2e/page-object tests to cover new context fields and compact tab visibility.

## How the notebook stream works

**What's actually happening:**
1. The backend streams SSE chunks (`token`, `tool_call`, `end`, …).
2. `runStream` reads them one at a time and appends each into an in-memory `working` array (the growing transcript).
3. After every chunk it calls `commit()` → `emit()`, which updates the store's snapshot and notifies subscribers.
4. `NotebookView`, via `useSyncExternalStore`, re-renders and paints the new token immediately.

So the UI updates live, token by token — nothing waits for completion.

**The reason it's stored outside the component (the whole point of this PR):**

The transcript doesn't live in `NotebookView`'s own React state. It lives in the `notebookStreamStore`, which is owned by `NotebookStreamProvider` — a component mounted above the overlay/docked/fullscreen remount boundary.

```
Normal chat:   tokens → component state → component unmounts → stream & tokens LOST
This design:   tokens → store (above the boundary) → component unmounts →
               stream keeps running, transcript preserved → new mount re-subscribes → sees it
```

So "storing the stream in the frontend" is doing two jobs at once:
- **Live rendering** — show tokens as they arrive (same as any chat).
- **Continuity** — because it's stored in a stable, component-independent place, switching display modes (which unmounts and remounts `NotebookView`) doesn't kill the stream or lose what's been received. The remounted view just reads the current snapshot and keeps painting.

**On the source-chip fix in this frame:** the sources aren't a separate thing — they're just one more field attached to the last bot message in that same stored transcript, delivered by the `end` event. Before the backend fix, that field never arrived over the stream, so the stored transcript had everything except sources — which is why only a refresh (which fell back to server-loaded messages) showed the chip.

## Jira
[RHIDP-14656](https://issues.redhat.com/browse/RHIDP-14656)

## Test plan
- [ ] Open lightspeed in **docked mode** → switch to Notebooks tab → verify notebooks list renders in single-column layout
- [ ] Create a new notebook in docked mode → verify it opens inline (no route navigation)
- [ ] Open a notebook → verify the compact header shows close (X), add document (+), and sidebar-toggle icons
- [ ] Verify disclaimer, footer input, and footnote stay within docked panel bounds (no overflow)
- [ ] Toggle sidebar in docked mode → verify it takes full panel width
- [ ] Open add-document / delete / overwrite dialogs in compact mode → verify they render inside the panel, not the full window
- [ ] Repeat the above in **overlay mode**
- [ ] Open lightspeed in **fullscreen mode** → verify notebooks still work with URL routing as before (no regression)
- [ ] **Mid-stream mode switch**: send a message in overlay → switch to docked mid-stream → verify tokens keep flowing, Stop works, and Send re-enables on completion
- [ ] **Mid-stream mode switch (dock → overlay)**: same in reverse
- [ ] **Mode switch after stream completes**: verify no stale disabled Send button
- [ ] **Close notebook during stream**: verify untitled empty notebooks are auto-deleted on close but preserved on mode switch
- [ ] **Source chip**: ask a question that cites documents → verify the sources chip appears live as the response completes (no page refresh)
- [ ] Run `yarn tsc`, `yarn lint`, `yarn prettier:check`, `yarn test:all` — all pass

RHIDP-14656


its-mitesh-kumar:
@rohitratannagar please add screen recordings.

rohitratannagar:
cc: @ShiranHi 

https://github.com/user-attachments/assets/7852627b-29c5-4d1b-9d65-a278f74ec684

ShiranHi:
@rohitratannagar looks good to me, thank you! I have a couple of minor points:

1. The modal header doesn't seem aligned with the main content. Is this only happening in the docked or overlay modes?
image01

image02

2. The title "My notebook" shouldn't split into two lines. Instead, the layout should be responsive, if there isn't enough space for both the title and the "Create" button, the button should drop to a new line (as shown in the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/)).
image03

rohitratannagar:
cc: @ShiranHi 
these are the latest changes in the UI


https://github.com/user-attachments/assets/683b7192-6abe-4614-9b72-b4c58201f756

karthikjeeyar:
cc: @HusneShabbir

HusneShabbir (changes requested):
When accessing notebooks in any mode other than fullscreen, the chatbot header components (Close Notebook, Add, Expand Sidebar, and Display Options) should be hidden and non-interactive. This matches the behavior in the prototype, but it hasn't been implemented in the current PR.


**Prototype:**

https://github.com/user-attachments/assets/d4169f62-edb3-4544-bbce-1feaf7e8ae0c

**PR Local behaviour:**

https://github.com/user-attachments/assets/11fae3a6-67ea-4502-987d-1041bd8992b7

HusneShabbir (changes requested):
Also, making the chatbot header components (Close Notebook, Add, and Expand Sidebar) equal-sized would improve the visual consistency and make the UI look cleaner.

**Prototype:**
Image

**UI:**
Image

HusneShabbir (changes requested):
The Expand Sidebar icon in Notebooks should mirror its direction when clicked. This behavior exists in the prototype but is missing in the current PR.

**Prototype:**

https://github.com/user-attachments/assets/df2278af-bf2e-4ddc-9644-ca60d81f179c

**PR Behaviour:**

https://github.com/user-attachments/assets/fab62227-899b-40d0-adfd-035c356e2611

rohitratannagar:
@HusneShabbir — addressed your review feedback in the latest commits:

1. **Hidden notebook header actions in compact mode** — Close Notebook, Add, Expand Sidebar, and the document sidebar are now hidden in overlay/docked modes, matching the prototype behavior.
2. **Equal-sized header action buttons** — buttons in the expand strip are now uniformly sized (36×36px).
3. **Sidebar icon mirroring** — the expand/collapse icon now flips direction based on sidebar state, with a `size` prop for compact header usage.

Additional improvements:
- Upload-in-progress state disables the Add button with an appropriate tooltip
- Header actions are dimmed and non-interactive when the upload modal is open

Could you take another look? Thanks!

HusneShabbir (approved):
/lgtm

All the changes look good, and the behavior closely resembles the prototype.

HusneShabbir:
cc: @its-mitesh-kumar

HusneShabbir (commented):
Not sure if the prototype needs an update, but I noticed one difference. In the notebook fullscreen mode, when the sidebar is collapsed, our implementation (PR #4093) still shows the **Add Resources** button, whereas the prototype hides it along with the sidebar.

https://github.com/user-attachments/assets/d4a1c5e1-dd34-4960-b7b0-2c563a59a9c9

cc: @ShiranHi

its-mitesh-kumar:
> Not sure if the prototype needs an update, but I noticed one difference. In the notebook fullscreen mode, when the sidebar is collapsed, our implementation (PR #4093) still shows the **Add Resources** button, whereas the prototype hides it along with the sidebar.
> 
>  Screen.Recording.2026-08-04.at.7.57.03.PM.mov 
> cc: @ShiranHi

Prototype need to be updated.

HusneShabbir (approved):

tobiastal:
I've updated the [prototype](https://lightspeed-plugin-477a0f.pages.redhat.com/intelligent-assistant/notebooks) (VPN required) to match the implementation. The Add Resources button now remains visible when the resource list is collapsed in fullscreen mode.
cc @HusneShabbir @its-mitesh-kumar @ShiranHi

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -283,540 +310,585 @@ type NotebookViewProps = {
...
+      onClick: () => sendMessage(title),
+    }));
+
+    const [sidebarCollapsed, setSidebarCollapsed] = useState(isCompact);
+    const [isUploadModalOpen, setIsUploadModalOpen] = useState(false);
+
+    useImperativeHandle(ref, () => ({
+      openUploadModal: () => setIsUploadModalOpen(true),
+      toggleSidebar: () => setSidebarCollapsed(prev => !prev),
+    }));
```
karthikjeeyar: The sidebar and upload-modal state is owned by this `NotebookView` but mirrored back to the parent via useEffect callbacks below in this file, and the parent talks back via an imperative `ref`. 

This means two copies of same state, you could just lift the `sidebarCollapsed` and `isUploadModalOpen` states into the parent and pass them down as props, this will remove the forwardRef, imperative handle and 2 of 3 use effects to sync states.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -283,540 +310,585 @@ type NotebookViewProps = {
...
+      onUploadsInProgressChange?.(hasUploadsInProgress);
+    }, [hasUploadsInProgress, onUploadsInProgressChange]);
+
+    useEffect(() => {
+      onSidebarCollapsedChange?.(sidebarCollapsed);
+    }, [sidebarCollapsed, onSidebarCollapsedChange]);
+
+    useEffect(() => {
+      onUploadModalOpenChange?.(isUploadModalOpen);
+    }, [isUploadModalOpen, onUploadModalOpenChange]);
```
karthikjeeyar: This ties back to my previous comment on moving the state to the parent to avoid the state mirroring logic.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/scoped-dialog-utils.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    disablePortal: true,
+    disableScrollLock: true,
+    fullWidth: true,
+    maxWidth: false,
+    sx: {
+      position: 'absolute',
+      inset: 0,
+      margin: 0,
+      // padding: 0,
```
karthikjeeyar:
```suggestion

```

karthikjeeyar (commented):
Overall looks good, I have few observations on UI and some code-wise improvements.

---

Couple of issues on the confirmation modal:

Image

1. Modal on top of a modal is not good. I know this isn't introduced by this PR but it is a easy win in my opinion to momentarily close the Upload modal behind before opening a confirmation modal and when user completes the interaction in confirmation modal we can bring back the add modal or directly upload the files after confirmation.
2. Overwrite modal styles needs some update, at least the modal header seems off.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -709,35 +742,51 @@ export const LightspeedChat = ({
...
+  } = useNotebookSession(effectiveNotebookId);
 
   useEffect(() => {
-    if (routeNotebookId && routeNotebook && !routeNotebookLoading) {
+    if (effectiveNotebookId && routeNotebook && !routeNotebookLoading) {
       setActiveNotebook(routeNotebook);
-    } else if (routeNotebookId && routeNotebookError) {
-      navigate(`${LIGHTSPEED_PATH}/notebooks`, { replace: true });
-    } else if (!routeNotebookId && notebooksRouteMatch) {
+      setActiveNotebookId(routeNotebook.session_id);
```
karthikjeeyar: Two pieces of state tracking same notebook:

  - **activeNotebookId**  coming from the drawer state, and the ID persisted across display modes
  - **activeNotebook** - full NotebookSession object saved in local useState


I am aware that `activeNotebook` wasn't introduced in this PR but the concern I have here is the maintenance.

 Every place that opened, closed, or switched a notebook had to update both in sync, miss one place then it silently fails:
 
```
  setActiveNotebook(session);
  setActiveNotebookId(session.session_id);
```

This UseEffect exists to keep them in sync, My suggestion is to remove the local state `activeNotebook` and since you have added `effectiveNotebookId`, it can help to to query the notebookSession like this

```
     const { data: activeNotebook, isError: activeNotebookError } = useNotebookSession(effectiveNotebookId);

```

With this approach you can get drop all the `setActiveNotebook(session);` calls, the sync effect reduces to just the error handling branch and every mutation only needs `setActiveNotebookId(notebookId)`

its-mitesh-kumar:
@karthikjeeyar We are fixing the modal on top of modal for full screen in this PR https://github.com/redhat-developer/rhdh-plugins/pull/3849

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -1943,6 +1984,70 @@ export const LightspeedChat = ({
                 aria-label={t('aria.chatHistoryMenu')}
               />
             )}
+            {!isFullscreenMode && showNotebooksPanel && activeNotebook && (
```
its-mitesh-kumar, line 2028: This ~60-line `Tooltip`/`PFButton` block is a self-contained unit with its own conditional — could you extract it into a `NotebookHeaderActions` component? It would keep the render function easier to scan and make the header actions independently testable.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -1943,6 +1984,70 @@ export const LightspeedChat = ({
...
+              <div
+                className={classes.notebookHeaderActions}
+                style={
+                  notebookUploadModalOpen
+                    ? { opacity: 0.4, pointerEvents: 'none' }
```
its-mitesh-kumar: `opacity: 0.4` + `pointerEvents: 'none'` blocks mouse clicks, but keyboard users can still Tab into these buttons and activate them while the upload modal is open. Would it make sense to use `isDisabled` on each button (or the `inert` attribute on the container) instead? That way `aria-disabled` gets set and keyboard nav is covered too.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -712,30 +742,25 @@ export const LightspeedChat = ({
...
-  const [activeNotebook, setActiveNotebook] = useState<NotebookSession | null>(
-    null,
-  );
-  const {
-    data: routeNotebook,
-    isLoading: routeNotebookLoading,
-    isError: routeNotebookError,
-  } = useNotebookSession(routeNotebookId);
+  const effectiveNotebookId =
+    routeNotebookId || (!isFullscreenMode ? activeNotebookId : undefined);
```
its-mitesh-kumar: Minor: `||` falls through on any falsy value (including `""`), while the intent here is "use route ID if present, otherwise context ID." `??` would express that more precisely:
```tsx
const effectiveNotebookId =
  routeNotebookId ?? (isFullscreenMode ? undefined : activeNotebookId);
```

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -683,9 +716,6 @@ export const LightspeedChat = ({
   const [filterValue, setFilterValue] = useState<string>('');
   const [announcement, setAnnouncement] = useState<string>('');
   const [activeTab, setActiveTab] = useState<number>(() => {
-    if (!isFullscreenMode) {
-      return 0;
-    }
     if (notebooksRouteMatch || notebookViewRouteMatch) {
```
its-mitesh-kumar, line 715: Now that the `if (!isFullscreenMode) { return 0; }` guard is removed, this route-match check can technically fire in compact modes — but it never will, since overlay/docked don't navigate to `/notebooks` URLs. Could you either restore a compact-mode early return using `shellViewTab`, or add a brief comment explaining why this branch is fullscreen-only in practice? It'll save the next reader some head-scratching.

its-mitesh-kumar (commented):
Review of `LightSpeedChat.tsx` — a few suggestions to tighten things up.

rohitratannagar:
Updated header to match prototype:

https://github.com/user-attachments/assets/0ac1635b-1ec9-4e36-b9e3-40117559f21d

ShiranHi:
> Updated header to match prototype:
> 
>  Screen.Recording.2026-08-11.at.11.48.48.PM.mov

Looks good! Can we update the + icon to use an outlined style? Also, when a user clicks on a chat in the list, let's automatically close the side panel so they can actually see the chat.

@tobiastal could you review this once more to ensure everything looks ok?

rohitratannagar:
> Updated header to match prototype:
> 
>  Screen.Recording.2026-08-11.at.11.48.48.PM.mov


https://github.com/user-attachments/assets/e51c2bf2-d57f-454a-b5be-3a922a46a4e8

tobiastal:
Hey @rohitratannagar, looks good!
Just a couple of small UX updates:

**On the Notebooks tab-** 

1. Move the icon so it appears before “Upload a resource.”

Current:
image

Expected:
image

2. Update the text at the bottom to: “AI content may be inaccurate. Verify before use.” instead of “Always review AI generated content prior to use.”
Also, please adjust the spacing so the text has equal padding above and below it- it currently appears slightly too low.

3.  The input area should be disabled when no resources have been added. On hover, show the following tooltip: “Select at least one loaded resource to start chatting.”

4.  A few icons are missing: the plus icon on the left, and the microphone and send icons on the right.

Current (2,3,4):
image

Expected (2,3,4):
image

**On the Chat tab-**

5. Update the input placeholder from “Send a message” to “Enter a prompt.”

6. Update the text at the bottom to: “Red Hat Developer Hub Intelligent Assistant uses AI. Check for mistakes.” instead of “Always review AI generated content prior to use.”

7. The plus and microphone icons appear to be too large, and the send icon is missing.

8. There’s some extra space between the bottom row (with the plus icon) and the bottom of the card.

Current (5,6,7,8):
image

Expected (5,6,7,8):
image

9. Add the missing icons to the cards in the chat area.

Current:
image

Expected:
image

10. In the Chat history left-side drawer, align the icons and text with the titles under each category.

Current:
image

Expected:
image

11. In docked display mode, the panel should be resizable by dragging its edge.

Current:
image

Expected:
image

Let me know if you have any questions or if anything needs clarification. Thanks!
cc @ShiranHi

HusneShabbir (commented):
cc: @its-mitesh-kumar

rohitratannagar:
> Hey @rohitratannagar, looks good! Just a couple of small UX updates:
> 
> **On the Notebooks tab-**
> 
> 1. Move the icon so it appears before “Upload a resource.”
> 
> Current: image
> 
> Expected: image
> 
> 2. Update the text at the bottom to: “AI content may be inaccurate. Verify before use.” instead of “Always review AI generated content prior to use.”
>    Also, please adjust the spacing so the text has equal padding above and below it- it currently appears slightly too low.
> 3. The input area should be disabled when no resources have been added. On hover, show the following tooltip: “Select at least one loaded resource to start chatting.”
> 4. A few icons are missing: the plus icon on the left, and the microphone and send icons on the right.
> 
> Current (2,3,4): image
> 
> Expected (2,3,4): image
> 
> **On the Chat tab-**
> 
> 5. Update the input placeholder from “Send a message” to “Enter a prompt.”
> 6. Update the text at the bottom to: “Red Hat Developer Hub Intelligent Assistant uses AI. Check for mistakes.” instead of “Always review AI generated content prior to use.”
> 7. The plus and microphone icons appear to be too large, and the send icon is missing.
> 8. There’s some extra space between the bottom row (with the plus icon) and the bottom of the card.
> 
> Current (5,6,7,8): image
> 
> Expected (5,6,7,8): image
> 
> 9. Add the missing icons to the cards in the chat area.
> 
> Current: image
> 
> Expected: image
> 
> 10. In the Chat history left-side drawer, align the icons and text with the titles under each category.
> 
> Current: image
> 
> Expected: image
> 
> 11. In docked display mode, the panel should be resizable by dragging its edge.
> 
> Current: image
> 
> Expected: image
> 
> Let me know if you have any questions or if anything needs clarification. Thanks! cc @ShiranHi

Have addressed 1,3 and 4. For 2 the warning text for chat also need to be updated, Chat is out of scope for this task, we can handled that in another PR. 

for 11: Docked mode is now resizable. However, the resizer bar used to resize the Intelligent Assistant is not part of the Intelligent Assistant plugin and will be handled in the app-defaults/app-react plugin.

https://github.com/user-attachments/assets/abecce7a-d6a9-4fa4-b21b-b2728d6cd6e9

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -184,13 +194,35 @@ const useStyles = makeStyles(theme => ({
...
+    '&.pf-v6-c-drawer__panel': {
+      width: '100%',
+      minWidth: '100%',
+      maxWidth: '100%',
+      flexBasis: '100%',
+    },
+  },
+  headerNewChatButton: {
+    '&.pf-v6-c-button': {
+      color: 'var(--pf-t--global--color--brand--default)',
```
its-mitesh-kumar: @rohitratannagar is this overriding required? Can't we get rid of it ?

its-mitesh-kumar (commented):
Also if I am switching the display mode the page user is on is not respected. 

https://github.com/user-attachments/assets/7b1ce4af-bf2a-43d7-9980-fda94b64a4c9

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
rohitratannagar: for having the blue color of plus icon to match prototype it's necessary

its-mitesh-kumar: Hi @rohitratannagar, I think we can get rid of this override by switching from `variant="plain"` to `variant="link"` on the PF Button. The `link` variant natively renders brand-colored text with correct hover and disabled states — so the entire `headerNewChatButton` class should be unnecessary.

```tsx
// Current (plain + CSS override):
<PfButton variant="plain" className={classes.headerNewChatButton} ...>

// Suggested (no override needed):
<PfButton variant="link" ...>
```

The `link` variant uses `--pf-t--global--text--color--brand--default` / `--hover` / `--on-disabled` tokens out of the box, which is what the override is manually replicating.

Ref: https://www.patternfly.org/components/button/

There might be a slight padding difference between `plain` and `link` at `size="sm"`, so it would be worth a quick visual check — but this should let us drop the custom styles entirely.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -74,12 +71,20 @@ const useStyles = makeStyles(theme => ({
...
   },
   drawerContainer: {
     flex: 1,
     minHeight: 0,
+    minWidth: 0,
+    '& .pf-v6-c-drawer__content, & .pf-v5-c-drawer__content': {
+      display: 'flex',
+      flexDirection: 'column',
+      overflow: 'hidden',
+    },
```
its-mitesh-kumar: Also Could you please look into this override, if we can get rid of it?

its-mitesh-kumar:
@rohitratannagar please focus on just supporting notebook in overlay and docked mode. If you are adding any new component/elements like topbar then only try to match with the prototype, otherwise just keep other screens/elements as-is since we will have a dedicated PR for those improvements. Let's keep the scope of this PR limited. I can see you have added many PF's style override to address the feedbacks but the old PF's style override is already headache to us, lets not increase it. 

> 1. Move the icon so it appears before "Upload a resource."

This is being handled in https://github.com/redhat-developer/rhdh-plugins/pull/3849.

> 2. Update the text at the bottom to: "AI content may be inaccurate. Verify before use." instead of "Always review AI generated content prior to use."
> 5. Update the input placeholder from "Send a message" to "Enter a prompt."
> 6. Update the text at the bottom to: "Red Hat Developer Hub Intelligent Assistant uses AI. Check for mistakes." instead of "Always review AI generated content prior to use."
> 9. Add the missing icons to the cards in the chat area.

 You can create a bug/story under https://redhat.atlassian.net/browse/RHIDP-14512 to track them separately.

> 3. The input area should be disabled when no resources have been added. On hover, show the following tooltip: "Select at least one loaded resource to start chatting."

Please handle this in this PR itself — the behavior should match how it works in fullscreen.

https://github.com/user-attachments/assets/cd334ac2-ca23-4959-881b-1751d04d39bd

> 4. A few icons are missing: the plus icon on the left, and the microphone and send icons on the right.

The `+` icon and message bar improvements are part of https://github.com/redhat-developer/rhdh-plugins/pull/3849. Also, we will not be overriding much of the PF styles.

> 7. The plus and microphone icons appear to be too large, and the send icon is missing.
> 8. There's some extra space between the bottom row (with the plus icon) and the bottom of the card.

We are avoiding overriding PF styles — these come from the PF component itself and will mostly remain as-is.

> 10. In the Chat history left-side drawer, align the icons and text with the titles under each category.

This padding/spacing needs to be fixed upstream. An issue has already been raised: https://github.com/patternfly/chatbot/issues/848.

> 11. In docked display mode, the panel should be resizable by dragging its edge.

I believe this is fixed now.

Hope we are aligned on it @ShiranHi  @tobiastal . Let me know if need any clarification, happy to discuss more on it 🤝 .

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -74,12 +71,20 @@ const useStyles = makeStyles(theme => ({
...
     '& .pf-v6-c-drawer__panel, & .pf-v5-c-drawer__panel': {
```
its-mitesh-kumar, line 96: Also Could you please look into this override, if we can get rid of it?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -248,22 +248,78 @@ const useStyles = makeStyles(theme => ({
...
+      margin: '0 !important',
+      gap: '4px !important',
     },
     '& .pf-chatbot__message-bar': {
       backgroundColor:
         theme.palette.type === 'light'
           ? theme.palette.grey[100]
           : 'var(--pf-t--global--background--color--secondary--default)',
     },
+    '& .pf-chatbot__button--stop, & .pf-chatbot__button--attach, & .pf-chatbot__button--send, & .pf-chatbot__button--microphone':
```
its-mitesh-kumar: Also Could you please look into this override, if we can get rid of it?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -248,22 +248,78 @@ const useStyles = makeStyles(theme => ({
...
+    },
     '& .pf-chatbot__button--send, & .pf-chatbot__button--microphone': {
-      '--pf-v6-c-button--BorderRadius':
-        'var(--pf-t--global--border--radius--pill)',
-      borderRadius: 'var(--pf-t--global--border--radius--pill) !important',
+      opacity: '0.5 !important',
     },
   },
+  sendButtonDimmed: {
+    '& .pf-chatbot__button--send, & .pf-chatbot__button--send svg, & .pf-chatbot__button--send .pf-v6-c-button__icon':
```
its-mitesh-kumar: Also Could you please look into this override, if we can get rid of it?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -248,22 +248,78 @@ const useStyles = makeStyles(theme => ({
...
+      {
+        opacity: '0.5 !important',
+        color: 'inherit !important',
+        transition: 'opacity 0.15s',
+      },
+  },
+  sendButtonActive: {
+    '& .pf-chatbot__button--send, & .pf-chatbot__button--send svg, & .pf-chatbot__button--send .pf-v6-c-button__icon':
+      {
```
its-mitesh-kumar: Also Could you please look into this override, if we can get rid of it?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/ToastAlertGroup.tsx
```diff
@@ -0,0 +1,82 @@
...
+  Alert,
+  AlertActionCloseButton,
+  AlertGroup,
+  AlertVariant,
+  type AlertProps,
+} from '@patternfly/react-core';
+
+const useStyles = makeStyles(theme => ({
+  toastAlertGroup: {
+    '--pf-v6-c-alert-group--m-toast--InsetInlineEnd': `${theme.spacing(2.5)}px`,
```
its-mitesh-kumar: Also Could you please look into this if its required?

HusneShabbir (changes requested):
When a notebook is created, changing the display mode and then trying to rename the notebook causes the rename to fail, and the notebook gets deleted automatically.

Also, when a notebook is created in other display modes (e.g., Docto Window or Overlay mode), it does not persist when switching to Full Screen mode.

https://github.com/user-attachments/assets/287b4b25-250b-4ede-9310-475541222330

HusneShabbir (changes requested):
Additionally, file attachments are also failing when the notebook is used in other display modes.

https://github.com/user-attachments/assets/2167beba-ed8d-4014-bad7-42d85511af71

HusneShabbir (commented):
Shouldn’t the same consideration mentioned in [(https://github.com/redhat-developer/rhdh-plugins/pull/3849#pullrequestreview-4897282565)](https://github.com/redhat-developer/rhdh-plugins/pull/3849#pullrequestreview-4897282565) apply here as well?

https://github.com/user-attachments/assets/053aeaa0-1531-41c0-9674-778b6762858a

cc: @its-mitesh-kumar

HusneShabbir (changes requested):
When a user switches the display mode from dock to window and then to full-screen after using it for a while, the notebook unexpectedly closes when entering full-screen mode & **the issue is reproducible**

https://github.com/user-attachments/assets/34bcbb7b-a958-4830-9fdc-c1254a47d903

HusneShabbir (changes requested):
**1) Fullscreen mode / empty notebook deletion**

When a user opens a notebook and opts for fullscreen mode, **even after already being in fullscreen mode**, then closes the empty notebook without doing anything else, the expectation is that the notebook should be deleted. However, it currently isn’t, which is causing the test to fail. Please refer to the recording below for a better understanding.

I did some debugging, and Cursor points to this being an implementation issue. I’ll leave it to you to decide the best way to address it.

https://github.com/user-attachments/assets/173ac35b-a6ee-4243-85a9-95bba25b8741

**2) Switching to fullscreen from other display modes**

Also, when a user creates a notebook in any other display mode and then tries to switch to fullscreen mode, the notebook still closes. I’d encourage you to take another look at this behavior as well.

https://github.com/user-attachments/assets/8ae75236-4ec2-46df-af75-ebd513347555

HusneShabbir (commented):
Thanks for fixing **1) Fullscreen mode / empty notebook deletion**. However, **2) Switching to fullscreen from other display modes is still reproducible**.

https://github.com/user-attachments/assets/9ade5cf0-00f3-44e5-8e33-5dc40971fb6e

I’d appreciate it if @its-mitesh-kumar  / @karthikjeeyar could look into this and help @rohitratannagar understand that the issue isn’t specific to my system, so it can be fixed accordingly.

HusneShabbir (commented):
I suspect that when a user has notebooks open in another display mode and switches to fullscreen, the notebooks may be getting closed automatically. Could you please look into this?

This could also explain why empty notebooks are being deleted during the transition between display modes.


https://github.com/user-attachments/assets/945289cc-425e-4d7c-8466-c9c239b4bf5d

its-mitesh-kumar:
@HusneShabbir, it's not reproducible for me !!! 

https://github.com/user-attachments/assets/27dc8b7b-c69b-4af4-8139-e83f9fab374b


https://github.com/user-attachments/assets/138d4df3-ee98-4c7e-976b-5fb77ff6dd1f

cc: @rohitratannagar

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -261,41 +266,23 @@ const useStyles = makeStyles(theme => ({
...
-    border: '1px solid var(--pf-t--global--border--color--default)',
-    borderRadius: 24,
-    padding: theme.spacing(0.5),
-    '&::after': {
-      display: 'none',
+    '& .pf-chatbot__button--send, & .pf-chatbot__button--microphone': {
+      '--pf-v6-c-button--BorderRadius':
+        'var(--pf-t--global--border--radius--pill)',
+      borderRadius: 'var(--pf-t--global--border--radius--pill) !important',
     },
```
its-mitesh-kumar: I think lets keep this as default style whatever is coming from PF components, if we really need style improvement please create an upstream issue for the same on https://github.com/patternfly/chatbot/issues

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
```diff
@@ -261,41 +266,23 @@ const useStyles = makeStyles(theme => ({
...
     display: 'flex',
     flexDirection: 'column',
     flex: 1,
     overflow: 'auto',
+    '& .pf-chatbot__message-contents': {
+      overflowX: 'hidden',
+      overflowWrap: 'break-word',
+      wordBreak: 'break-word',
+    },
   },
```
its-mitesh-kumar, line 255: If needed We may keep it for now, but create an upstream issue on https://github.com/patternfly/chatbot/issues and put the upstream issue link as part of comment in the code.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -163,6 +163,16 @@ const useStyles = makeStyles(theme => ({
...
   },
+  bodyCompact: {
+    height: '100% !important',
+    minHeight: '0 !important',
+    overflow: 'hidden',
+    '& .pf-chatbot-container': {
+      minHeight: '0 !important',
+      display: 'flex',
+      flexDirection: 'column',
+    },
```
its-mitesh-kumar: Please check it once if this is also required ?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/LightSpeedChat.tsx
```diff
@@ -184,13 +194,24 @@ const useStyles = makeStyles(theme => ({
...
+    display: 'flex',
+    alignItems: 'center',
+    gap: theme.spacing(0.5),
+  },
+  compactDrawerPanel: {
+    '&.pf-v6-c-drawer__panel': {
+      width: '100%',
+      minWidth: '100%',
+      maxWidth: '100%',
+      flexBasis: '100%',
```
its-mitesh-kumar, line 204: Please check it once if this is also required ?

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/ToastAlertGroup.tsx
```diff
@@ -0,0 +1,72 @@
...
+  AlertActionCloseButton,
+  AlertGroup,
+  AlertVariant,
+  type AlertProps,
+} from '@patternfly/react-core';
+
+const useStyles = makeStyles(() => ({
+  toastAlertGroup: {
+    '--pf-v6-c-alert-group--m-toast--ZIndex': '9999',
+  },
```
its-mitesh-kumar, line 34: Is this really needed ?

rohitratannagar: Docked mode drawer is set on 1300, so it is needed.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
rohitratannagar: removed!!

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
rohitratannagar: Needed for, while stream is coming in full-screen and then you switch to overlay/docked mode.

HusneShabbir (commented):
/lgtm

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/components/notebooks/NotebookView.tsx
karthikjeeyar: Please create upstream issue as Mitesh mentioned above and share it

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/notebooks/notebooksRouters.ts
```diff
@@ -70,6 +69,103 @@ export interface NotebooksRouterOptions {
   permissions: PermissionsService;
 }
 
+/**
```
its-mitesh-kumar: @JslYoon Could you please look into the backend changes.

its-mitesh-kumar:
@rohitratannagar Please create a bug/follow up pr for 

1.  Alert is not looking good.
S_ 2026-09-01 at 11 58 27 AM

2. User should be able to rename the Notebook from here as well.
S_ 2026-09-01 at 11 59 42 AM

3. In overlay/docked mode in the first response `sources` are not coming, they are appearing after refresh.

its-mitesh-kumar (approved):
/lgtm

https://github.com/user-attachments/assets/62e626f2-af36-43cb-ae7e-45073f885bc5

HusneShabbir (approved):

karthikjeeyar:
Please review the full send messages, see if there is anything you can address, add comments and resolve the open comments. Maybe then the merge button will be enabled.

karthikjeeyar (approved):

PR #4104, opened 2026-07-30
fix(theme): make `app` optional in config schema

christoph-jerolimov:
## Summary
- Makes the `app` property optional in the theme plugin's `config.d.ts` schema
- Avoids requiring the `app` key when no branding is configured

## Test plan
- [ ] Verify the theme plugin works without an `app` key in `app-config.yaml`
- [ ] Verify existing configurations with `app.branding` still work as expected


karthikjeeyar (approved):
/approve
/lgtm

PR #4110, opened 2026-07-30
feat(install-dynamic-plugins): add ref:// plugin reference resolution

jonkoops:
Adds `ref://` plugin reference resolution to the init container's config merger, matching the behaviour of the operator.

Plugin configurations can now use `ref://plugin-name` to reference another plugin by its extracted name, instead of repeating full OCI or HTTP package URLs. The resolver scans already-merged plugins, extracts their names from the package URL (OCI, HTTP, or local path), and replaces the ref with the matched package string.

NPM package references (e.g. `@backstage/plugin-catalog`) are not supported as `ref://` targets, as `ref://` is primarily designed to simplify references to OCI-distributed plugins in the catalog where full image URLs with digests would otherwise need to be repeated.

Ref: [RHIDP-15875](https://redhat.atlassian.net/browse/RHIDP-15875)

### DIFF
--- a/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.test.ts
+++ b/workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.test.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 import { InstallException } from './errors';
-import { deepMerge, mergePlugin } from './merger';
+import { deepMerge, mergePlugin, resolveRefPlugin } from './merger';
 import type { PluginMap } from './types';
 
 describe('deepMerge', () => {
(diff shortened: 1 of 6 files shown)


gashcrumb (approved):
Nice!  Do you have other stuff to add before creating a new release of this package; or if not does this need a changeset?

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.test.ts
```diff
@@ -138,3 +138,80 @@ describe('mergePlugin — NPM', () => {
...
   });
 });
+
+describe('mergePlugin — ref://', () => {
+  it('resolves ref to an OCI plugin by name', async () => {
+    const all: PluginMap = {};
+    await mergePlugin(
+      {
+        package:
+          'oci://quay.io/rhdh/backstage-plugin-foo@sha256:abc123!backstage-plugin-foo',
```
nickboldt, line 148: ugh are we still supporting the redundant `!<path>` suffix?

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/plugin-name.ts
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+import { basename } from 'node:path';
+import { isHttpUrl, isLocalPath, isOciUrl } from './protocols';
+
+/**
+ * Extract the human-readable plugin name from a package URL.
+ *
+ * Supports OCI (`oci://`), HTTP(S) (`.tgz`/`.tar.gz` archives), and
+ * local (`./`) paths. Returns `null` for unrecognized formats.
```
nickboldt, line 23: local paths? in this economy? 

Is this for wrappers and if so... can we at least mark this DEPRECATED?

davidfestal (commented):
I'm wondering whether it is expected to allow smooth transition from the wrappers local references to oci artifacts ?
If that's the case, how do we manage the `-dynamic` suffix at the end of the wrapper folder name ?

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.test.ts
jonkoops: It's used in the equivalent operator code, so I mirrored the behavior from there.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/plugin-name.ts
jonkoops: This is also mirrored from the same functionality in the operator.

jonkoops:
> I'm wondering whether it is expected to allow smooth transition from the wrappers local references to oci artifacts ? If that's the case, how do we manage the `-dynamic` suffix at the end of the wrapper folder name ?

There will need to be documentation for customers to migrate over; this will be a breaking change for 2.1. Either way, it is not a concern of this PR, but a general issue that we need to handle as part of the wrapper removal.

The same goes for removing support for local paths and HTTP URL base `ref://` plugin references. We should consider their removal holistically and not leave gaps between implementations in the interim.

davidfestal (approved):

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/plugin-name.ts
gazarenkov: Yes, it is rhdh image's path, so if we do not support wrappers, no need to support/deprecate it I think.

workspaces/install-dynamic-plugins/packages/install-dynamic-plugins/src/merger.test.ts
gazarenkov: would love to not supporting "!" but not sure it is decided.

gazarenkov (approved):

PR #4131, opened 2026-07-31
fix(fullsend): allow rhdh-skill URLs on code and fix harnesses

johnmcollier:
## Summary
- Code/fix dispatch was failing with `URL .../rhdh-skill/.../rhdh-coding is not in allowed_remote_resources` (e.g. [run 30655318049](https://github.com/redhat-developer/rhdh-plugins/actions/runs/30655318049/job/91238191583)).
- [#4096](https://github.com/redhat-developer/rhdh-plugins/pull/4096) added the skill and the **config.yaml** allowlist, but URL skills are resolved against the **harness** `allowed_remote_resources`, which is not inherited from `base:` or config.
- Add `https://github.com/redhat-developer/rhdh-skill/` to `.fullsend/rhdh/harness/{code,fix}.yaml` (still covered by the existing config.yaml entry).

## Test plan
- [ ] Confirm CI/dispatch for a `/fs-code` (or issue code trigger) gets past remote resource resolution
- [ ] Spot-check that harness prefixes remain covered by `.fullsend/config.yaml` `allowed_remote_resources`


gabemontero (approved):

rohitkrai03 (approved):
/approve

PR #4155, opened 2026-08-04
FLPATH-4285 | [DCM] Server-side pagination not implemented

asmasarw:
## Summary

Add server-side cursor pagination to all DCM tabs.

Previously only Service Types, Catalog Items, and Instances fetched data page-by-page. Providers, Policies, and Resources loaded everything in a single call, silently dropping records beyond the first page.

### What changed

- **Providers & Policies** — converted to cursor pagination with Next/Previous controls; search resets to page 1 client-side.
- **Resources** — converted to `usePaginatedFetch` (read-only layout preserved).
- **`dcm-common`** — extracted `buildPaginationQuery` utility; `listProviders` and `listPolicies` now accept optional `PaginationParams`; `next_page_token` made optional on list response types to match real backend behaviour.
- **Dropdown calls capped at 25 items** — service-type and catalog-item selects now pass `max_page_size: 25`.
- **Tests** — added `ProvidersTabContent.test.tsx` and `ResourcesTabContent.test.tsx`; extended `PoliciesTabContent.test.tsx` with cursor navigation tests.
- **Translations** — added `common.previousPage` / `common.nextPage` to all locale files.

### DIFF
--- a/workspaces/dcm/.changeset/server-side-pagination-all-tabs.md
+++ b/workspaces/dcm/.changeset/server-side-pagination-all-tabs.md
@@ -1,0 +1,28 @@
+---
+'@red-hat-developer-hub/backstage-plugin-dcm': minor
+'@red-hat-developer-hub/backstage-plugin-dcm-common': minor
+---
+
+Add server-side cursor pagination to all tabs and harden pagination handling across APIs.
+
+**All six tabs now use server-side cursor pagination**
+
+Previously only Service Types, Catalog Items, and Catalog Item Instances fetched data page-by-page from the backend. Providers, Policies, and Resources loaded everything in a single call and silently lost records beyond the first page.
+
+- **Providers** and **Policies** — converted to the `useCrudTab` + manual token-stack pattern (same as Catalog Items). Next / Previous buttons appear below the table; search resets to page 1 client-side without an extra round-trip.
+- **Resources** — converted to `usePaginatedFetch` (same as Service Types). Retains the same read-only layout.
+
+**`dcm-common`: new pagination utilities and updated API interfaces**
+
+- `buildPaginationQuery` extracted from `CatalogClient` into `dcm-common/src/utils/buildPaginationQuery.ts` and exported publicly so all clients can share the same URL-builder.
+- `ProvidersApi` / `ProvidersClient` — `listProviders` now accepts an optional `PaginationParams` argument.
+- `PolicyManagerApi` / `PolicyManagerClient` — `listPolicies` now accepts an optional `PaginationParams` argument.
+- `ServiceTypeList`, `CatalogItemList`, `CatalogItemInstanceList` — `next_page_token` is now optional (`?`) to match the real backend behaviour where the field is absent (not just empty) when there is only one page of results.
+
+**Dropdown calls capped at 25 items**
+
+Service-type dropdown loads (used in the Providers and Catalog Items create/edit forms) and the catalog-item dropdown load (used in the Instances create form) now pass `max_page_size: 25`. This prevents the select from silently showing an incomplete list when there are more than the default page size of items.
+
+**Test coverage**
+
+Added `ProvidersTabContent.test.tsx` and `ResourcesTabContent.test.tsx` with full cursor navigation test suites (initial load, error/retry, Next/Previous button states and token passing). Updated `PoliciesTabContent.test.tsx` with equivalent cursor navigation tests and refreshed mock return types.
--- a/workspaces/dcm/.changeset/server-side-pagination-all-tabs.md
+++ b/workspaces/dcm/.changeset/server-side-pagination-all-tabs.md
@@ -9,7 +9,7 @@ Add server-side cursor pagination to all tabs and harden pagination handling acr
 
 Previously only Service Types, Catalog Items, and Catalog Item Instances fetched data page-by-page from the backend. Providers, Policies, and Resources loaded everything in a single call and silently lost records beyond the first page.
 
-- **Providers** and **Policies** — converted to the `useCrudTab` + manual token-stack pattern (same as Catalog Items). Next / Previous buttons appear below the table; search resets to page 1 client-side without an extra round-trip.
+- **Providers** and **Policies** — converted to `usePaginatedCrudTab` (same pattern as Catalog Items). Next / Previous buttons appear below the table; search filters the current page client-side without an extra round-trip.
 - **Resources** — converted to `usePaginatedFetch` (same as Service Types). Retains the same read-only layout.
 
 **`dcm-common`: new pagination utilities and updated API interfaces**
@@ -19,9 +19,9 @@ Previously only Service Types, Catalog Items, and Catalog Item Instances fetched
 - `PolicyManagerApi` / `PolicyManagerClient` — `listPolicies` now accepts an optional `PaginationParams` argument.
 - `ServiceTypeList`, `CatalogItemList`, `CatalogItemInstanceList` — `next_page_token` is now optional (`?`) to match the real backend behaviour where the field is absent (not just empty) when there is only one page of results.
 
-**Dropdown calls capped at 25 items**
+**Dropdown options loaded once on mount**
 
-Service-type dropdown loads (used in the Providers and Catalog Items create/edit forms) and the catalog-item dropdown load (used in the Instances create form) now pass `max_page_size: 25`. This prevents the select from silently showing an incomplete list when there are more than the default page size of items.
+Service-type dropdown loads (used in the Providers and Catalog Items create/edit forms) are now fetched once on component mount via a dedicated `useEffect`, not on every page navigation. The request uses `max_page_size: 100` to avoid silently truncating valid options.
 
 **Test coverage**
 
--- a/workspaces/dcm/plugins/dcm/src/components/CursorPaginationControls.tsx
+++ b/workspaces/dcm/plugins/dcm/src/components/CursorPaginationControls.tsx
@@ -63,7 +63,7 @@ export function CursorPaginationControls({
   loading = false,
   pageSize,
   onPageSizeChange,
-  pageSizeOptions = [5, 15, 25],
+  pageSizeOptions = [5, 10, 25],
 }: CursorPaginationControlsProps) {
   const classes = useStyles();
   const { t } = useTranslation();
@@ -87,7 +87,7 @@ export function CursorPaginationControls({
         </Select>
       )}
       <Tooltip title={t('common.previousPage')}>
-        <span>
+        <Box component="span">
           <IconButton
             size="small"
             disabled={!hasPrev || loading}
@@ -96,10 +96,10 @@ export function CursorPaginationControls({
           >
             <ChevronLeftIcon />
           </IconButton>
-        </span>
+        </Box>
       </Tooltip>
       <Tooltip title={t('common.nextPage')}>
-        <span>
+        <Box component="span">
           <IconButton
             size="small"
             disabled={!hasNext || loading}
@@ -108,7 +108,7 @@ export function CursorPaginationControls({
           >
             <ChevronRightIcon />
           </IconButton>
-        </span>
+        </Box>
       </Tooltip>
     </Box>
   );
--- a/workspaces/dcm/plugins/dcm/src/components/CursorPaginationControls.tsx
+++ b/workspaces/dcm/plugins/dcm/src/components/CursorPaginationControls.tsx
@@ -1,0 +1,150 @@
+/*
+ * Copyright Red Hat, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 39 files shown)


workspaces/dcm/plugins/dcm/src/hooks/usePaginatedFetch.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  const resetToFirstPage = useCallback(() => {
+    currentTokenRef.current = undefined;
+    tokenStackRef.current = [];
+    setTokenStack([]);
+    setNextToken('');
+    fetch();
+  }, [fetch]);
+
+  const resetCursor = useCallback(() => {
```
mareklibra: `resetCursor()` clears `nextToken` without a refetch, and Service Types / Resources call it on every `search` change. After any search interaction, Next is permanently disabled until remount/page-size change (even after clearing the query) because the token is never restored.

Please either:
1. Remove these `resetCursor()`-on-search effects (keep Prev/Next tied to the loaded page), or
2. Call `resetToFirstPage()` so search resets to page 1 with a real reload.

Also add a regression test: load page 1 with `nextPageToken`, change search, clear search, assert Next is still enabled / token preserved or page 1 re-fetched.

workspaces/dcm/plugins/dcm/src/hooks/usePaginatedCrudTab.ts
```diff
@@ -0,0 +1,211 @@
...
+      tokenStackRef.current = [];
+      setTokenStack([]);
+      crudReload();
+    },
+    [setPageSize, crudReload],
+  );
+
+  // When the search changes, reset cursor navigation state (no API call
+  // needed — search filters the current page client-side).
+  const handleSearchChange = useCallback(
```
mareklibra, line 180: It clears `currentTokenRef` + token stack but leaves `nextPageToken` and the displayed page data as-is. Comments say this “resets to page 1”, but the user can still be viewing page N rows with Next enabled. Prefer not touching cursor state for client-side filter, or call a real `resetToFirstPage` + reload.

workspaces/dcm/plugins/dcm/src/pages/providers/ProvidersTabContent.tsx
```diff
@@ -72,15 +72,26 @@ export function ProvidersTabContent() {
...
+        providersApi
+          .listProviders({
+            page_token: pageToken,
+            max_page_size: ps,
+          })
+          .then(r => ({
+            items: r.providers ?? [],
+            nextPageToken: r.next_page_token,
+          })),
+        catalogApi
```
mareklibra: Capping form dropdowns at `max_page_size: 25` does not “prevent incomplete lists” (changeset wording). API default is already 100; 25 truncates sooner with no “load more” / typeahead. Prefer `100`, full pagination of options, or an explicit incomplete-list indicator.

workspaces/dcm/plugins/dcm/src/hooks/usePaginatedCrudTab.ts
```diff
@@ -0,0 +1,211 @@
...
+  // Stack of tokens for previously-visited pages (enables Previous navigation).
+  const [tokenStack, setTokenStack] = useState<string[]>([]);
+  const tokenStackRef = useRef<string[]>([]);
+
+  // Keep the latest options in a ref so the stable loadFn wrapper always
+  // calls the freshest implementation without needing it in its dep array.
+  const optsRef = useRef(options);
+  optsRef.current = options;
+
+  const crud = useCrudTab<T, F>({
```
mareklibra, line 133: `usePaginatedCrudTab` calls `usePersistedPageSize(storageKey)` and then spreads `...options` (including `storageKey`) into `useCrudTab`, which calls `usePersistedPageSize` again on the same key. Two React states back the same localStorage entry; only the outer one is updated on page-size change. Destructure `storageKey` out before spreading into `useCrudTab`.

workspaces/dcm/plugins/dcm/src/components/DcmCrudTabLayout.tsx
mareklibra: `items.length === 0` always renders the illustration empty-state and drops cursor controls. After deleting the last row on page 2+, or an empty cursor page with `hasPrev`, the user cannot go Previous. In cursor mode, treat “empty current page + hasPrev” as an empty table with pagination (or auto-`goPrev`), not the global empty state.

workspaces/dcm/plugins/dcm/src/hooks/usePaginatedCrudTab.ts
```diff
@@ -0,0 +1,211 @@
...
+  loadFn: (params: PaginatedLoadParams) => Promise<T[] | PagedLoadResult<T>>;
+  /**
+   * `localStorage` key used to persist the selected page size (e.g.
+   * `'providers'`, `'policies'`).  Must be unique per table.
+   */
+  storageKey: string;
+}
+
+/** Props that can be passed directly to {@link DcmCrudTabLayout}'s `cursorPagination`. */
+export interface CursorPaginationProps {
```
mareklibra: Three near-identical pagination prop types will drift. Export one from `CursorPaginationControls.tsx` and reuse it.

workspaces/dcm/plugins/dcm/src/components/CursorPaginationControls.tsx
```diff
@@ -0,0 +1,150 @@
...
+    display: 'flex',
+    justifyContent: 'flex-end',
+    alignItems: 'center',
+    gap: theme.spacing(0.5),
+    padding: theme.spacing(1, 2),
+    borderTop: `1px solid ${theme.palette.divider}`,
+  },
+}));
+
+export type CursorPaginationControlsProps = Readonly<{
```
mareklibra, line 35: Three near-identical pagination prop types will drift. Export one from `CursorPaginationControls.tsx` and reuse it.

workspaces/dcm/plugins/dcm/src/components/dcmTabListHelpers.tsx
```diff
@@ -107,12 +108,26 @@ function ActionsCell({ onEdit, onDelete }: ActionsCellProps) {
   );
 }
 
+export type CursorPaginationProps = Readonly<{
```
mareklibra: Three near-identical pagination prop types will drift. Export one from `CursorPaginationControls.tsx` and reuse it.

workspaces/dcm/.changeset/server-side-pagination-all-tabs.md
```diff
@@ -0,0 +1,28 @@
...
+'@red-hat-developer-hub/backstage-plugin-dcm-common': minor
+---
+
+Add server-side cursor pagination to all tabs and harden pagination handling across APIs.
+
+**All six tabs now use server-side cursor pagination**
+
+Previously only Service Types, Catalog Items, and Catalog Item Instances fetched data page-by-page from the backend. Providers, Policies, and Resources loaded everything in a single call and silently lost records beyond the first page.
+
+- **Providers** and **Policies** — converted to the `useCrudTab` + manual token-stack pattern (same as Catalog Items). Next / Previous buttons appear below the table; search resets to page 1 client-side without an extra round-trip.
```
mareklibra: Changeset still says Providers/Policies use a “manual token-stack around useCrudTab”, but the code uses `usePaginatedCrudTab`. Please align the changeset with the final design.

mareklibra (changes requested):
- No `usePaginatedCrudTab` unit tests, no `buildPaginationQuery` tests, no search+`resetCursor` regression, and `ProvidersClient.test` still only asserts unpaginated `GET /providers`. Worth adding at least query-param client coverage and the search/Next regression above.

- Cursor controls default to `[5, 15, 25]` while the Table pager uses `[5, 10, 25]`. Align them unless 15 is intentional.

- Providers/Catalog loadFns re-fetch service-types/catalog-items (capped at 25) on every Next/Prev. Load dropdown options once (separate effect) to avoid redundant traffic.

workspaces/dcm/plugins/dcm/src/pages/catalog-item-instances/CatalogItemInstancesTabContent.tsx
```diff
@@ -87,15 +87,26 @@ export function CatalogItemInstancesTabContent() {
...
+          .listCatalogItemInstances({
+            page_token: pageToken,
+            max_page_size: ps,
+          })
+          .then(r => ({
+            items: r.results ?? [],
+            nextPageToken: r.next_page_token,
+          })),
+        catalogApi
+          .listCatalogItems({ max_page_size: 25 })
```
mareklibra: Catalog Items / Providers were moved to a one-shot mount fetch with
`max_page_size: 100`, but Catalog Item Instances still does
`listCatalogItems({ max_page_size: 25 })` inside `loadFn` via
`Promise.all` — so every Next/Prev re-hits catalog-items and the create
dropdown silently truncates after 25.

Please mirror Providers/Catalog Items: load catalog items once in a
`useEffect` with `max_page_size: 100` (or paginate/typeahead), and keep
`loadFn` to instances only.

mareklibra: Sort of:

```
useEffect(() => {
  catalogApi
    .listCatalogItems({ max_page_size: 100 })
    .then(r => setCatalogItems(r.results ?? []))
    .catch(() => {});
}, [catalogApi]);

const crud = usePaginatedCrudTab({
  loadFn: ({ pageToken, pageSize: ps }) =>
    catalogApi
      .listCatalogItemInstances({ page_token: pageToken, max_page_size: ps })
      .then(r => ({
        items: r.results ?? [],
        nextPageToken: r.next_page_token,
      })),
  // ...
});
```

mareklibra: Also update the changeset line that only mentions Providers/Catalog Items dropdowns.

workspaces/dcm/plugins/dcm/src/hooks/usePaginatedCrudTab.ts
```diff
@@ -0,0 +1,203 @@
...
+ * });
+ *
+ * // In JSX:
+ * <DcmCrudTabLayout
+ *   onSearchChange={crud.handleSearchChange}
+ *   cursorPagination={crud.cursorPagination}
+ *   ...
+ * />
+ */
+export function usePaginatedCrudTab<T, F extends Record<string, unknown>>(
```
mareklibra, line 110: `usePaginatedCrudTab` is the shared cursor stack for four CRUD tabs, but
there’s still no dedicated unit test file. Tab tests help, but they won’t
catch hook regressions (token stack / page-size reset / storageKey
stripping) as cheaply.

Please add `usePaginatedCrudTab.test.ts` covering: initial load params,
goNext/goPrev token passing, handlePageSizeChange → page 1 reload, and
search leaving cursor/next token untouched.

workspaces/dcm/plugins/dcm/src/pages/catalog-items/CatalogItemsTabContent.tsx
```diff
@@ -103,15 +103,23 @@ export function CatalogItemsTabContent() {
...
-      ]);
-      setServiceTypes(serviceTypeList);
-      return itemList;
-    },
+  // Fetch dropdown options once on mount; page navigation does not re-fetch.
+  useEffect(() => {
+    catalogApi
+      .listServiceTypes({ max_page_size: 100 })
+      .then(r => setServiceTypes(r.results ?? []))
+      .catch(() => {});
```
mareklibra: `.catch(() => {})` on the mount-time service-types fetch leaves create/edit
forms with an empty dropdown and no error. At least surface a non-blocking
alert or log via the app logger so failures aren’t silent.

workspaces/dcm/plugins/dcm/src/components/DcmCrudTabLayout.tsx
```diff
@@ -60,11 +61,27 @@ export type DcmCrudTabLayoutProps<T extends object> = Readonly<{
...
+  pageSize?: number;
+  onPageChange?: (page: number, pageSize: number) => void;
+  onRowsPerPageChange?: (pageSize: number) => void;
+
+  /**
+   * When provided, server-side cursor-based pagination is used instead of the
+   * Backstage Table's built-in pager. The table is rendered with `paging:
+   * false` and {@link CursorPaginationControls} is shown below it.
+   */
+  cursorPagination?: {
```
mareklibra, line 75: `.catch(() => {})` on the mount-time service-types fetch leaves create/edit
forms with an empty dropdown and no error. At least surface a non-blocking
alert or log via the app logger so failures aren’t silent.

workspaces/dcm/plugins/dcm/src/components/DcmCrudTabLayout.tsx
```diff
@@ -237,28 +259,38 @@ export function DcmCrudTabLayout<T extends object>({
...
-            page={page}
-            onPageChange={onPageChange}
-            onRowsPerPageChange={onRowsPerPageChange}
-            localization={{
-              pagination: { labelRowsPerPage: t('common.rows') },
-            }}
-          />
+          {cursorPagination ? (
+            <CursorPaginatedTable<T>
+              data={filtered}
```
mareklibra, line 266: In cursor mode the table now uses `data={filtered}` (current server page).
The card title still uses `filtered.length`, so the number is page-local,
not a dataset total. Consider dropping the count, labeling it as “showing”,
or only showing it when a real total exists.

mareklibra (changes requested):
Adding new comments. Something is left from the last time.

mareklibra (approved):

PR #4160, opened 2026-08-04
feat(scorecard): disable annotations customization via config

dzemanov:

- Adds global disable of customization by entity annotations with `scorecard.entityAnnotations.enabled` in `app-config`
- Disable or restrict threshold customization by entity annotations with `scorecard.entityAnnotations.thresholds` in `app-config`
- Follows pattern introduced for disabled metrics in https://github.com/redhat-developer/rhdh-plugins/pull/2393

Example configuration:
```yaml
scorecard:
  entityAnnotations:
    # Global switch: if false, all scorecard entity annotations are ignored
    # Default: true.
    enabled: true
    thresholds:
      # If false, threshold override annotations have no effect. Default: true.
      enabled: true
      # When thresholds.enabled is true: metric IDs listed here cannot have
      # thresholds customized via entity annotations.
      except:
        - github.openPRs
```

#### Fixes
Fixes https://redhat.atlassian.net/browse/RHIDP-11724

#### How to test
Navigate to `github-scorecard-only.yaml` to see the result of applied changes.

**1. global disable**

  Uncomment annotations in `github-scorecard-only.yaml` for disabled metrics and thresholds customization:
  ```
  annotations:
      scorecard.io/disabled-metrics: github.openPRs
      scorecard.io/github.openPRs.thresholds.rules.warning: '10-25'
      scorecard.io/github.openPRs.thresholds.rules.error: '>25'
  ```
  
  Add to app-config:
  ```
  scorecard:
    entityAnnotations:
      enabled: false
  ```
  
  Entity annotations are not applied, `github.openPRs` metric runs and default number thresholds are used.

**2. global enable for disabled metrics annotations**
  Enable global entity annotations
  ```
  scorecard:
    entityAnnotations:
      enabled: true
  ```
  Entity annotations are applied, `github.openPRs` metric doesn't run.

**3. global enable for threshold annotations**
  Comment out disabling metric:
  ```
  annotations:
      # scorecard.io/disabled-metrics: github.openPRs
  ```
  
  Custom entity threshold annotations are applied.

**4. Disable threshold annotations**
  Disable threshold annotations:
  ```
  scorecard:
    entityAnnotations:
      thresholds:
        enabled: false
  ```
  Custom entity threshold annotations are not applied, using default number thresholds.

**5. Enable threshold annotations**
  ```
  scorecard:
    entityAnnotations:
      thresholds:
        enabled: true
  ```
  Custom entity threshold annotations are applied.

**6. Threshold annotations override custom app-config annotations**
  ```
  scorecard
    metricProviders:
      github:
        openPRs:
          schedule:
            frequency: { minutes: 5 }
            timeout: { minutes: 10 }
            initialDelay: { seconds: 10 }
          thresholds:
            rules:
              - key: low
                expression: '<10'
                icon: star
                color: '#B4D3B2'
              - key: warning
                expression: '10-1000'
              - key: error
                expression: '>1000'
  ```

Comment out threshold annotations to see custom annotations from app-config are applied:
```yaml
    # scorecard.io/github.openPRs.thresholds.rules.warning: '10-25'
    # scorecard.io/github.openPRs.thresholds.rules.error: '>25'
```

**7. Disable threshold annotations for specific metric**
  ```
  scorecard:
    entityAnnotations:
      thresholds:
        enabled: true
        except:
          - github.openPRs
  ```
  Custom entity threshold annotations are not applied, using default number thresholds.


workspaces/scorecard/plugins/scorecard-backend/src/threshold/thresholdAnnotations.ts
```diff
@@ -0,0 +1,53 @@
...
+ * `scorecard.entityAnnotations.thresholds` (`enabled` / `except`).
+ * When unset, annotation overrides are allowed (same default as disabledMetrics).
+ */
+export function areThresholdAnnotationOverridesAllowed(
+  config: Config,
+  metricId: string,
+): boolean {
+  const entityAnnotationsGlobalEnabled = config.getOptionalBoolean(
+    'scorecard.entityAnnotations.enabled',
+  );
```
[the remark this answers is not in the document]
dzemanov: Fixed together with issues raised in https://github.com/redhat-developer/rhdh-plugins/pull/4160#issuecomment-5177708506 in https://github.com/redhat-developer/rhdh-plugins/pull/4160/commits/96e7594bc8ada51167e9867ab3048b4426b28512

workspaces/scorecard/examples/components/github-scorecard-only.yaml
```diff
@@ -5,6 +5,9 @@ metadata:
   annotations:
     github.com/project-slug: redhat-developer/rhdh-plugins
     backstage.io/source-location: url:https://github.com/redhat-developer/rhdh-plugins
+    # scorecard.io/disabled-metrics: github.openPRs
+    # scorecard.io/github.openPRs.thresholds.rules.warning: '10-25'
+    # scorecard.io/github.openPRs.thresholds.rules.error: '>25'
```
imykhno, line 12: Did you mean to keep these commented lines? If so, please add a brief note explaining when they should be uncommented

workspaces/scorecard/plugins/scorecard-backend/docs/thresholds.md
imykhno, line 1: I think article “Threshold Priority Order” is stale in this file, I would propose updated the article similar as `workspaces/scorecard/AGENTS.md`

workspaces/scorecard/plugins/scorecard-backend/docs/disabled-metrics-logic.md
```diff
@@ -1,22 +1,28 @@
...
+
+| `scorecard.disabledMetrics` includes `metricId` | `entityAnnotations.enabled` | `entityAnnotations.disabledMetrics.enabled` | `entityAnnotations.disabledMetrics.except`  | `scorecard.io/disabled-metrics` entity annotation has `metricId`? | Metric check skipped for `metricId` (not run) |
+| ----------------------------------------------- | --------------------------- | ------------------------------------------- | ------------------------------------------- | ----------------------------------------------------------------- | --------------------------------------------- |
+| Yes                                             | —                           | —                                           | —                                           | —                                                                 | **Yes**                                       |
+| No                                              | `false`                     | ignored                                     | ignored                                     | Yes                                                               | **No** (all annotations ignored)              |
```
imykhno: I noticed that for some columns used `-` for others `ignored`, do "values" have different meaning?

imykhno (changes requested):
I would like to ask are we planning to implement rule-level `enforced`? It was one of AC **"Enforced rules ignore overrides"**. 
Flag on each threshold rule may be following:
``` yaml 
scorecard:
  metricProviders:
    dependabot:
      alertsCritical:
        thresholds:
          rules:
            - key: success
              expression: '<1'
              # enforced omitted → customizable (backward compatible)
            - key: warning
              expression: '1-7'
            - key: error
              expression: '>7'
              enforced: true
```

imykhno:
**Quick follow-up**: did we discuss showing locked thresholds in the UI? It would help users understand why attribute customization isn't working.

dzemanov:
> I would like to ask are we planning to implement rule-level `enforced`? It was one of AC **"Enforced rules ignore overrides"**.

Thank you for raising this question, I will ask PM if it is required also on individual rule level and if not remove it from AC, otherwise create a separate task for it.

The original Feature request was about forcing metric checks to always run (was done as part of OpenSSF Scorecard https://redhat.atlassian.net/browse/RHIDP-12183, so this is following the same syntax) and to prevent users from overriding thresholds for checks via entity annotations.

Personally I think having the option to set each individual rule as enforced is unnecessary and enforcing all rules per check as a whole is enough, since in most cases, you would need to anyway override all rules via entity annotations to cover the whole real line (e.g moving low from `<1` to `<4` will force you to also override middle rule from `1-7` to e.g. `4-7`). This makes locking per rule  less useful and potentially confusing. However, thank you for catching this, we should double-check the granularity level.

workspaces/scorecard/examples/components/github-scorecard-only.yaml
dzemanov: Yes, wanted to have them already ready in examples, as I always search in docs for the right syntax for these.
Added comments

dzemanov:
> **Quick follow-up**: did we discuss showing locked thresholds in the UI? It would help users understand why attribute customization isn't working.

We did not. Let me get back on this with answer about the granularity

workspaces/scorecard/plugins/scorecard-backend/docs/thresholds.md
dzemanov: Added link to disabling section in the same file, thank you!

workspaces/scorecard/plugins/scorecard-backend/docs/disabled-metrics-logic.md
dzemanov: Fixed, `-` is only needed, thank you

imykhno:
I was testing changes locally:

- **Global disable**: works as expected ✅ 
- **Global enable for disabled metrics annotations**: works as expected ✅ 
- **Disable threshold annotations**: works as expected ✅ 
- **Enable threshold annotations**: works as expected ✅ 
- **Threshold annotations override custom app-config annotations**: works as expected ✅ 
- **Disable threshold annotations for specific metric**: works as expected ✅ 
- **Enable threshold annotations for specific metric**: doesn't work as expected ❌
   - If I set `thresholds.enabled: false` with `thresholds.except: - github.openPRs`, customization should be off globally, except for `github.openPRs`. Is that right?

dzemanov:
@imykhno thank you for testing!

> Enable threshold annotations for specific metric: doesn't work as expected ❌
> If I set thresholds.enabled: false with thresholds.except: - github.openPRs, customization should be off globally, except for github.openPRs. Is that right?

You're correct - with `thresholds.enabled: false`, the `except` key isn't honored. This is by design, as `disabledMetrics` also do not honor it, as documented in `config.d.ts` [here](https://github.com/dzemanov/rhdh-plugins/blob/545967445413ec767a6ffec5257fbc672d82297d/workspaces/scorecard/plugins/scorecard-backend/config.d.ts#L114-#L125) or [official documentation](https://docs.redhat.com/en/documentation/red_hat_developer_hub/1.10/html-single/evaluate_project_health_using_scorecards/index#disable-scorecard-metrics-globally_install-and-configure-scorecards-to-view-metrics). Initially, I suggested that the `except` key should work as you described, but it was decided not to pursue that approach. I still think it makes sense to implement it, created bug for it: https://redhat.atlassian.net/browse/RHDHBUGS-3639

imykhno (approved):
Thank you! 
/lgtm

workspaces/scorecard/plugins/scorecard-backend/src/threshold/thresholdAnnotations.ts
```diff
@@ -0,0 +1,53 @@
...
+ * Honors `scorecard.entityAnnotations.enabled` (global) and
+ * `scorecard.entityAnnotations.thresholds` (`enabled` / `except`).
+ * When unset, annotation overrides are allowed (same default as disabledMetrics).
+ */
+export function areThresholdAnnotationOverridesAllowed(
+  config: Config,
+  metricId: string,
+): boolean {
+  const entityAnnotationsGlobalEnabled = config.getOptionalBoolean(
+    'scorecard.entityAnnotations.enabled',
```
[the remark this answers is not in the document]
dzemanov: Since this is only 1 line of reading config with low priority for extraction, I will address it in some following PR

PR #4210, opened 2026-08-09
feat(intelligent-assistant): add screenshot capture utility with sensitive data redaction

its-mitesh-kumar:
## Description

Adds an on-demand screenshot capture utility for the intelligent-assistant plugin's deep context awareness feature (RHIDP-14318). The implementation:

- Uses `html2canvas-pro` to capture the current RHDH page as JPEG
- Includes a sensitive data redaction layer that sanitizes passwords, tokens, and secrets from the cloned DOM before rendering
- Applies performance guardrails: tab visibility check, idle scheduling, configurable timeout (default 1s), and immediate canvas memory release via `try/finally`
- Excludes the chatbot panel and user-specified elements from capture
- Dynamically imports `html2canvas-pro` to avoid blocking initial page load

## Fixed
- [RHIDP-14318](https://issues.redhat.com/browse/RHIDP-14318)

## Screenshots example

https://github.com/user-attachments/assets/5b8eb6e8-e817-4a33-b0e7-82a235fe4359

screenshot-1786298094918
screenshot-1786298176913
screenshot-1786298316493

## Steps to Test
1. Add the following to `app-config.yaml`:
   ```yaml
   intelligent-assistant:
     screen-context:
       enabled: true
       screenshots:
         enabled: true
   ```
2. Cherry-pick the test helper commit to enable screenshot download on every message send:

```bash
git cherry-pick 97e764fa1
```

> This commit ([97e764fa1](https://github.com/redhat-developer/rhdh-plugins/commit/97e764fa159cfab38323bc221f2390466e59b9db) from branch `test/screen-capture-helper`) adds temporary code that triggers `captureScreenshot()` on send and auto-downloads the result. **Revert it before merging.**

Then:
1. Run `yarn start` from the workspace root
2. Open the Intelligent Assistant chat panel
3. Send any message — a screenshot should auto-download
4. Verify:
   - The downloaded image is JPEG format
   - The chatbot panel itself is excluded from the screenshot
   - Any passwords/tokens visible on the page are redacted in the screenshot
   - Console shows capture metrics (format, size, dimensions, time)

### Verifying Sensitive Data Redaction

1. Navigate to a page with visible secrets (e.g., a secret details page with "Show secret" toggle revealed)
2. Trigger screenshot capture (send a message)
3. Confirm the downloaded screenshot shows `[REDACTED]` instead of actual secret values

## Changes Summary

| Commit | Description |
|--------|-------------|
| `f42991670` | Core implementation: screenshot capture + sensitive data redaction |
| `eb7d83167` | Dynamic import of `html2canvas-pro` (code-split) |
| `5dec1cc41` | Address review comments (pattern fixes, test token redaction) |
| `e80882177` | Reduce default capture timeout from 3s to 1s |
| `fe9014693` | Release canvas memory immediately after capture via `try/finally` |

## Checklist


HusneShabbir (commented):
/lgtm

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
```diff
@@ -0,0 +1,145 @@
...
+  /gho_[a-zA-Z0-9]{36,}/g,
+  /github_pat_[a-zA-Z0-9_]{22,}/g,
+  /glpat-[a-zA-Z0-9\-_]{20,}/g,
+  /sk-[a-zA-Z0-9]{32,}/g,
+  /AKIA[0-9A-Z]{16}/g,
+  /xoxb-[0-9]+-[A-Za-z0-9]+/g,
+  /xoxp-[0-9]+-[A-Za-z0-9]+/g,
+  /eyJ[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]{10,}\.[A-Za-z0-9_-]*/g,
+  /-----BEGIN[A-Z ]*PRIVATE KEY-----[\s\S]*?-----END[A-Z ]*PRIVATE KEY-----/g,
+  /Bearer\s+[A-Za-z0-9\-._~+/]{20,}=*/g,
```
karthikjeeyar, line 53: See if you can extend this list a bit more with some more patterns like npm tokens, more slack token patterns, maybe Basic auth pattern etc

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
```diff
@@ -0,0 +1,145 @@
...
+ * WARNING: These regexes are stateful (/g). Reset lastIndex before calling
+ * .test() or .exec() directly. Prefer using redactText() instead.
+ */
+export const SECRET_PATTERNS: RegExp[] = [
+  /ghp_[a-zA-Z0-9]{36,}/g,
+  /ghs_[a-zA-Z0-9]{36,}/g,
+  /gho_[a-zA-Z0-9]{36,}/g,
+  /github_pat_[a-zA-Z0-9_]{22,}/g,
+  /glpat-[a-zA-Z0-9\-_]{20,}/g,
+  /sk-[a-zA-Z0-9]{32,}/g,
```
karthikjeeyar:
```suggestion
  /sk-[a-zA-Z0-9-]{32,}/g,
```


OpenAI API key pattern is separated by hyphens, right now if you pass a Open AI key it doesn't redact it.

Image

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
```diff
@@ -0,0 +1,145 @@
...
+ * Used in html2canvas onclone callback.
+ * Mutates the clone — safe because it's not the live DOM.
+ */
+export function sanitizeClonedDom(clonedDoc: Document): void {
+  // 1. Mask all sensitive inputs/textareas (password fields + label-detected)
+  clonedDoc.querySelectorAll('input, textarea').forEach(el => {
+    if (isSensitiveElement(el, clonedDoc)) {
+      (el as HTMLInputElement).value = MASK;
+    }
+  });
```
karthikjeeyar, line 133: This check works well for all the sensitiveElements with password fields, and element with label such as, token, apiKey etc.

I think we should to add an extra layer of check below this, where it loops through all the `input, textarea` and redact the values in it.

For example, if i accidentally pasted my token in search field, it should still redact the token value before sending it to llm.

See this downloaded screenshot:
Image

Since you already have secret patterns, you could simply `redactText(inp.value);` so that only the token part is [Redacted].

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
```diff
@@ -0,0 +1,145 @@
...
+    return true;
+  }
+
+  const ariaLabel = el.getAttribute('aria-label') || '';
+  if (
+    SENSITIVE_LABEL_PATTERN.test(ariaLabel) &&
+    !SAFE_LABEL_EXCLUSIONS.test(ariaLabel)
+  ) {
+    return true;
+  }
```
karthikjeeyar: Instead of checking only for `aria-label` add more attributes to this check. i.e `name, title, placeholder` etc.

karthikjeeyar (changes requested):

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
karthikjeeyar: Add tests for the new and updated patterns.

karthikjeeyar:
> The Lightspeed chat window/panel is excluded from the captured area

@ShiranHi @benwilcock I see the above point in the Acceptance criteria of [RHIDP-14318](https://issues.redhat.com/browse/RHIDP-14318), but still I want to clarify if this is only for overlay mode so that we capture the page behind the chat panel as main context or it even applies to Lightspeed full screen mode?

Excluding fullscreen mode offers little value, as the chat takes up 90% of the screen while our header and sidebar make up the remaining 10%. 

In Full screen mode, what user sees:

screenshot-1786558900850

---
LLM will see only this content area and it may not have much context:
screenshot-1786559033918

---

Edit: @its-mitesh-kumar Clarified that it is only for overlay and docked mode.

its-mitesh-kumar:
> > The Lightspeed chat window/panel is excluded from the captured area
> 
> @ShiranHi @benwilcock I see the above point in the Acceptance criteria of [RHIDP-14318](https://issues.redhat.com/browse/RHIDP-14318), but still I want to clarify if this is only for overlay mode so that we capture the page behind the chat panel as main context or it even applies to Lightspeed full screen mode?

 @karthikjeeyar this deep context feature is only for overlay and docked mode.

karthikjeeyar:
> @karthikjeeyar this deep context feature is only for overlay and docked mode.

Got it, thanks for clarifying 👍🏻 . From the story ticket it wasn't clear but I see Overlay and docked mode in parent ticket's title.

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/validation.ts
```diff
@@ -25,6 +25,9 @@ import {
 import { Attachments, QueryRequestBody } from './types';
 
 const JPEG_MAGIC = [0xff, 0xd8, 0xff];
+const WEBP_RIFF = [0x52, 0x49, 0x46, 0x46]; // "RIFF"
+const WEBP_MARKER = [0x57, 0x45, 0x42, 0x50]; // "WEBP" at offset 8
+const WEBP_MIN_BYTES = 12;
```
karthikjeeyar: You added `WebP` support to the intelligent-assistant backend which lets RHDH backend to accept `WebP` images. However, as far as I know, the Lightspeed stack doesn't support `WebP`. @JslYoon  @Jdubrick please correct me if I'm wrong.

Has WebP support been discussed with the lightspeed-core or AI team for a follow-up PR? If not, the backend `WebP` validation changes should be removed and the capture utility should use `JPEG` as  the primary format instead.

karthikjeeyar (commented):
The [original PoC ](https://github.com/redhat-developer/rhdh-plugins/compare/main...thepetk:rhdh-plugins:deep-context-retrieval-poc) by @thepetk for this screenshot feature used `snapdom`  library, which seems like the better choice here , it’s smaller, faster, has zero dependencies, and was already validated. I am curious, what was the reason for switching to `html2canvas-pro`?


Did some quick comparison: 

| | [`@zumer/snapdom`](https://github.com/zumerlab/snapdom) ([npm](https://www.npmjs.com/package/@zumer/snapdom)) | [`html2canvas-pro`](https://github.com/yorickshan/html2canvas-pro) ([npm](https://www.npmjs.com/package/html2canvas-pro)) |
|--|--|--|
| **GitHub stars** | 7,992 | 545 |
| **Package size** | 168 KB | 5.17 MB |
| **Dependencies** | 0 | 2 |
| **Performance** | 0.4–364 ms (simple to complex pages) | 68–1,380 ms  (simple to complex pages) |
| **Open issues** | 0 | 2 |
| **Output formats** | SVG, PNG, JPG, WebP, Canvas, Blob | Canvas  (PNG, JPEG, WebP via `toDataURL`) |
| **Validated in PoC** | Yes | No |

Snapdom self reported performance benchmarks can be found here - https://www.npmjs.com/package/@mindblowing/snapdom?activeTab=readme

>[!NOTE]
>Benchmarks compare snapdom vs the original `html2canvas`; `html2canvas-pro` is a fork with the same rendering approach.

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/validation.ts
JslYoon: Yes, as of right now, we only accept JPEG image types. `WebP` I think would be nice to have, but it would take more cycles as it requires upstream changes in lightspeed.

JslYoon: I think `WebP` is something that we could definitely use, but since I'm not certain on when this might go upstream, what if we split this PR into two, so we merge in `JPEG` bit first, then merge in `WebP` later once upstream is ready? wdyt @karthikjeeyar @Jdubrick

its-mitesh-kumar: @JslYoon Lets try to  get `webp` supported at upstream  in 2.1 itself, it 30% more optimized than `JPEG`. I am reverting the Webp first approach for now, once we support for it in upstream can have another PR for webp first approch.

its-mitesh-kumar: @karthikjeeyar I tried snapdom — the end-to-end time for producing the final screenshot is ~2x slower than html2canvas-pro (~761ms vs ~364ms).

Then I was just checking why ?

The benchmarks on the [snapdom README](https://www.npmjs.com/package/@mindblowing/snapdom?activeTab=readme) only measure the **capture step** (`snapdom(el)` — DOM clone + SVG serialization). They don't include the **export step** (`.toWebp()` / `.toJpg()`), which involves rasterizing the SVG to a Canvas and then encoding it to WebP/JPEG. That extra rasterization step is where the additional time goes.

html2canvas-pro outputs a Canvas directly, so encoding to WebP/JPEG is a single native `canvas.toDataURL()` call. snapdom has to go through SVG → Image decode → Canvas draw → `toDataURL()`, which roughly doubles the total time.


https://github.com/user-attachments/assets/2fd90897-e10c-4d1e-ba27-3d0a9452fcf8

karthikjeeyar: @its-mitesh-kumar Thats a fair point and thanks for benchmarking this end-to-end. Given the <500 ms AC target, let us continue with `html2canvas-pro` library. 

For the `WebP` type, I agree this will be a good improvement for the future. lets us kick it off with Jpeg first support for now.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
its-mitesh-kumar: @karthikjeeyar The existing token regex patterns are sufficient for all plugins in rhdh-plugins, backstage/backstage, and backstage/community-plugins — I don't see any UI in these repos that handles or displays npm tokens, additional Slack token formats, Google API keys, Azure AD secrets, etc.

Since the screenshots are not stored and are only used for a single query context, adding defensive patterns (npm, Slack app/refresh, Google AIza, Azure 7Q~/8Q~, DB connection strings, Slack webhooks, generic key=value, etc.) would grow the list from 12 to ~25 regex patterns, and each one runs against every text node in the cloned DOM.

That said, if you feel we should add them defensively (e.g. to protect against users pasting arbitrary secrets into chat, or for custom plugins that might handle these token types), I'm happy to add them. Same for the input field value redaction — it would add an extra pass over all `input`/`textarea` elements. Let me know your preference.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
its-mitesh-kumar: Please see the comment https://github.com/redhat-developer/rhdh-plugins/pull/4210#discussion_r3774227152

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/sensitive-data-redactor.ts
karthikjeeyar: I understand keeping the scope reasonable, but consider a user working with the Orchestrator workflow form or a Scaffolder template, these let users design custom forms with text inputs. If a user accidentally pastes a token into one of those fields and then asks Lightspeed a question, that token ends up in the screenshot sent to the LLM, which defeats the purpose of sensitive data redaction. 

We can't scope redaction to "what RHDH plugins currently display" because we don't control what users type into those fields. This is a security concern, so the cost is a few extra regex patterns check is minimal.

I was not asking to cover all the available token formats, it would be good to add most common formats.


 For the   input field value redaction, it is a bug that you can reproduce, pasting a GitHub PAT into the RHDH search bar shows it unredacted in the screenshot. Also It's not an extra pass if you merge it with the existing querySelectorAll('input, textarea') loop.

karthikjeeyar: Your current code already has two separate Slack token patterns:

```  
/xoxb-[0-9]+-[A-Za-z0-9]+/g,
/xoxp-[0-9]+-[A-Za-z0-9]+/g,
```

you could consolidate them into a single regex that covers all 5 Slack token prefixes (xoxa-, xoxb-, xoxr-, xoxs-, xoxp-):

```
/xox[abrsp]-[0-9]+-[A-Za-z0-9]+/g,
```
Reduces the pattern length by 1 but covers all 5 slack patterns 😉

its-mitesh-kumar:
@karthikjeeyar I've addressed all your comments:
1. **Removed WebP** — JPEG only now; backend fully reverted to `main` (zero diff)
2. **Hardened redaction** — fixed `sk-` regex (hyphens), expanded `isSensitiveElement` to check `name`/`title`/`placeholder`, and regex-scan all input values for pasted tokens
3. **Expanded `SECRET_PATTERNS`** — consolidated GitHub/Slack regexes, added npm, Google API, Azure AD, Slack webhooks, Basic auth in URIs
4. **Added comprehensive tests** for all new patterns with real-world token formats
5. **Updated changeset** — frontend `minor` only, removed backend bump

Please have a look.


https://github.com/user-attachments/assets/2ae2bfc0-95d0-406a-955c-ffff060eb694


https://github.com/user-attachments/assets/8c154245-8f1b-4752-8bb5-ac7cad92193f

its-mitesh-kumar:
Also reduced `DEFAULT_TIMEOUT_MS` from 3s → 1s to tighten the performance guardrail — if capture exceeds 1s, it gracefully times out instead of blocking the user experience.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/screen-capture.ts
```diff
@@ -0,0 +1,195 @@
...
+
+  const { default: html2canvas } = await import('html2canvas-pro');
+
+  const start = window.performance.now();
+
+  const canvas = await html2canvas(targetElement as HTMLElement, {
+    useCORS: true,
+    allowTaint: false,
+    logging: false,
+    onclone: (clonedDocument: Document) => {
```
karthikjeeyar, line 110: Please wrap this in a try/finally block. In the finally block, set `canvas.width = 0` and `canvas.height = 0` to ensure the canvas memory is released immediately after taking the screenshot.

its-mitesh-kumar: Done

its-mitesh-kumar: I placed the try/finally after the html2canvas() call rather than wrapping it:
**Reasoning:** if html2canvas throws, no canvas exists to clean up — so the narrower scope avoids a null guard and makes the resource ownership clearer. Also handles the scaled canvas (separate buffer when downscaling is needed) to avoid leaking  in that path.

workspaces/intelligent-assistant/plugins/intelligent-assistant/src/utils/screen-capture.ts
```diff
@@ -0,0 +1,205 @@
...
+const DEFAULT_EXCLUDE_SELECTOR = '[data-screen-capture-exclude]';
+
+export interface CaptureOptions {
+  /** Image quality 0-1 (default: 0.7) */
+  quality?: number;
+  /** CSS selector for elements to exclude (default: '[data-screen-capture-exclude]') */
+  excludeSelector?: string;
+  /** Max output width in pixels for compression (default: 1280) */
+  maxWidth?: number;
+  /** Maximum time allowed for capture in ms (default: 3000) */
```
karthikjeeyar: nit: With your recent changes it should be `1000`.

```suggestion
  /** Maximum time allowed for capture in ms (default: 1000) */
```

karthikjeeyar (approved):
/approve
/lgtm

PR #4216, opened 2026-08-10
fix(intelligent-assistant): chat UI squashed when opening drawer in overlay/docked modes

rohitratannagar:
#### fixes: https://redhat.atlassian.net/browse/RHDHBUGS-3575 
- Fixes the chat history drawer rendering inline in overlay and docked display modes, which squashed the chat content area to ~55% of its width (overlay) or hid it entirely (docked)
- Root cause: `ChatbotConversationHistoryNav` was hardcoded to `displayMode={ChatbotDisplayMode.embedded}`, which tells PatternFly to render the drawer panel **inline** (side-by-side). In overlay/docked modes it should use the actual `displayMode` so the drawer renders as a floating **overlay** with a backdrop instead
- One-line fix: pass `displayMode={displayMode}` instead of the hardcoded value

## Before / After

| Mode | Before (bug) | After (fix) |
|------|-------------|-------------|
| **Overlay** | Drawer opens inline, squashing footer from 480px → 264px. Suggestion cards and input text truncated | Drawer overlays with backdrop. Chat content stays full width (480px) |
| **Docked** | Drawer replaces chat content entirely — no chat visible | Drawer overlays with backdrop. Chat content visible behind |

## Test plan

- [x] Reproduction test confirms footer width drops to 55% when drawer opens (fails before fix, passes after)
- [x] All 980 unit tests pass (73 suites, zero failures)
- [x] Before/after video recordings captured
- [ ] Manual verification: open chatbot in overlay mode → click hamburger → drawer should float over content
- [ ] Manual verification: switch to docked mode → click hamburger → drawer should float over content
- [ ] Manual verification: fullscreen mode → click expand history → drawer should still render inline (unchanged behavior)

Resolves: [RHDHBUGS-3575](https://redhat.atlassian.net/browse/RHDHBUGS-3575)

### DIFF
--- a/workspaces/intelligent-assistant/.changeset/fix-drawer-squash-overlay-docked.md
+++ b/workspaces/intelligent-assistant/.changeset/fix-drawer-squash-overlay-docked.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-intelligent-assistant': patch
+---
+
+Fixed chat UI being squashed when opening the chat history drawer in overlay and docked display modes. The drawer now correctly overlays the content instead of rendering inline.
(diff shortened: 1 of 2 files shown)


rohitratannagar:
Closing this in the favor of https://github.com/redhat-developer/rhdh-plugins/pull/4093#issuecomment-5257160088

PR #4234, opened 2026-08-10
fix(theme): wrap toggle button items to prevent clipped overflow

logonoff:

before:
image

after:
image

### DIFF
--- a/workspaces/theme/.changeset/polite-eggs-care.md
+++ b/workspaces/theme/.changeset/polite-eggs-care.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+Wrap toggle button items to prevent clipped overflow
--- a/workspaces/theme/.changeset/polite-eggs-care.md
+++ b/workspaces/theme/.changeset/polite-eggs-care.md
@@ -2,4 +2,6 @@
 '@red-hat-developer-hub/backstage-plugin-theme': patch
 ---
 
-Wrap toggle button items to prevent clipped overflow
+Wrap toggle button items to prevent clipped overflow (RHDHBUGS-3622)
+
+This is a global `MuiToggleButtonGroup` theme override — it applies to every `ToggleButtonGroup` in RHDH, including those in dynamic plugins. Groups with two items will not wrap in practice, but consumers with wider groups should be aware.
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -273,6 +273,14 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         },
       },
     };
+    components.MuiToggleButtonGroup = {
+      styleOverrides: {
+        root: {
+          flexWrap: 'wrap',
+          rowGap: '8px',
+        },
+      },
+    };
     components.MuiIconButton = {
       styleOverrides: {
         root: {
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -292,6 +292,14 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         },
       },
     };
+    components.MuiToggleButtonGroup = {
+      styleOverrides: {
+        root: {
+          flexWrap: 'wrap',
+          rowGap: '8px',
+        },
+      },
+    };
     components.MuiIconButton = {
       styleOverrides: {
         root: {
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.ts
@@ -297,6 +297,7 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         root: {
           flexWrap: 'wrap',
           rowGap: '8px',
+          maxWidth: '100%',
         },
       },
     };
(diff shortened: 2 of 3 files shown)


workspaces/theme/plugins/theme/src/utils/createComponents.ts
```diff
@@ -273,6 +273,14 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         },
       },
     };
+    components.MuiToggleButtonGroup = {
+      styleOverrides: {
+        root: {
+          flexWrap: 'wrap',
+          rowGap: '8px',
```
ciiay, line 299: Wrapped `ToggleButtonGroup` rows still use MUI grouped styles (`:not(:first-of-type)` removes the start border / uses `-1px` margin). Can you confirm the first button on the second row still looks correct with a docked Quickstart/IA drawer?

**Update:** Please ignore the earlier suggestion to reset grouped borders or use `gap`. That splits the control into detached buttons and is against our UX — theme toggles should stay one attached segmented control.

`maxWidth: '100%'` and the `flexWrap: 'wrap'` regression test look good.

workspaces/theme/.changeset/polite-eggs-care.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+Wrap toggle button items to prevent clipped overflow
```
ciiay: Please mention this is a global `MuiToggleButtonGroup` override (not Settings-only) and link RHDHBUGS-3622 so release notes are accurate.

This will affect every MUI `ToggleButtonGroup` in RHDH, including dynamic plugins (e.g. augment RAG/MCP toggles, scorecard layout switcher). Most of those are 2-item groups and will not wrap, which is fine — worth a brief callout so consumers aren't surprised.

ciiay (commented):
## Review

Approach looks right for [RHDHBUGS-3622](https://redhat.atlassian.net/browse/RHDHBUGS-3622). The Settings appearance picker is Backstage `UserSettingsThemeToggle` (`@material-ui/lab/ToggleButtonGroup`), so we can't patch it from this repo; a theme override is the practical fix. Wrapping is one of the two behaviors the ticket asked for.

CI is green. Copilot's `rowGap` unit comment is already addressed in the force-push (`'8px'`).

Not blocking, but please see the inline notes on grouped borders when wrapping, the global behavior in the changeset, and adding a small unit test.

Also: this PR was marked stale. This review should keep it open.

### Out of scope / FYI
Boost BUI `ToggleButtonGroup` is unaffected. Placement next to `MuiToggleButton` inside `if (options.buttons !== 'mui')` is consistent. UnifiedTheme maps `components.MuiToggleButtonGroup` onto MUI v4 `overrides`, so the v4 Settings picker should pick this up — the screenshots support that.

logonoff:
Addressed both review comments in f66f4302b:

- **`createComponents.ts`**: Added `maxWidth: '100%'` to the `MuiToggleButtonGroup` root override for reliable `inline-flex` wrapping. Also added a regression test asserting `flexWrap: 'wrap'` is set on `MuiToggleButtonGroup.styleOverrides.root`.
  On the grouped border question: MUI's `:not(:first-of-type)` selector operates on DOM order rather than row position, so the first button on a second wrapped row will have its left border removed and `-1px` margin applied by MUI's grouped styles. If you confirm it looks broken in context (e.g. with a docked drawer), I can add a CSS reset (`& > :first-of-type { margin-left: 0; border-left: 1px solid; }`) or switch to `gap`.

- **Changeset**: Updated to reference RHDHBUGS-3622, note this is a global `MuiToggleButtonGroup` override (not Settings-only), and include the brief callout about 2-item groups not wrapping in practice.

\- claude

workspaces/theme/plugins/theme/src/utils/createComponents.ts
ciiay: **Retracted.** Please do not land the `gap` + `border-left` snippet previously in this comment.

That approach detaches every option into a separate button, which is against our UX. Theme toggles should stay one attached segmented control (no gutters), matching RHDHBUGS-3622’s clip-fix (wrap or scrollbar) without changing the grouped look.

Keep wrap + `rowGap` + `maxWidth` only.

logonoff: All items now have a border-left and a gap which doesn't look right..

image

Some quick googling reveals that selecting the first item of a flex row is impossible in pure CSS..?

https://stackoverflow.com/a/55088805

logonoff: As far as I'm aware the closest that we can get to is turning ToggleButtonGroup and ToggleButton into separate, distinct buttons:

image

The issue may lie upstream, I can reproduce this in the backstage demo site:

image

The core issue is that extending the theme selector to have 7 options isn't really an ideal use of ToggleGroup, you would want to se a select for that. The solution might be to get them to convert this MUI ToggleGroup into a BUI Select.

ciiay:
Hi @ShiranHi , do you think the theme options wrapping in two rows in PR description look good to you?

ShiranHi:
> Hi @ShiranHi , do you think the theme options wrapping in two rows in PR description look good to you?

LGTM, thank you!

ciiay (approved):
/lgtm

Thanks for the PR 💯

workspaces/theme/plugins/theme/src/utils/createComponents.ts
```diff
@@ -273,6 +273,14 @@ export const createComponents = (themeConfig: ThemeConfig): Components => {
         },
       },
     };
+    components.MuiToggleButtonGroup = {
+      styleOverrides: {
+        root: {
+          flexWrap: 'wrap',
+          rowGap: 8,
```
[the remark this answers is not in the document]
ciiay: Hi @logonoff , if this has been resolved you can mark it so we can get this PR in 🙌

PR #4117, opened 2026-07-31
feat(orchestrator): add 5 Orchestrator MCP actions (RHIDP-14041)

jordigilh:
## Summary

Adds 5 MCP (Model Context Protocol) actions to the Orchestrator backend plugin so LLM/CLI clients can discover and run Orchestrator workflows, per [RHIDP-14041](https://issues.redhat.com/browse/RHIDP-14041) and its sub-tasks:

- `orchestrator.list-workflows` ([RHIDP-14044](https://issues.redhat.com/browse/RHIDP-14044)) - list workflow definitions, optionally filtered by name/status
- `orchestrator.get-workflow-schema` ([RHIDP-14045](https://issues.redhat.com/browse/RHIDP-14045)) - fetch a workflow's input JSON schema (prerequisite for execute-workflow)
- `orchestrator.execute-workflow` ([RHIDP-14046](https://issues.redhat.com/browse/RHIDP-14046)) - start a workflow run, validating inputs against its own schema (ajv) first
- `orchestrator.list-instances` ([RHIDP-14047](https://issues.redhat.com/browse/RHIDP-14047)) - list workflow runs, optionally filtered by status
- `orchestrator.get-instance` ([RHIDP-14048](https://issues.redhat.com/browse/RHIDP-14048)) - fetch a single workflow run's status/timestamps/output

Architecture follows the precedent established by Scorecard's own MCP actions (#3332): each action calls the internal `OrchestratorService` directly, enforces the **existing** Orchestrator RBAC permissions with full fidelity (conditional policies, deprecated per-workflow fallback, instance-ownership checks) - **no new permissions are introduced** - and surfaces failures as structured `NotFoundError`/`InputError`/`NotAllowedError`/`ServiceUnavailableError` types from `@backstage/errors`.

### Key changes

- `service/workflowAuthorization.ts` (new): credentials-based authorization helpers extracted from `router.ts` so the HTTP router and the new MCP actions share one RBAC implementation. `router.ts` keeps thin, request-scoped wrappers; `router.test.ts` is unchanged and stays green as the behavior-preservation safety net.
- `service/resolveWorkflowDefinition.ts` (new): shared workflow info/serviceUrl/definition resolution (with 404 handling) used by `get-workflow-schema` and `execute-workflow`.
- `actions/*.ts` (new): the 5 actions, each with its own unit test file tracing directly to its story's acceptance criteria.
[...]

PR #3664, opened 2026-07-02
test(scorecard): add MCP integration coverage for scorecard tools

HusneShabbir:
## Summary
Adds MCP integration coverage for Scorecard backend tools with a reusable test helper.

**Resolves:**
https://redhat.atlassian.net/browse/RHIDP-14592

### `scorecard.list-metrics`
- Tool is discoverable via `tools/list`
- Tool metadata is validated (`readOnlyHint`, `destructiveHint` shape)
- Returns configured metrics on success
- Applies conditional permission filtering (returns only allowed metric IDs)
- Not exposed when `scorecard` is excluded from `backend.actions.pluginSources`

### `scorecard.get-entity-metrics`
- Returns metrics for a valid existing entity
- Returns error when entity is not found
- Returns error when entity access is denied
- Returns validation error when `entityRef` is missing
- Returns validation error when `entityRef` type is invalid

## Test Infrastructure
- Introduced reusable MCP integration helpers in `packages/backend/src/__fixtures__/mcpTestUtils.ts`
- Added backend script: `test:integration`

## Verification
- `yarn --cwd workspaces/scorecard/packages/backend run test src/mcp-tools.integration.test.ts --watch=false`
- Result: **10/10 tests passing**

PR #4273, opened 2026-08-12
test(orchestrator): expand MCP integration coverage for orchestrator

HusneShabbir:
## Summary

Expands Orchestrator MCP integration coverage to match the Scorecard QE pattern (#3664), on top of the wiring tests from #4117.

**Coverage**

- **Discovery / metadata** — all 5 tools via `tools/list`; `readOnlyHint` / `destructiveHint` (incl. `execute-workflow` destructive)
- **`pluginSources` gating** — tools hidden when `orchestrator` is excluded
- **`list-workflows`** — happy path; conditional RBAC → only allowed workflow IDs; access denied
- **`get-workflow-schema`** — happy path; missing workflow; missing/invalid `workflowId`
- **`execute-workflow`** — happy path; Ajv validation failure; access denied; missing `workflowId`
- **`list-instances` / `get-instance`** — happy paths; missing instance; missing/invalid `instanceId`

**Resolves:**
https://redhat.atlassian.net/browse/RHIDP-15230

## Test plan

- [x] `yarn --cwd workspaces/orchestrator/plugins/orchestrator-backend test:integration`
- [x] Result: **19/19 passing**

### DIFF
--- a/workspaces/orchestrator/plugins/orchestrator-backend/src/__testUtils__/mcpTestUtils.ts
+++ b/workspaces/orchestrator/plugins/orchestrator-backend/src/__testUtils__/mcpTestUtils.ts
@@ -28,20 +28,13 @@ import {
 import { Client } from '@modelcontextprotocol/sdk/client/index.js';
 import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
 
+import { ORCHESTRATOR_WORKFLOW_RESOURCE_TYPE } from '@red-hat-developer-hub/backstage-plugin-orchestrator-common';
+
 import type { Server } from 'node:http';
 
-// Must be imported after the `jest.mock` calls above so that `orchestratorPlugin`'s
-// transitive imports resolve to the mocked service classes.
-// eslint-disable-next-line import/first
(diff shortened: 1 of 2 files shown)


gabemontero:
eventually @HusneShabbir , once we get a sign off from @lholmquist @lokanandaprabhu  from the plugins / orchestrator team, we have the  change https://github.com/redhat-developer/rhdh-plugins/pull/4117 that @jordigilh and @yangcao77 have been collaborating on that I suspect we would want coverage here for as well (but I'll defer to @yangcao77 for the final say on that )

lholmquist (commented):
I think you might need to add a changeset here

HusneShabbir:
@lholmquist This PR is test-only (integration coverage under __testUtils__ / *.integration.test.ts), so no published package surface changed. Per the Backstage changeset guidance, test-only changes don't need a changeset — same approach as #3664. Happy to add an empty one if you prefer, but I don't think it's required here.

lholmquist:
right.  Looks like the github action reported it because it didn't think that MCPTestUtils was a test file 🤷 .  LGTM

lholmquist (approved):

PR #4309, opened 2026-08-14
fix(theme): prevent app remount on theme switch

karthikjeeyar:

Fixes: [RHDHBUGS-3650](https://redhat.atlassian.net/browse/RHDHBUGS-3650)

 Switching themes in RHDH caused a full React subtree remount instead of a re-render, losing all component state and
  re-firing all API calls

**Root Cause:**
  Each theme previously had its own Provider function created by createThemeProviderForThemeName(). React's reconciliation treats different function references as different component types, triggering a full unmount/remount of the entire tree below the provider on every theme switch.

**Fix:**

 All built-in themes now share a single Provider component via `createSharedThemeProvider`, so React re-renders in place instead of remounting

## Screenshots

https://github.com/user-attachments/assets/f96f1c7c-85f3-44e3-86d0-0040e29bfe8f

# Verification: 

When you switch between themes: 
  - No extra API calls in network tab.
  - Elements tab, shows that components are not re-mounted.

### DIFF
--- a/workspaces/theme/.changeset/shared-theme-provider.md
+++ b/workspaces/theme/.changeset/shared-theme-provider.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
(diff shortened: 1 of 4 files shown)


debsmita1 (approved):
tested this on rhdh-local


/lgtm

PR #4342, opened 2026-08-17
fix(orchestrator): center empty state content on wide viewports

lokanandaprabhu:
## Summary

- Fix misaligned Orchestrator empty states ("No workflows added yet" / "No runs yet") on wide viewports
- Replace CSS selectors targeting `BackstageEmptyState-*` class names (no longer present in newer Backstage/JSS builds) with stable `MuiGrid` selectors
- Vertically center empty-state text and illustration for both Workflows and All runs tabs

## Before

On wide viewports, empty-state text sits top-left and the illustration is pushed to the side — the CSS override never applied because hashed JSS classes replaced the old `BackstageEmptyState-root` selector.

Screenshot 2026-08-04 at 11 55 02 AM

## After

Text and illustration are vertically centered together using `& > [class*="MuiGrid-container"]` with `align-items: center`.

Screenshot 2026-08-04 at 12 27 41 PM

## Test plan

- [ ] Open Orchestrator → Workflows tab with no workflows — empty state text and illustration are vertically aligned
- [ ] Open Orchestrator → All runs tab with no runs — same alignment
- [ ] Resize to narrow viewport — responsive empty-state layout still stacks correctly with illustration on top

### DIFF
--- a/workspaces/orchestrator/.changeset/fix-empty-state-alignment.md
+++ b/workspaces/orchestrator/.changeset/fix-empty-state-alignment.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Fix Orchestrator empty state layout on wide viewports by targeting MuiGrid selectors instead of deprecated BackstageEmptyState class names.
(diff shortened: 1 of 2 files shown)


karthikjeeyar (approved):
/approve
/lgtm

PR #4344, opened 2026-08-17
fix(global-header): resolve accessibility violations in nested interactive controls

rohitratannagar:
## Description

Fixes accessibility violations in the Global Header plugin, primarily the `nested-interactive` axe-core rule violation (WCAG 2.1 AA). Search result items and starred dropdown items had interactive elements (links) nested inside other interactive elements (buttons/links). Additionally fixes missing accessible names, improper `aria-expanded` values, and a decorative SVG missing `aria-hidden`.

**Changes:**
- **SearchResultItem/SearchOption**: Merged `Link` wrapper into `ListItem` via `component={Link}` prop to eliminate nested interactive controls in search results
- **StarredDropdown**: Replaced `MenuItem component={Link}` containing `IconButton` with `ListItem` + `secondaryAction` + `ListItemButton component={Link}` pattern to separate navigation from star/unstar action
- **HeaderDropdownComponent**: Fixed `aria-expanded` to use `Boolean(anchorEl)` instead of `true | undefined`
- **DefaultLogo**: Added `aria-hidden="true"` to decorative SVG (parent link already has `aria-label="Home"`)
- **ProfileDropdown**: Added accessible `tooltip` prop to `GlobalHeaderDropdown` for the Settings button
- **SearchBar**: Pass i18n-safe `noResultsText` prop to avoid hardcoded string comparison
- **E2e tests**: Re-enabled `nested-interactive` rule in axe scans (scoped to `#global-header`), updated aria snapshots and selectors to match new DOM structure

## Fixed
- [RHDHBUGS-2247](https://redhat.atlassian.net/browse/RHDHBUGS-2247) — Accessibility Violation in Global Header – Nested Interactive Controls

## Checklist

### DIFF
--- a/workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchResultItem.tsx
+++ b/workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchResultItem.tsx
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import React from 'react';
 import { Link } from '@backstage/core-components';
 import ListItem from '@mui/material/ListItem';
 import Box from '@mui/material/Box';
@@ -28,6 +29,7 @@ interface SearchResultItemProps {
   query: SearchResultProps['query'];
   result: Result<SearchDocument> | undefined;
   renderProps: any;
+  noResultsText?: string;
 }
 
 export const SearchResultItem = ({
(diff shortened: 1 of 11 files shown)


HusneShabbir (changes requested):
I noticed a couple of deviations in the starred items compared to Main:

1. The star should only be visible when hovering over a list item, but this is currently broken when there’s only a single item in the list.
2. The star appears to overlap the list items.
3. When there are multiple items in the starred list, the first item still shows the star even without hovering.

https://github.com/user-attachments/assets/7073fd82-6bb6-4e86-b306-bde5b1b92e3c

Could you please take a look at these?

HusneShabbir (approved):
/lgtm

workspaces/global-header/plugins/global-header/src/components/SearchComponent/SearchResultItem.tsx
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import React from 'react';
 import { Link } from '@backstage/core-components';
```
its-mitesh-kumar, line 18: Please remove it. Only import the needed one like `MouseEvent`

its-mitesh-kumar (approved):
/lgtm

PR #4349, opened 2026-08-17
fix(bulk-import): addressing security concerns

lholmquist:

This fixes the JIRA https://redhat.atlassian.net/browse/RHIDP-16077 as part of the GA of the bulk-import plugin.

The biggest change is that when using the bulk-import with orchestrator, a github PAT can no longer be used

### Code
* Added GithubApiService.getAppInstallationCredentials() — uses CustomGithubCredentialsProvider and fails closed unless type === 'app'
* Wired orchestrator GitHub path in createWorkflowImportJobs to use it; GitLab still uses getCredentials

### Tests
* Updated execute-orchestrator-workflow mocks/assertions for the App path
* Added fail-closed case when App credentials are unavailable
* Added unit tests for getAppInstallationCredentials (app success, PAT-only, missing token)

### Docs & changeset
* README: App requirement, TTL/mid-run residual risk, GitLab stance, RBAC, deployment checklist

@AndrienkoAleksandr would you mind taking a look at this since I think you originally had added the orchestrator support?


workspaces/bulk-import/plugins/bulk-import-backend/src/github/githubApiService.ts
```diff
@@ -139,6 +139,29 @@ export class GithubApiService implements GitApiService {
...
+   * Returns a short-lived GitHub App installation token for the given repo URL.
+   * Fails closed if only a classic PAT is available or no App installation exists.
+   * Used by orchestrator import mode so long-lived PATs are never forwarded to SonataFlow.
+   */
+  async getAppInstallationCredentials(
+    repoUrl: string,
+  ): Promise<{ token: string }> {
+    const creds = await this.githubCredentialsProvider.getCredentials({
+      url: repoUrl,
+    });
```
PatAKnight: I believe that there is a scenario where we can unintentionally pass the wrong app info if the app is installed for one org but we are attempting to import from another org.

One way that we could potentially avoid this is to update the `getCredentials` to now return the cred type. Then call it here and check if it is app or not and log our error.

So, it would look something like:

```typescript
async getCredentials(repoUrl: string): Promise<{
  token: string;
  type: GithubCredentialType;
}> {
  const provider = DefaultGithubCredentialsProvider.fromIntegrations(
    this.integrations,
  );
  const creds = await provider.getCredentials({ url: repoUrl });
  if (!creds?.token) {
    throw new Error(`Token not configured for 'github' provider`);
  }
  return { token: creds.token, type: creds.type };
}
```

and then `getAppInstallationCredentials` would get updated to

```typescript
async getAppInstallationCredentials(
  repoUrl: string,
): Promise<{ token: string }> {
  const creds = await this.getCredentials(repoUrl);
  if (creds.type !== 'app') {
    throw new Error(
      `Orchestrator import requires a GitHub App installation token for '${repoUrl}'. ` +
        `Configure integrations.github with an App that has access to this repository; ` +
        `classic personal access tokens are not forwarded in orchestrator mode.`,
    );
  }
  return { token: creds.token };
}
```

workspaces/bulk-import/.changeset/orchestrator-github-app-tokens.md
```diff
@@ -0,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-bulk-import-backend': minor
```
PatAKnight: A topic for discussion, but I wonder if we should mark this as `major` since it does technically break things for users.

PatAKnight (commented):
Couple of comments. Also, I think that it might be worth to align the Orchestrator with the Scaffolder import approach mentioned in the remediation `Optional hygiene: use auth.getPluginRequestToken({ targetPluginId: 'orchestrator' }) instead of raw getBearerTokenFromReq (Backstage plugin auth only; does not mitigate SCM exposure)`

lholmquist:
@PatAKnight if you wouldn't mind giving this another look.  I added your suggestions and also added that optional  recommended remediation

PatAKnight (commented):
LGTM, going to hold off on giving approval in case @kim-tsao or someone from security wants to also take a look.

Omar-AlJaljuli:
@kim-tsao this lgtm, everything the ticket outlined as necessary was resolved (there were a few optional criteria but they are OPTIONAL). But for the Gitlab risks, I assume this will also need to be documented in our official documentation, not just on the repo.

kim-tsao (approved):
/lgtm

The PR addresses all remediations (except for GL but that's understandable), so I'm approving. 

As an aside, @PatAKnight @lholmquist, there doesn't seem to be any warning message to indicate that the token expired, so how would users know they need to re-trigger the workflow?  This is not a new problem, but PAT expiration was previously longer than 1 hour, so these types of failures may not have been frequently encountered.  Just wondering if we can add some troubleshooting tips here.

AndrienkoAleksandr (dismissed):
_Requesting changes on the service-token part; the GitHub App token work looks fine._

I tested this pull request.
I may be missing something, but using a plugin service token here looks like it has a couple of side effects. Wanted to check whether this is expected.
Context: after an import, Bulk Import shows a link to the Orchestrator workflow run. Users can click through from the import job to inspect that run. The issues below show up on that path.

1) RBAC disabled. The user can execute the workflow and it completes. Opening the Orchestrator run from that link then shows - Run by: "system", so we lose who actually started the import.

Image

2) RBAC enabled. The user can still execute from Bulk Import (the call goes to Orchestrator as the plugin). Clicking the same Orchestrator workflow link is then blocked by instance ownership, because the initiator is `user:default/system`, not the logged-in user. So the import succeeds, but the user cannot open the run they just started.

Image

Would it still meet the hygiene goal if we mint an orchestrator-scoped plugin token on behalf of the user, instead of forwarding the raw bearer?
```ts
const credentials = await httpAuth.credentials(req);
const { token } = await auth.getPluginRequestToken({
  onBehalfOf: credentials,
  targetPluginId: 'orchestrator',
});
```

@kim-tsao, @PatAKnight, @lholmquist  what do you think?

lholmquist:
@AndrienkoAleksandr 

> RBAC disabled. The user can execute the workflow and it completes. Opening the Orchestrator run from that link then shows - Run by: "system", so we lose who actually started the import.

That does seem incorrect.  I would think it should probably show the user that started the import.  What would that look like if they were previously using a PAT token?  

> RBAC enabled. The user can still execute from Bulk Import (the call goes to Orchestrator as the plugin). Clicking the same Orchestrator workflow link is then blocked by instance ownership, because the initiator is user:default/system, not the logged-in user. So the import succeeds, but the user cannot open the run they just started.

If the answer to the first question is to make sure the user who ran the import is kept, then this issue would possibly resolve itself?

lholmquist:
I just updated the front-end code to display a message if there is no app configured

Screenshot 2026-08-31 at 1 46 33 PM

lholmquist:
I've updated the code with the suggestion from @AndrienkoAleksandr 

@kim-tsao @PatAKnight would you mind giving this another review?

kim-tsao:
> I've updated the code with the suggestion from @AndrienkoAleksandr
> 
> @kim-tsao @PatAKnight would you mind giving this another review?

/lgtm

The changes make sense and make the caller more identifiable.  Just a small nit, can we update the docs so they clearly state what is happening with this particular flow?  I think you have other flows that call `getOwnServiceCredentials()` which may be confusing

PR #4371, opened 2026-08-18
fix(orchestrator): preserve All Runs table column order after sorting

lokanandaprabhu:
## Summary

- Fixes the All Runs page table where dragging columns to reorder them would revert to the original sequence after clicking a column header to sort.

## Root Cause

Two layers caused the column order reset:

1. **Unstable `columns` memo**: The `columns` useMemo depended on `applyBackendSort`, which was recreated on every sort/data change (it closes over `orderDirection` and `runItems`). Each recreation produced a new columns array reference, causing material-table to reset its internal column order state.

2. **material-table loses column order on data changes**: Even with a stable `columns` reference, sorting triggers a backend data refetch producing a new `data` prop. material-table's `componentDidUpdate` detects this and calls `setColumns` to reconcile column state. It tries to preserve column order by matching columns via a top-level `id` property — but Backstage's `Table` component runs columns through `convertColumns()`, which creates new objects without `id`. The match always fails, so `columnOrder` resets to `[0, 1, 2, ...]`.

The Workflow List page is unaffected because it uses client-side sorting with no `onOrderChange`, no sorting state, and no data refetching on sort.

## Fix

- **Layer 1 — Stabilize the `columns` memo**: Wrapped `applyBackendSort` in a ref-based stable callback so the columns memo isn't recreated on sort/data changes.
- **Layer 2 — Manage column order in React state**: Track drag-and-drop order via `columnOrder` state and reorder the columns array before passing it to material-table. When material-table resets its internal `columnOrder` to `[0, 1, 2, ...]` after a data change, the columns are already in the user's desired visual order, making the reset a no-op.

### Video:

https://github.com/user-attachments/assets/7502902a-8602-4a9a-b9e4-7e52bd0c59e7

## Test plan

- [ ] Navigate to Orchestrator → All Runs tab
- [ ] Drag a column header to a different position
- [ ] Click a sortable column header (e.g., Started, Workflow Name) to sort
- [ ] Verify the dragged column order is preserved after sorting
- [ ] Verify sorting still works correctly with reordered columns
- [ ] Verify the Workflow List tab still works as before

### DIFF
--- a/workspaces/orchestrator/.changeset/fix-allruns-column-order-reset.md
+++ b/workspaces/orchestrator/.changeset/fix-allruns-column-order-reset.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-orchestrator': patch
+---
+
+Fix All Runs table column order resetting to default after sorting. Dragged column positions are now preserved across sort and data refetch operations.
--- a/workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/WorkflowRunsTabContent.tsx
+++ b/workspaces/orchestrator/plugins/orchestrator/src/components/OrchestratorPage/WorkflowRunsTabContent.tsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { useSearchParams } from 'react-router-dom';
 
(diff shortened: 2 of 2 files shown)


karthikjeeyar (approved):
This fix preserves the table column order in react state, which means after refresh the order is gone. However if we need it to be preserved in local-storage is a different story because it is not just for Orchestrator, the solution should be global in all the tables.

Merging this PR now as it addresses the outlined issue.

/lgtm

PR #4394, opened 2026-08-19
fix(adoption-insights): use service-to-service auth for TechDocs metadata lookup

PatAKnight:

The top techdocs chart does a small convenience lookup to TechDocs so we can show `site_name` instead of the raw entity name. That call was building the URL from unsanitized analytics fields (`namespace` / `kind` / `name`) and copying the caller's `Authorization` header onto the outbound request. If those fields are poisoned, that's SSRF plus leaking the user's token.

This keeps the lookup — if it fails we still just show the entity name, same as before — but:

- Base URL comes from Discovery (`techdocs`), not `backend.baseUrl` string concat
- Path segments that don't look like catalog names (`..`, slashes, etc.) skip the fetch entirely
- Auth is a plugin-to-plugin token (`getPluginRequestToken` with our own service credentials), not the user's bearer token

### DIFF
--- a/workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
+++ b/workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
@@ -104,7 +104,7 @@ If you want to run the entire project, including the frontend, run `yarn start` 
 
 `GET /api/adoption-insights/events`
 
-When `type=top_techdocs`, the backend may enrich each row with a TechDocs `site_name`. That lookup uses Backstage service-to-service authentication to the TechDocs plugin. If the lookup fails or the entity path is invalid, `site_name` falls back to the entity name.
+When `type=top_techdocs`, the backend may enrich each row with a TechDocs `site_name`. That lookup uses Backstage plugin-to-plugin authentication (`getPluginRequestToken`) on behalf of the calling user — it does not forward the caller's raw `Authorization` header. TechDocs therefore evaluates `catalog.entity.read` for that user; entities they cannot read fall back to the entity name, as do failed lookups and invalid entity paths. Rows with empty `namespace` / `kind` / `name` keep an empty `site_name` (TechDocs root).
 
 ## Query Parameters
 
--- a/workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
+++ b/workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
@@ -104,6 +104,8 @@ If you want to run the entire project, including the frontend, run `yarn start` 
 
 `GET /api/adoption-insights/events`
 
+When `type=top_techdocs`, the backend may enrich each row with a TechDocs `site_name`. That lookup uses Backstage service-to-service authentication to the TechDocs plugin. If the lookup fails or the entity path is invalid, `site_name` falls back to the entity name.
+
 ## Query Parameters
 
 | Parameter    | Type                | Required | Description                                                                                                                                     |
--- a/workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.test.ts
+++ b/workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.test.ts
@@ -94,6 +94,7 @@ describe('trackEvents', () => {
   let mockProcessorAddEvent: jest.Spied<EventBatchProcessor['addEvent']>;
 
   beforeEach(() => {
+    jest.clearAllMocks();
     controller = createController();
 
     mockProcessIncomingEvents = jest.spyOn(
@@ -235,11 +236,11 @@ describe('trackEvents', () => {
 
 describe('GetInsights', () => {
   beforeEach(() => {
+    jest.clearAllMocks();
     controller = createController();
     jest
       .mocked(mockAuth.getOwnServiceCredentials)
       .mockResolvedValue(mockCredentials.service());
-    jest.mocked(mockAuth.getPluginRequestToken).mockClear();
     jest.mocked(mockAuth.getPluginRequestToken).mockResolvedValue({
       token: 'mock-service-token',
     });
@@ -426,6 +427,7 @@ describe('GetInsights', () => {
     await controller.getTechdocsMetadata(result);
 
     expect(fetch).not.toHaveBeenCalled();
+    expect(mockDiscovery.getBaseUrl).not.toHaveBeenCalled();
     expect(mockAuth.getPluginRequestToken).not.toHaveBeenCalled();
     expect(result.data[0].site_name).toBe('');
   });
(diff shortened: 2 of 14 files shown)


workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
```diff
@@ -180,40 +201,73 @@ class EventApiController {
...
+
+      rowsToFetch.push({ row, url });
+    }
+
+    if (rowsToFetch.length === 0) {
+      return;
+    }
+
+    const { token } = await this.auth.getPluginRequestToken({
+      onBehalfOf: await this.auth.getOwnServiceCredentials(),
```
gustavolira: Question about the credentials choice here rather than the token handling itself.

`onBehalfOf: getOwnServiceCredentials()` means TechDocs resolves the entity as our service principal. Its `/metadata/techdocs/:namespace/:kind/:name` handler calls `httpAuth.credentials(req)` with no `allow`, so a service principal is accepted, and the catalog read it does downstream lands in `ServerPermissionClient`, which short-circuits every service-principal request to ALLOW without consulting any policy. With the caller's token, that same catalog read used to evaluate `catalog.entity.read` against the actual user, so someone who couldn't read the entity got a 404 and fell back to the raw name.

So with `permission.enabled: true`, anyone holding `adoptionInsightsEventsReadPermission` now gets `site_name` for TechDocs entities they aren't allowed to read. And since `POST /events` accepts free-form `attributes` (`src/validation/event.ts:27`) and `getTopTechDocsViews` aggregates globally, a user can seed a row for an entity of their choosing and read its title back.

Worth knowing that passing the caller's credentials wouldn't reintroduce what this PR fixes: for user principals `DefaultAuthService.getPluginRequestToken` wraps a limited user token inside a freshly issued plugin token, so it's still plugin-to-plugin, just carrying the user identity. `httpAuth` is already wired up in `plugin.ts`. Was dropping the per-entity check a deliberate trade, or a side effect of moving off the forwarded header?

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
```diff
@@ -180,40 +201,73 @@ class EventApiController {
...
   }
 
-  async getTechdocsMetadata(
-    req: Request<{}, {}, {}, QueryParams>,
-    result: TopTechDocsCount,
-  ) {
-    const promises: Promise<void>[] = [];
-    const baseUrl = await this.getBaseUrl('techdocs');
+  async getTechdocsMetadata(result: TopTechDocsCount) {
+    const baseUrl = await this.discovery.getBaseUrl('techdocs');
```
gustavolira: This await and the `getPluginRequestToken` at 239 both sit outside any error handling, and `getInsights` calls this method inside the try whose catch returns a 500 (line 195). Every other failure in here degrades to `row.site_name = row.name`, but if either of these rejects — the signing-key read inside `issueToken`, a discovery misconfiguration — the whole `GET /events?type=top_techdocs` fails and the user loses the counts as well as the titles. Before this PR that class of failure couldn't break the request at all, since the token was just a header copied off `req` and each fetch had its own `.catch`.

The PR description and the README line you added both promise the fallback, so I think this wants a try/catch around 205 and 239-242 that sets `row.site_name = row.name` across `rowsToFetch` and returns.

Separately, moving `getBaseUrl` below the `rowsToFetch.length === 0` early return would stop an empty result from calling discovery at all.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
```diff
@@ -180,40 +201,73 @@ class EventApiController {
...
+    await Promise.all(
+      rowsToFetch.map(({ row, url }) =>
+        fetch(url, {
+          headers: {
+            Accept: 'application/json',
+            Authorization: `Bearer ${token}`,
+          },
+        })
+          .then(async response => {
+            if (!response.ok) {
```
gustavolira, line 305: The `.catch` right below logs, but this branch doesn't. If TechDocs rejects the service token — wrong audience, access restrictions on the plugin token — every row comes back 401/403, every title silently degrades to the raw entity name, and there's nothing in the logs explaining why. A `logger.debug` for 404 (expected when docs were never built) and a `logger.warn` carrying `response.status` for anything else would make that diagnosable.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/utils/techdocsMetadata.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  const base = techdocsBaseUrl.endsWith('/')
+    ? techdocsBaseUrl.slice(0, -1)
+    : techdocsBaseUrl;
+  const path = `/metadata/techdocs/${encodeURIComponent(
+    namespace,
+  )}/${encodeURIComponent(kind)}/${encodeURIComponent(name)}`;
+  const url = `${base}${path}`;
+
+  if (!url.startsWith(`${base}/metadata/techdocs/`)) {
```
gustavolira: This can never be false. `url` is built two lines up as `${base}` followed by a template literal that starts with that exact prefix, so `startsWith` holds by construction no matter what the segments contain. It reads like a containment check backstopping the regex, but it isn't one — and it's the uncovered line Codecov is flagging. The test named "does not allow path traversal to escape the metadata prefix" passes because `isSafeTechDocsEntitySegment` rejects `..`, not because of this.

Either drop it, or make it a real check by comparing parsed pathnames so dot-segment normalization actually gets exercised:

```ts
const parsed = new URL(url);
if (!parsed.pathname.startsWith(`${new URL(base).pathname}/metadata/techdocs/`)) {
  return undefined;
}
```

workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
```diff
@@ -104,6 +104,8 @@ If you want to run the entire project, including the frontend, run `yarn start`
 
 `GET /api/adoption-insights/events`
 
+When `type=top_techdocs`, the backend may enrich each row with a TechDocs `site_name`. That lookup uses Backstage service-to-service authentication to the TechDocs plugin. If the lookup fails or the entity path is invalid, `site_name` falls back to the entity name.
```
gustavolira: Two things here don't match the code.

The fallback isn't universal — if `getPluginRequestToken` or `getBaseUrl` reject, the request returns a 500 instead of falling back (see my note on `EventApiController.ts:205`).

And when `namespace`/`kind`/`name` are empty, `site_name` is set to `''`, not the entity name (`EventApiController.ts:210`); the frontend then renders the literal string `docs` for that row (`plugins/adoption-insights/src/components/Techdocs/Techdocs.tsx:187-189`).

Since this paragraph is the operator-facing description, it'd also be worth saying that titles are now resolved with service credentials rather than the caller's.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.test.ts
```diff
@@ -56,6 +56,23 @@ const mockAuditor = {
   }),
 } as any;
 
+const mockAuth = mockServices.auth.mock();
```
gustavolira, line 59: nit: These are module-level and shared across both describe blocks, and the `jest.restoreAllMocks()` in the afterEach only restores `jest.spyOn` spies — it doesn't clear call history on the `jest.fn()`s that `mockServices.*.mock()` hands back. You're already working around that by calling `.mockClear()` on `getPluginRequestToken` explicitly, but `getBaseUrl`, `getOwnServiceCredentials` and `logger.warn` keep accumulating across the whole file.

A `beforeEach(() => jest.clearAllMocks())` would cover all of them, and would let you assert on `logger.warn` in the unsafe-segment test — right now that warn is the only operational signal for that path and nothing checks it.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
PatAKnight: Yep, I believe I misunderstood the requirements, `Dedicated Backstage service-to-service authentication must replace credential forwarding`, to mean that it should be plugin to plugin auth and completely neglected to think about OBO with service to service. Should be fixed in the recent commit.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
PatAKnight: Should be fixed in the recent commit.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/controllers/EventApiController.ts
PatAKnight: Should be fixed in the recent commit.

workspaces/adoption-insights/plugins/adoption-insights-backend/src/utils/techdocsMetadata.ts
PatAKnight: Should be fixed in the recent commit.

workspaces/adoption-insights/plugins/adoption-insights-backend/README.md
PatAKnight: Should be fixed in the recent commit.

AndrienkoAleksandr (approved):

AndrienkoAleksandr (approved):

kim-tsao (approved):
/lgtm

PR #4395, opened 2026-08-19
feat(scorecard): render scalar aggregation KPIs as a single-value card

ciiay:
## Description
Homepage aggregation cards now render scalar KPI types (`sum`, `average`, `min`, `max`, `count`) as a single large number with an optional threshold color, instead of the unsupported-type fallback. `AggregatedMetricCard` chooses the UI from the API result shape: a numeric `value` uses `ScalarStatCard`; a `values[]` distribution keeps the existing pie/donut cards.

This matches the backend scalar payload from RHIDP-14171 and the agreed 2.1 design (one number per card). Combined Min+Max on one card and Count/Sum bucket donuts are out of scope.

## Fixed
- [RHIDP-15517](https://redhat.atlassian.net/browse/RHIDP-15517) — Advanced Scorecard Aggregation logic card component

## Checklist

https://github.com/user-attachments/assets/16d0fbf7-ca51-46aa-b0ae-c7dca2373e11


workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+  aggregationId,
+  showSubheader = true,
+  showInfo = true,
+  dataTestId,
+}: ScalarStatCardProps) => {
+  const theme = useTheme();
+  const { t } = useTranslation();
+  const { result, metadata, id: scorecardId } = scorecard;
+
+  const matchingThresholdKey = getMatchingThresholdKey(
```
dzemanov, line 71: I don't think it makes sense to recalculate this in both backend and frontend. Backend should provide this information if it is needed.
cc @imykhno

dzemanov: (no need to block on this, can be updated in following PRs)

Eswaraiahsapram: Please skip threshold coloring when there are no successful samples, i.e. when `total === 0 && calculationErrorCount > 0`.

In the attached screenshot, all 5 entities failed, but the result is still being displayed in green. This is not expected. When there are no successful samples, it should default to grey.

Image

Eswaraiahsapram (changes requested):
Thanks @ciiay , everything looks good to me. I’ve requested one change regarding the homepage KPI issue.

1. All error / empty scalar aggregates can render a green 0 (value 0 matches the default success rule).

Agree with @dzemanov on not duplicating threshold evaluation in the frontend (follow-up is fine)

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+import { ScalarStatTile } from './ScalarStatTile';
+import type { ScalarStatCardProps } from './types';
+
+const aggregationTypeLabelKeys = {
+  min: 'metric.aggregationTypeMin',
+  max: 'metric.aggregationTypeMax',
+  sum: 'metric.aggregationTypeSum',
+  count: 'metric.aggregationTypeCount',
+  average: 'metric.aggregationTypeAverage',
+} as const;
```
imykhno: Are we sure that using `metric`. is the right approach?

I assumed that under the `metric` key, we would store all translations related specifically to metric translations. The terms `Min`, `Max`, etc., relate specifically to the aggregation logic. Therefore, I propose adding these translations under an `aggregation` key instead.

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
```diff
@@ -0,0 +1,120 @@
...
+  t: TranslationFunction<typeof scorecardTranslationRef.T>,
+): string {
+  if (isScalarAggregationType(aggregationType)) {
+    return t(aggregationTypeLabelKeys[aggregationType]);
+  }
+  if (!aggregationType) {
+    return '';
+  }
+  return aggregationType.charAt(0).toUpperCase() + aggregationType.slice(1);
+}
```
imykhno, line 51: I noticed that we already have a [resolveMetricTranslation](https://github.com/redhat-developer/rhdh-plugins/blob/99da62cced45fbd3410c8193cefbde8c4b06b721/workspaces/scorecard/plugins/scorecard/src/utils/translationUtils.ts#L34) method. However, it currently works only for the metric key. I’m thinking about refactoring resolveMetricTranslation to the following design:
``` typescript

export const resolveMetricTranslation = (t: ScorecardTranslationFunction, metricId: string, field: 'title' | 'description', fallback?: string) => {
  return resolveTranslation(`metric.${metricId}.${field}`, t, metricId, field, fallback);
}

export const resolveAggregationTranslation = (t: ScorecardTranslationFunction, aggregationId: string, field: 'title' | 'description', fallback?: string) => {
  return resolveTranslation(`aggregation.${aggregationId}.${field}`, t, aggregationId, field, fallback);
}

export function resolveTranslation(
  key: string,
  t: ScorecardTranslationFunction,
  metricId: string,
  field: 'title' | 'description',
  fallback?: string,
): string {
  const translated = t(key as any, {});
  if (translated !== key) return translated;

  const dotIndex = metricId.indexOf('.');
  if (dotIndex !== -1) {
    const base = metricId.slice(0, dotIndex);
    const name = metricId.slice(dotIndex + 1);
    const templateKey = `metric.${base}.${field}`;
    const templateTranslated = t(templateKey as any, { name });
    if (templateTranslated !== templateKey) return templateTranslated;
  }

  return fallback ?? translated;
}
```

What do you think about this approach? Could it help?

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
imykhno: > I don't think it makes sense to recalculate this in both backend and frontend. Backend should provide this information if it is needed.

I agree. I can update the response for scalar types as follows:

- Add `result.aggregationChartDisplayColor` (using the same logic as the `weightedStatusScore` aggregation type). The color will be calculated on the backend, returning only the correct value.
- Remove `result.thresholds`, as we are not using it on the UI.

If we are aligned on this approach, I will create a bug ticket.

CC @dzemanov

imykhno (changes requested):

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
ciiay: Thanks for the suggestion. A shared helper would be useful later if we need per-aggregation-id titles/descriptions (`aggregation.<id>.title`).

This card only needs the five type labels (Min / Max / Sum / Count / Average), so I moved those to `aggregation.min` (etc.) instead of building the cascade now. The shared `resolveTranslation` fallback in the snippet still looks up `metric.${base}.${field}`, which wouldn’t be right for aggregations.

I’d rather add `resolveAggregationTranslation` when we actually translate aggregation IDs the same way we do metrics. Does that work for you?

cc. @Eswaraiahsapram

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
ciiay: Agreed, not blocking. I added a grey fallback for `total === 0` in this PR so all-failed/empty cards don’t render as success. We can switch to a backend-provided color (`aggregationChartDisplayColor`) in a follow-up.

Frontend follow-up: https://issues.redhat.com/browse/RHIDP-16510 (related to RHIDP-15517). If you have a backend Jira for this, please link it as a blocker of RHIDP-16510.

cc. @Eswaraiahsapram

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
imykhno: Yes, I agree that it's make sense. Thank you for changes you made!

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/ScalarStatCard.tsx
```diff
@@ -0,0 +1,122 @@
...
```
imykhno: I’d like to get your perspective on a small refactoring I made to this file:
```suggestion
  resolveStatusColor,
} from '../../../utils';
import { useTranslation } from '../../../hooks/useTranslation';
import { scorecardTranslationRef } from '../../../translations';
import { CardInfoButton } from '../components/CardInfoButton';
import { CardSubheader } from '../components/CardSubheader';
import { CardChartContainer } from '../components/CardChartContainer';
import { formatAggregationScoreDetail } from '../WeightedStatusScoreCard/TooltipContent';
import { ScalarStatTile } from './ScalarStatTile';
import type { ScalarStatCardProps } from './types';
import { AggregationType } from '@red-hat-developer-hub/backstage-plugin-scorecard-common';

function getAggregationTypeLabel(
  const key = `aggregation.${aggregationType}`;
  const translated = t(key as any, {});

  if (translated !== key) {
    return translated;
  }

  return aggregationType.charAt(0).toUpperCase() + aggregationType.slice(1);
}
```

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/types.ts
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+import {
+  AggregatedMetricResult,
+  AggregationMetadata,
+  ScalarAggregationResult,
+} from '@red-hat-developer-hub/backstage-plugin-scorecard-common';
+import { AggregatedMetricCardBaseProps } from '../types';
+
```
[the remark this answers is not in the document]
imykhno: CC @ciiay

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/ScalarStatCard/types.ts
```diff
@@ -0,0 +1,29 @@
...
+export type ScalarStatCardProps = AggregatedMetricCardBaseProps & {
```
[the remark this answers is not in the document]
imykhno: CC @ciiay

workspaces/scorecard/plugins/scorecard/README.md
```diff
@@ -241,7 +241,7 @@ To align with the legacy EntityPage (Scorecard on component pages and default en
                    xxs: { w: 4, h: 6, x: 8 }
```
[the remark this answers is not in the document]
imykhno: CC @ciiay

imykhno (approved):
Thank you for the changes! I tested the PR locally, and everything works as expected:

Image

Translation also works fine:

Image

/lgtm

Eswaraiahsapram (approved):
Thanks @ciiay , I tested the PR locally, and everything works as expected:

Image

PR #4405, opened 2026-08-20
fix(theme): keep page tabs below the global header masthead

ciiay:
## Description
NFS never reserved masthead height, so theme `100vh` page-shell rules and the fixed sidebar sat under the global header and clipped HeaderTabs (RHDHBUGS-3627). This change publishes `--rhdh-global-header-height` from the header layout, subtracts it from theme viewport sizing, and offsets `BackstageSidebar-drawer` so tabs and sidebar content sit fully below the masthead.

The NFS demo app now matches OFS: branding stays in the header, not a duplicate sidebar logo.

Inspect Entity being covered by the masthead is a separate overlay/z-index issue and is tracked in RHDHBUGS-3603.

## Fixed
- [RHDHBUGS-3627](https://redhat.atlassian.net/browse/RHDHBUGS-3627) — Top navigation items partially cut off by masthead/global header across all themes

## Checklist

https://github.com/user-attachments/assets/af441deb-6fb5-4937-a4bb-0ded2a6901ef

screen recording on app-next:

https://github.com/user-attachments/assets/6654d0fd-abf6-44bf-a2e7-07ca7b66295e


ciiay:
RHDHBUGS-3603 (Inspect Entity covered by the masthead) was split out of this PR. That overlay/z-index issue is independent of the page-shell HeaderTabs fix here.

rohitratannagar:
Tested it in rhdh-local, found some issues 
main content in the rhdh themes is displaced towards the sidebar

https://github.com/user-attachments/assets/767aecc4-a97e-4e7a-9254-2eff999781dd

ciiay:
@rohitratannagar Thanks for testing this.

The recording looks like **rhdh-local / OFS**. This change is aimed at **NFS (app-next)** — OFS is being deprecated in the upcoming release, so the page-shell contract we need to preserve is NFS.

Could you re-test on **NFS / app-next** with this PR’s theme + global-header overlays, especially:
- Catalog / Create / entity HeaderTabs vs the masthead
- Home and Settings with **RHDH Light (latest)** and **RHDH Dark (latest)**
- Whether the content well still sits flush to the sidebar the way it does on current NFS

The RHDH-theme well sitting against the sidebar is existing PF6 page-inset behavior (`margin-left: 0` when the sidebar `nav` is present), not something this PR introduced for NFS. Changing that inset would shift NFS as well, which we don’t want to do in this ticket.

its-mitesh-kumar:
> Tested it in rhdh-local, found some issues main content in the rhdh themes is displaced towards the sidebar
> 
>  Screen.Recording.2026-09-02.at.3.34.41.PM.mov

@rohitratannagar Could you please verify the changes again?

christoph-jerolimov (changes requested):
I need to take a deeper look, but I don't think that its needed to calculate and update any heights at runtime! Instead we should think about an App Layout change in app-defaults.

PR #4410, opened 2026-08-20
fix(install-dynamic-plugins): resolve ref:// before the pre-merge disabled pass

jonkoops:
The pre-merge pass only inspects `oci://` entries to determine which plugins can be skipped. A `ref://` entry with `enabled: true` was invisible to this pass, so it could not prevent a disabled DPDY entry from being filtered out. This caused the later `resolveRefPlugin` call in `mergePlugin` to fail with "no plugin named … found in included plugins" because the target entry had already been removed.

This moves `ref://` resolution earlier in `loadAllPlugins`, before the pre-merge pass runs. Once resolved, the entries are standard `oci://` URLs and the level-override logic works correctly. The old `resolveRefPlugin` function and its call in `mergePlugin` are removed since resolution now happens up front.

Also extracts `IncludePluginList` into `types.ts` as a shared type.

[RHIDP-13226](https://redhat.atlassian.net/browse/RHIDP-13226)

### DIFF
--- a/workspaces/install-dynamic-plugins/.changeset/fix-ref-pre-merge-disabled.md
+++ b/workspaces/install-dynamic-plugins/.changeset/fix-ref-pre-merge-disabled.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/cli-module-install-dynamic-plugins': patch
+---
+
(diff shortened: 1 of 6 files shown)


gashcrumb (approved):
👍

PR #4415, opened 2026-08-20
Add SQLite3 build natively and no cache on build

asmasarw:
## Summary

Fixes two issues with the Docker image build that caused:
1. Stale UI being served from the container (local changes not reflected in the built image)
2. `better-sqlite3` native addon missing at runtime, breaking all database-backed plugins (catalog, permission, dcm, search)

### 1. Docker layer caching serves stale frontend

The `generate-image.sh` script runs `podman build` / `docker build` without `--no-cache`. When source files in `plugins/` or `packages/` change, the build engine may reuse cached layers for the `yarn workspace app build` step, producing an image with an outdated frontend bundle.

### 2. `better-sqlite3` native binary not compiled

The production stage of the Dockerfile ran:

```dockerfile
RUN yarn workspace backend exec node-gyp rebuild || true
```

This executes `node-gyp rebuild` in the **backend package directory** (`packages/backend/`), which has no `binding.gyp` — so it silently does nothing. The `|| true` swallows the failure. As a result, `better-sqlite3` never gets its `.node` native addon compiled, and at runtime every plugin that uses SQLite fails with:

```
Could not locate the bindings file. Tried:
  → /app/node_modules/better-sqlite3/build/better_sqlite3.node
  → ...
```

### 1. Disable build cache (`generate-image.sh`)

Added `--no-cache` to the `podman build` / `docker build` invocation to ensure every layer is rebuilt from scratch, guaranteeing the latest source is always compiled into the image.

### 2. Rebuild `better-sqlite3` native addon for the container's architecture

Replaced the no-op `node-gyp rebuild` with:

```dockerfile
RUN npm rebuild better-sqlite3
```

This runs `node-gyp` inside the `better-sqlite3` package where `binding.gyp` lives, correctly producing the native `.node` binary for the container's architecture. Removed `|| true` so the build fails loudly if compilation breaks.

### DIFF
--- a/workspaces/dcm/Dockerfile
+++ b/workspaces/dcm/Dockerfile
@@ -54,17 +54,13 @@ RUN tar xzf skeleton.tar.gz && rm skeleton.tar.gz
 COPY --from=build /app/package.json /app/yarn.lock /app/.yarnrc.yml ./
 COPY --from=build /app/.yarn ./.yarn
 
-# FIX 2: REMOVED the inline Node script that was muting postinstall scripts.
-# FIX 3: Run production install allowing native C++ compilation to happen naturally.
 RUN YARN_ENABLE_IMMUTABLE_INSTALLS=false \
     YARN_CACHE_FOLDER=/root/.yarn/berry/cache \
     yarn workspaces focus --all --production
 
-# FIX 4: Explicitly force a targeted native compilation for your runtime environment
-RUN yarn workspace backend exec node-gyp rebuild || true
-
-# Clean up yarn cache after native build is done
-RUN rm -rf "$(yarn cache dir)"
+# Purge build tools and yarn cache — no longer needed at runtime
+RUN apt-get purge -y --auto-remove python3 make g++ \
+    && rm -rf /var/lib/apt/lists/* "$(yarn cache dir)"
 
 COPY --from=build /app/packages/app/package.json packages/app/package.json
 RUN ln -sf ../packages/app node_modules/app
--- a/workspaces/dcm/Dockerfile
+++ b/workspaces/dcm/Dockerfile
@@ -60,8 +60,8 @@ RUN YARN_ENABLE_IMMUTABLE_INSTALLS=false \
     YARN_CACHE_FOLDER=/root/.yarn/berry/cache \
     yarn workspaces focus --all --production
 
-# FIX 4: Explicitly force a targeted native compilation for your runtime environment
-RUN yarn workspace backend exec node-gyp rebuild || true
+# Rebuild better-sqlite3 native addon for the container's architecture
+RUN npm rebuild better-sqlite3
 
 # Clean up yarn cache after native build is done
 RUN rm -rf "$(yarn cache dir)"
--- a/workspaces/dcm/scripts/generate-image.sh
+++ b/workspaces/dcm/scripts/generate-image.sh
@@ -31,7 +31,7 @@ function _usage {
 Usage:
   $script_name clean
   VERSION=<semver> $script_name [oci|tgz|push]
-  VERSION=<semver> $script_name backstage-image [--push]
+  VERSION=<semver> $script_name backstage-image [--push] [--no-cache]
 
 Environment variables:
   VERSION         Required. The plugin version; also used to tag the container image.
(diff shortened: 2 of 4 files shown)


workspaces/dcm/Dockerfile
```diff
@@ -60,8 +60,8 @@ RUN YARN_ENABLE_IMMUTABLE_INSTALLS=false \
     YARN_CACHE_FOLDER=/root/.yarn/berry/cache \
     yarn workspaces focus --all --production
 
-# FIX 4: Explicitly force a targeted native compilation for your runtime environment
-RUN yarn workspace backend exec node-gyp rebuild || true
+# Rebuild better-sqlite3 native addon for the container's architecture
+RUN npm rebuild better-sqlite3
```
mareklibra, line 64: This workspace is Yarn-only (`packageManager: yarn@4.15.0`). Yarn 4.14+ skips install scripts by default, which is why `better-sqlite3` has no `.node` after `yarn workspaces focus --all --production`. Please don't work around that with `npm rebuild` (GHAS already flagged it).

Preferred: opt the package in so the existing Yarn install compiles it, matching Backstage:

In package.json:
```json
"dependenciesMeta": {
  "better-sqlite3": { "built": true }
}

Comment (mareklibra) [reply workspaces/dcm/Dockerfile:64]: If you want to keep an explicit rebuild (I doubt):

```
RUN YARN_ENABLE_SCRIPTS=true yarn rebuild better-sqlite3
```

File: workspaces/dcm/scripts/generate-image.sh [hunk 2]
```diff
@@ -217,6 +217,7 @@ function backstage-image {
 
   echo "Building Backstage app image: $image_tag"
   "$_pocker" build \
+    --no-cache \
```
mareklibra, line 220: Are you sure this is the right approach?
Can you double-check that a stale container was/wasn't used?

mareklibra, line 220: `--no-cache` on every `backstage-image` run will rebuild apt, yarn install, tsc, and the frontend bundle from scratch, including when only a small source file changed.

The Dockerfile already COPYs `packages/` and `plugins/` before `yarn workspace app build`, and `.dockerignore` excludes `dist`/`node_modules`, so those COPY layers should already invalidate when sources change. If a stale UI was observed, it is worth confirming it was actually a cache-invalidation miss (and fixing that) rather than disabling cache globally.

Please make `--no-cache` opt-in (same pattern as `--push`), e.g. `VERSION=… ./scripts/generate-image.sh backstage-image --no-cache`.

workspaces/dcm/Dockerfile
```diff
@@ -60,8 +60,8 @@ RUN YARN_ENABLE_IMMUTABLE_INSTALLS=false \
...
```
mareklibra, line 65: Please add a build-time smoke test after the rebuild, e.g. `node -e "require('better-sqlite3')(':memory:').close()"`.

CI does not build this Dockerfile, and `app-config.production.yaml` uses `client: better-sqlite3`, so a missing `.node` only appears when the container starts. Failing the image build is much cheaper than debugging a crashloop.

workspaces/dcm/Dockerfile
mareklibra, line 44: Now that native compilation actually runs, consider purging `python3`/`make`/`g++` after the rebuild. They are only needed at build time and otherwise stay in the runtime image that runs as an arbitrary OpenShift UID.

workspaces/dcm/Dockerfile
```diff
@@ -60,8 +60,8 @@
...
```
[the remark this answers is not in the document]
mareklibra: Swicthing to yarn rebuild is the cleanest solution

mareklibra (changes requested):

workspaces/dcm/scripts/generate-image.sh
mareklibra, line 47: Please mention `--no-cache` here as well (the usage line already has it), e.g. “Pass `--push` to push, `--no-cache` to rebuild layers from scratch.”

mareklibra (approved):
LGTM, just a small nit about documentation

workspaces/dcm/scripts/generate-image.sh
asmasarw: Done

PR #2800, opened 2026-04-16
feat(theme): add initial Backstage UI support

christoph-jerolimov:
## Hey, I just made a Pull Request!

WIP screenshots:

image
image
image
image
image

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ ] Added or Updated documentation
- [ ] Tests for new functionality and regression tests for bug fixes
- [x] Screenshots attached (for UI changes)

PR #4418, opened 2026-08-20
fix(theme): fill NFS catalog entity tab content well

ciiay:
## Description
On NFS catalog entity pages, Topology and Scorecard sat in a short BUI `Container` inside a tall `<main>`, leaving an empty band in the page well. Theme 1.0.2 already sized `SidebarPage > main`, but `main` stayed `display: block`, so the nested container’s `flex: 1` never grew.

This makes classless BUI `<main>` (not Backstage `Page`, which uses `display: grid`) a flex column so those tabs fill the well.

## Fixed
- [RHDHBUGS-3543](https://redhat.atlassian.net/browse/RHDHBUGS-3543) — [NFS] Empty gap below #root on Catalog entity tabs (Topology / Scorecard)

## Checklist

Verified on app-next, via https://github.com/redhat-developer/rhdh/pull/5115:
image

### DIFF
--- a/workspaces/theme/.changeset/bui-main-flex.md
+++ b/workspaces/theme/.changeset/bui-main-flex.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
+Make NFS BUI catalog entity `<main>` a flex column so Topology and Scorecard fill the content well instead of leaving a gap below short tab content (RHDHBUGS-3543).
--- a/workspaces/theme/plugins/theme/src/utils/createComponents.test.ts
+++ b/workspaces/theme/plugins/theme/src/utils/createComponents.test.ts
@@ -117,6 +117,24 @@ describe('createComponents', () => {
     );
   });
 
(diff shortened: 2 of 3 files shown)


logonoff (commented):
Can you check this isn't fixed by https://github.com/redhat-developer/rhdh-plugins/pull/2800

ciiay:
> Can you check this isn't fixed by #2800

Thanks for the comment. The other PR won't cover the fix of this bug. That PR adds BUI PatternFly tokens and chrome (page-inset borders, hide duplicate PluginHeader, `.bui-Container { max-width: none }`). The closest layout rule is `height: 100vh` on `BackstageSidebarPage` when there is a nav and no Backstage Page. That is the same kind of shell fill as theme 1.0.2. It does not make `<main>` a flex column. The gap I verified was: `main` 880px / `display: block`, Topology `bui-Container` 252px / `flex: 1 1 0%`. Flex-grow never ran. #2800 never touches `main`’s `display`.

logonoff (approved):

PR #4433, opened 2026-08-21
feat(intelligent-assistant): return supportsVision on GET /v1/models

JslYoon:
## What

Enriches the `GET /v1/models` response with a per-model `supportsVision` flag so the frontend knows each model's image capability at model-list load time, removing the need for a per-user `POST /v1/validate-model-vision` round-trip.

Jira: [RHIDP-16443](https://redhat.atlassian.net/browse/RHIDP-16443)

## Why

Previously vision support was only discoverable via `POST /v1/validate-model-vision`, which sends a test-inference request to LCS per model. Models are stable at runtime, so vision capability can be resolved once and surfaced on the model list instead of repeatedly per user/session.

## How

- `GET /v1/models` is now served by a dedicated handler (replacing the generic proxy for this one route). It fetches the LCS model list and, for each LLM, resolves vision capability and returns `{ ...model, supportsVision }`.
- Extracted a shared `probeModelVisionSupport` helper used by both the models handler and `POST /v1/validate-model-vision`. Results are memoised per-pod for 24h in the existing `ModelCapabilitiesCache`, so each model is probed at most once per pod.
- Non-LLM models (e.g. embeddings) are reported as `supportsVision: false` without a probe.
- A single failed/timed-out probe is logged and reported as non-vision rather than failing the whole list; the probe uses a 10s timeout.
- The shared helper caches a negative result on a non-ok probe response, which also resolves a prior gap where `/v1/validate-model-vision` neither cached nor responded on a non-ok (but non-throwing) upstream reply.
- Added optional `supportsVision?: boolean` to the frontend `LCSModel` type.

`/v1/models` remains in `SKIP_USER_ID_ENDPOINTS`, so no `user_id` is forwarded to LCS (unchanged behavior).

## Testing

- Added tests: probe success → `supportsVision: true`; non-ok probe → `false`; cache hit is not re-probed; embedding models are not probed.
- Full backend suite: 309 passing. `tsc`, `lint`, and `prettier` clean. API reports unchanged.


workspaces/intelligent-assistant/.changeset/models-supports-vision.md
```diff
@@ -0,0 +1,6 @@
+---
+'@red-hat-developer-hub/backstage-plugin-intelligent-assistant-backend': minor
+'@red-hat-developer-hub/backstage-plugin-intelligent-assistant': patch
```
Jdubrick: Out of curiosity, why do these diverge?

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/attachment-validation.ts
```diff
@@ -14,7 +14,11 @@
  * limitations under the License.
  */
 
-const CACHE_TTL_MS = 24 * 60 * 60 * 1000;
+// `true` is stable, so cache it for a day. `false` is cached only briefly: LCS
+// returns the same 5xx for a non-vision model and a transient error, so a short
+// TTL lets a real model recover without re-probing on every request.
+export const CACHE_TTL_MS = 24 * 60 * 60 * 1000;
+export const NEGATIVE_CACHE_TTL_MS = 20 * 60 * 1000;
```
Jdubrick: I know we can read it, but a comment with the human-readable time would be good for others

workspaces/intelligent-assistant/.changeset/models-supports-vision.md
JslYoon: It's because I thought it was patch as `/models` endpoint had changes, and a new feature `model validation` is being added to it. I can simply make it a `patch`

workspaces/intelligent-assistant/plugins/intelligent-assistant-backend/src/service/attachment-validation.ts
JslYoon: updated!

Jdubrick (approved):
lgtm

PR #4443, opened 2026-08-24
chore(app-defaults): add fixed changeset group for app-defaults and app-react

christoph-jerolimov:
## Summary
- Adds a `fixed` group to the changeset config so that `@red-hat-developer-hub/backstage-plugin-app-defaults` and `@red-hat-developer-hub/backstage-plugin-app-react` are always released together with the same version number.

## Test plan
- [ ] Verify the changeset config is valid JSON
- [ ] Confirm that future changeset version bumps apply to both packages simultaneously


davidfestal (approved):

christoph-jerolimov:
Thanks @davidfestal

PR #4449, opened 2026-08-24
feat(app-defaults): add empty state overrides for catalog pages

christoph-jerolimov:
## Summary
- Override page extensions for catalog, scaffolder (templates sub-page), API docs, TechDocs, and catalog graph plugins with an `EmptyCatalogGate` that checks whether matching entities exist before rendering the original page
- When no entities are found, a translatable empty state with a background illustration, action link, and support button is shown
- A `FrontendFeatureLoader` default export registers all overrides alongside the existing `appDefaultsModule` so a single package import loads everything
- Translations provided for en (default), de, es, fr, it, and ja

## Test plan
- [x] TypeScript compiles cleanly
- [x] ESLint passes
- [x] 11 unit tests pass (useCatalogEntities + EmptyCatalogGate)
- [x] Verify empty state renders on `/catalog` when no entities are registered
- [x] Verify original page renders on `/catalog` when entities exist
- [ ] Verify `/create`, `/api-docs`, `/docs`, `/catalog-graph` show empty states appropriately
- [ ] Verify scaffolder sub-pages (actions, editor, tasks) remain accessible

Ref: RHIDP-14871


karthikjeeyar (approved):
I have verified the empty state for the catalog pages. It looks good to me.


Only thing missing is, the translation needs to be wired for all the languages and the `appDefaultsTranslations` should be export & added to the `appDefaultsModule` (api-report needs to be re-generated)

```
export const appDefaultsTranslations = createTranslationResource({
  ref: translationRef,
  translations: {
    de: () => import('./de'),
    es: () => import('./es'),
    fr: () => import('./fr'),
    it: () => import('./it'),
    ja: () => import('./ja'),
  },
});

```

workspaces/app-defaults/plugins/app-defaults/src/translations/ref.ts
```diff
@@ -0,0 +1,66 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { createTranslationRef } from '@backstage/frontend-plugin-api';
+
+/**
+ * @internal
```
karthikjeeyar:
```suggestion
 * @public
```

Marked as  `@Internal`  which prevents it from being exported in the generated TypeScript declaration file (.d.ts).

PR #4464, opened 2026-08-25
FLPATH-4773: Reflect change introduced by environment agent to UI

asmasarw:
## Replace Providers tab with Agents tab

Removes all Providers-related code and introduces a new **Agents** tab as the default landing tab in the DCM plugin.

### `dcm-common`
- Deleted `ProvidersApi`, `ProvidersClient`, `ProvidersClient.test.ts`, and `types/providers.ts`
- Added `AgentsApi`, `AgentsClient`, `AgentsClient.test.ts`, and `types/agents.ts` with types derived from the Agent API OpenAPI spec (`v1alpha1`): `Agent`, `AgentList`, `AgentRegistrationRequest`, `AgentCost`, `AgentHealthStatus`, `HeartbeatRequest`

### `dcm` (frontend plugin)
- Replaced `providersApiRef` / `providersRouteRef` with `agentsApiRef` / `agentsRouteRef`
- Deleted the `pages/providers/` directory (all provider UI components)
- Added `pages/agents/` with:
  - `AgentsTabContent` — paginated table with health-status filtering and agent registration dialog
  - `AgentFormFields` — registration form backed by service types fetched live from the Catalog API
  - `AgentHealthStatus` — status badge component
  - `CopyButton` — reusable clipboard copy component
  - `agentFormTypes` — form state, Yup validation, and mapping helpers
- Updated all translation files (`ref.ts`, `de`, `es`, `fr`, `it`, `ja`) to replace `providers.*` keys with `agents.*` keys
- Updated API report files for both packages
- Added a changeset (`minor` bump for both packages)


workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,238 @@
...
+    loadMore: loadMoreServiceTypes,
+    error: serviceTypesError,
+  } = useInfiniteSelect((token?: string) =>
+    catalogApi.listServiceTypes({ max_page_size: 100, page_token: token }),
+  );
+
+  const crud = usePaginatedCrudTab<Agent, AgentForm>({
+    loadFn: ({ pageToken, pageSize: ps }) =>
+      agentsApi
+        .listAgents({ page_token: pageToken, max_page_size: ps })
```
mareklibra: The PR description and changeset say this tab supports health-status filtering, but `listAgents` is only called with pagination params and there is no filter UI.

Please either:
1. Add a health-status filter (and pass `health_status` into `listAgents`), or
2. Drop that claim from the changeset / PR body so the shipped behavior matches the docs.

The client already supports the query param, so wiring a select/chips control should be small.

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.test.tsx
```diff
@@ -0,0 +1,234 @@
...
+        [agentsApiRef, apis.agents],
+        [catalogApiRef, apis.catalog],
+      ]}
+    >
+      <AgentsTabContent />
+    </TestApiProvider>,
+  );
+}
+
+describe('AgentsTabContent', () => {
```
mareklibra, line 88: The PR description and changeset say this tab supports health-status filtering, but `listAgents` is only called with pagination params and there is no filter UI.

Please either:
1. Add a health-status filter (and pass `health_status` into `listAgents`), or
2. Drop that claim from the changeset / PR body so the shipped behavior matches the docs.

The client already supports the query param, so wiring a select/chips control should be small.

workspaces/dcm/.changeset/replace-providers-with-agents.md
```diff
@@ -0,0 +1,14 @@
+---
+'@red-hat-developer-hub/backstage-plugin-dcm': minor
```
mareklibra: Is there already any production use of the DCM?

This removes `@public` exports (`ProvidersApi`, `ProvidersClient`, `providersApiRef`, `providersRouteRef`) from packages already at v1.0.0. That is a breaking change.

Please bump both packages to **major** unless DCM 1.0.0 is explicitly unpublished / internal-only and that is documented in the changeset.

Also please mention the removed public symbols in the changeset so downstream consumers are not surprised.

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,238 @@
...
+  },
+}));
+
+export function AgentsTabContent() {
+  const classes = useStyles();
+  const agentsApi = useApi(agentsApiRef);
+  const catalogApi = useApi(catalogApiRef);
+  const { t } = useTranslation();
+
+  const {
```
mareklibra, line 98: `useInfiniteSelect` already exposes `loading` for the first page, but it is unused. `AgentFormFields` only spins on `loadingMoreServiceTypes`, so the first fetch looks like “no service types”.

Please pass `loading` through and show the same `CircularProgress` `MenuItem` while the first page is in flight.

workspaces/dcm/plugins/dcm/src/translations/ref.ts
mareklibra, line 248: `instances.emptyDescription` still says “registered provider infrastructure”. With Providers removed, please reword (e.g. “registered environment agents”) and update the locale files to match.

workspaces/dcm/plugins/dcm/src/pages/agents/agentFormTypes.ts
```diff
@@ -0,0 +1,132 @@
...
+      .array()
+      .of(yup.string().required())
+      .min(
+        1,
+        m(
+          'validation.agent.serviceTypesRequired',
+          'At least one service type is required',
+        ),
+      ),
+    cost: yup
```
mareklibra, line 71: `cost` is only `.required()`, so any non-empty string passes `isAgentFormValid`. Please constrain it to `AGENT_COST_OPTIONS`:

```ts
cost: yup
  .string()
  .oneOf(AGENT_COST_OPTIONS, m('validation.agent.costRequired', 'Cost is required'))
  .required(m('validation.agent.costRequired', 'Cost is required')),
```

workspaces/dcm/plugins/dcm-common/src/clients/AgentsApi.ts
```diff
@@ -15,17 +15,23 @@
...
  */
-export interface ProvidersApi {
-  listProviders(params?: PaginationParams): Promise<ProviderList>;
-  getProvider(providerId: string): Promise<Provider>;
-  createProvider(provider: Provider): Promise<Provider>;
-  applyProvider(providerId: string, provider: Provider): Promise<Provider>;
-  deleteProvider(providerId: string): Promise<void>;
+export interface AgentsApi {
+  listAgents(
+    params?: PaginationParams & { health_status?: string },
```
mareklibra: `Agent.health_status` is typed as `AgentHealthStatus`, but the list filter is still `health_status?: string`. Please use the same union so callers cannot pass arbitrary values:

```ts
listAgents(
  params?: PaginationParams & { health_status?: AgentHealthStatus },
): Promise<AgentList>;

Comment (mareklibra) [reply workspaces/dcm/plugins/dcm-common/src/clients/AgentsApi.ts]: Same change in AgentsClient.listAgents (line 41).

File: workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx [hunk 7]
```diff
@@ -0,0 +1,238 @@
...
+              bold={false}
+              maxWidth={200}
+              fallback={<DcmEmptyCell />}
+            />
+            {a.topic_name && <CopyButton text={a.topic_name} />}
+          </Box>
+        ),
+      },
+      {
+        title: t('agents.columns.health'),
```
mareklibra, line 235: `last_heartbeat` is on the Agent model but never shown. A “Last heartbeat” column next to Health (or a tooltip on the badge) would make `unavailable` vs a stale `ready` easier to diagnose. Other tabs already render timestamps the same way, e.g.:

    {
      title: t('agents.columns.lastHeartbeat'),
      field: 'last_heartbeat',
      render: a =>
        a.last_heartbeat ? (
          <Typography variant="body2">
            {new Date(a.last_heartbeat).toLocaleString()}
          </Typography>
        ) : (
          <DcmEmptyCell />
        ),
    },

Optional: in `AgentFormFields`, default `topic_name` to `dcm.agent.${name}` while the user is still on the generated value, so they do not have to retype the prefix.

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,238 @@
...
+        search={crud.search}
+        onSearchChange={crud.handleSearchChange}
+        cursorPagination={crud.cursorPagination}
+        emptyTitle={t('agents.emptyTitle')}
+        emptyDescription={t('agents.emptyDescription')}
+        primaryActionLabel={t('agents.registerButton')}
+        onPrimaryAction={crud.handleOpenCreate}
+        illustrationSrc={emptyIllustration}
+        entityLabel={t('agents.entityLabel')}
+        actionError={serviceTypesError}
```
mareklibra: Minor consistency nit vs Providers / Catalog items: they pass `onDismissActionError` so the service-type load alert can be closed. Here only `actionError={serviceTypesError}` is set, so MUI renders a non-dismissable banner.

`useInfiniteSelect` has no `clearError`, so a local flag is enough:

    const [serviceTypesErrorDismissed, setServiceTypesErrorDismissed] = useState(false);
    // ...
    actionError={serviceTypesErrorDismissed ? null : serviceTypesError}
    onDismissActionError={() => setServiceTypesErrorDismissed(true)}

Not blocking - sticky is reasonable for a failed catalog fetch. Nice to match the other tabs.

mareklibra (changes requested):
A few comments, nothing serious.

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,332 @@
...
+        search={crud.search}
+        onSearchChange={crud.handleSearchChange}
+        cursorPagination={crud.cursorPagination}
+        emptyTitle={t('agents.emptyTitle')}
+        emptyDescription={t('agents.emptyDescription')}
+        primaryActionLabel={t('agents.registerButton')}
+        onPrimaryAction={crud.handleOpenCreate}
+        illustrationSrc={emptyIllustration}
+        entityLabel={t('agents.entityLabel')}
+        toolbarExtra={healthFilterControl}
```
mareklibra, line 303: Health filter on empty results is a dead end. DcmCrudTabLayout only renders toolbarExtra in the table branch; a health filter that returns 0 agents hits items.length === 0 && !hasPrev and shows the global empty state (“No agents registered”) with no way to change the filter short of a full reload.

Please either:
1. Render toolbarExtra in the empty-state branch as well, and use different copy when a filter is active, or
2. Skip the illustration empty state when healthFilter is set and keep the table + select visible.

resetAndReload() already clears the token stack, so hasPrev will not save this path.

workspaces/dcm/plugins/dcm/src/components/DcmCrudTabLayout.tsx
mareklibra: See AgentsTabCOntent.tsxx:200 comment

mareklibra:
```suggestion
if (items.length === 0 && !cursorPagination?.hasPrev) {
  return (
    <>
      {toolbarExtra && (
        <Box className={classes.toolbarRow}>{toolbarExtra}</Box>
      )}
      {/* existing actionError + DcmDataCenterTabEmptyState */}
    </>
  );
}
```

workspaces/dcm/plugins/dcm/src/pages/agents/components/AgentFormFields.tsx
```diff
@@ -0,0 +1,242 @@
...
+  setForm: React.Dispatch<React.SetStateAction<AgentForm>>;
+  touched: TouchedMap;
+  setTouched: React.Dispatch<React.SetStateAction<TouchedMap>>;
+  serviceTypes?: ServiceType[];
+  loadingServiceTypes?: boolean;
+  loadingMoreServiceTypes?: boolean;
+  loadMoreServiceTypes?: () => void;
+}>;
+
+export function AgentFormFields({
```
mareklibra, line 66: The Providers form and CopyButton unit tests were deleted and not replaced. AgentsTabContent tests cover list/pagination/filter but never open the register dialog, and usePaginatedCrudTab.test.ts does not cover resetAndReload (the helper the filter depends on).

Please port the old ProviderFormFields / CopyButton cases onto AgentFormFields / CopyButton, and add a resetAndReload test that asserts pageToken is undefined and hasPrev is false after goNext().

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,332 @@
...
+  formToAgentRegistration,
+  isAgentFormValid,
+} from './agentFormTypes';
+import type { AgentForm } from './agentFormTypes';
+
+const HEALTH_FILTER_OPTIONS: Array<{
+  value: AgentHealthStatus;
+  label: string;
+}> = [
+  { value: 'ready', label: 'Ready' },
```
mareklibra: i18n

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -0,0 +1,332 @@
...
+        field: 'environment',
+        render: a => (
+          <Chip label={a.environment} size="small" variant="outlined" />
+        ),
+      },
+      {
+        title: t('agents.columns.serviceTypes'),
+        field: 'service_types',
+        sorting: false,
+        render: a => {
```
mareklibra, line 184: If service_types is [] the column renders an empty Box. Other tabs use <DcmEmptyCell /> for that. Please add the same fallback before the chip list.

workspaces/dcm/plugins/dcm/src/pages/agents/components/CopyButton.tsx
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+import { useState } from 'react';
+import { IconButton, Tooltip } from '@material-ui/core';
+import FileCopyOutlinedIcon from '@material-ui/icons/FileCopyOutlined';
+import { useTranslation } from '../../../hooks/useTranslation';
+
+type CopyState = 'idle' | 'copied' | 'failed';
+
+export function CopyButton({ text }: Readonly<{ text: string }>) {
```
mareklibra, line 46: This CopyButton no longer swaps the icon on copied/failed (the Providers version used CheckIcon / ErrorOutlineIcon). Tooltip-only feedback is easy to miss. Please restore the icon states or move CopyButton to components/ so every tab shares one implementation.

workspaces/dcm/plugins/dcm/src/pages/agents/components/AgentFormFields.tsx
```diff
@@ -0,0 +1,242 @@
...
+    touched[field] ? errors[field] : undefined;
+
+  return (
+    <Box display="flex" flexDirection="column" gridGap={16}>
+      <TextField
+        label={t('agents.form.nameLabel')}
+        helperText={err('name') ?? t('agents.form.nameHelper')}
+        error={Boolean(err('name'))}
+        value={form.name}
+        onChange={e => setForm(prev => ({ ...prev, name: e.target.value }))}
```
mareklibra: while topic_name is still the generated value, default it to `dcm.agent.${name}` on name change so users do not have to retype the prefix. Stop updating once they edit the topic field.

workspaces/dcm/plugins/dcm/src/hooks/usePaginatedCrudTab.ts
mareklibra: Still uses still uses Provider / listProviders / storageKey: 'providers'.

mareklibra (changes requested):
More comments

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
asmasarw: The empty-state branch now also renders toolbarExtra (with the filter Select) before the action error alert and illustration

workspaces/dcm/plugins/dcm/src/pages/agents/components/AgentFormFields.tsx
```diff
@@ -0,0 +1,258 @@
...
+    setTouched(prev => ({ ...prev, [field]: true }));
+
+  const err = (field: keyof AgentForm) =>
+    touched[field] ? errors[field] : undefined;
+
+  // Track whether the user has manually edited the topic_name field.
+  // While false, typing in the name field auto-fills topic_name to
+  // `dcm.agent.<name>`.  Once the user edits topic_name directly the
+  // auto-fill is disabled for the lifetime of this form instance.
+  const topicManuallyEdited = useRef(false);
```
mareklibra: `topicManuallyEdited` survives dialog close because AgentFormFields stays mounted (DcmFormDialog always renders children) while useCrudTab only resets createForm.

Repro: open Register → type a name (topic auto-fills) → edit topic → Cancel → open Register again → type a name. Topic stays empty.

Please reset the flag when the form is cleared, remount the fields (`key` on open), or derive “still generated” from `topic_name === \`dcm.agent.${name}\`` instead of a ref.

workspaces/dcm/plugins/dcm/src/pages/agents/components/CopyButton.test.tsx
```diff
@@ -52,10 +52,7 @@ describe('CopyButton', () => {
       expect(writeTextMock).toHaveBeenCalledWith('https://example.com'),
     );
 
-    // After success the button title changes — MUI Tooltip sets aria-label on
-    // the element passed to it; we can verify the icon switch via aria-label
-    // on the wrapping Tooltip span via title attribute propagation.
-    // A simpler check: the error icon must NOT be present.
+    // After success the error icon must NOT be present
     expect(screen.queryByTestId('ErrorOutlineIcon')).not.toBeInTheDocument();
```
mareklibra: these cases do not actually assert the icon swap

Success never looks for CheckIcon. Failure is `ErrorOutlineIcon || .MuiSvgIcon-root`, which is true for the idle copy icon too.

Please assert `CheckIcon` after resolve and `ErrorOutlineIcon` after reject (and drop the `.MuiSvgIcon-root` fallback).

workspaces/dcm/plugins/dcm/src/pages/data-center/DataCenterPage.tsx
```diff
@@ -79,8 +79,8 @@ export const DataCenterPage = () => {
       <Content>
         <Box className={classes.tabbedLayout}>
           <TabbedLayout>
-            <TabbedLayout.Route path="/" title={t('page.tabs.providers')}>
-              <ProvidersTabContent />
+            <TabbedLayout.Route path="/" title={t('page.tabs.agents')}>
```
mareklibra, line 82: FLPATH-4773 still says rename SP → Environments and mark resources degraded when the agent is gone. This change uses “Agents” and the Resources column is still “Provider”. If that work is out of scope, please point at the follow-up issue so the story is not silently closed.

workspaces/dcm/plugins/dcm/src/pages/agents/agentFormTypes.test.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  it('returns false when environment is empty', () => {
+    expect(isAgentFormValid({ ...VALID_FORM, environment: '' })).toBe(false);
+  });
+
+  it('returns false when service_types is empty', () => {
+    expect(isAgentFormValid({ ...VALID_FORM, service_types: [] })).toBe(false);
+  });
+
+  it('returns false when cost is empty', () => {
```
mareklibra, line 67: please add `isAgentFormValid({ ...VALID_FORM, cost: 'not-a-cost' as AgentForm['cost'] }) === false` so the oneOf constraint cannot regress.

mareklibra (changes requested):

mareklibra (approved):

PR #4465, opened 2026-08-25
feat(app-defaults): register common icons via IconBundleBlueprint

rohitkrai03:
## Summary
- Register the RHDH common icon catalog on `appDefaultsModule` with `IconBundleBlueprint`, so NFS gets the same icon IDs as the legacy shell.

Resolves RHIDP-16491.

## Test plan
- [ ] `yarn workspace @red-hat-developer-hub/backstage-plugin-app-defaults test`
- [ ] Sample app still loads drawer, template card, and named icons


divyanshiGupta (approved):
LGTM!

PR #4471, opened 2026-08-26
New homepage plugin

debsmita1:

**Resolves:**
https://redhat.atlassian.net/browse/RHDHBUGS-3512

**Solution description:**

Homepage NFS plugin its own page:homepage route so it works without community @backstage/plugin-home, with persona-based homepage.defaultWidgets filtering applied only on that page via homepage-backend.

When homepageHomeModule is installed, the same RH widgets are also registered on community page:home (without RH layout/filtering), so both pages can coexist and be disabled independently via app-config.

https://github.com/user-attachments/assets/b6b8c34b-1bdc-4f41-9e71-8670ecb9df10


christoph-jerolimov (approved):
Thanks @debsmita1 , this is great!!

PR #4474, opened 2026-08-26
fix(scorecard): score first in-window DORA deploys using the prior production deployment

PatAKnight:

First in-window DORA deploy wasn’t getting scored (lead time skipped PRs into it; CFR skipped the gap before it).

We now load the latest successful prod deploy before the 30-day window and use it as the pair/interval start. No prior deploy, same errors as today. DF/MTTR unchanged. Incident after the last in-window deploy still not counted.

GitHub collectors honor fetchItemsLimit (100) on that lookback so we don’t pull 1000 historical rows.

### DIFF
--- a/workspaces/scorecard/.changeset/dora-first-in-window-deploys.md
+++ b/workspaces/scorecard/.changeset/dora-first-in-window-deploys.md
@@ -1,0 +1,6 @@
+---
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-dora': minor
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-github': patch
+---
+
+Include the first in-window successful production deployment in DORA median lead time and change failure rate by using the latest prior successful production deployment as the pair/interval start. GitHub deployment collectors accept an optional fetchItemsLimit so that predecessor lookup stays capped.
--- a/workspaces/scorecard/plugins/scorecard-backend-module-dora/docs/metrics/change-failure-rate.md
+++ b/workspaces/scorecard/plugins/scorecard-backend-module-dora/docs/metrics/change-failure-rate.md
@@ -8,16 +8,28 @@
 Change Failure Rate measures how often production deployments lead to failures that require incident response.
 
 The metric computes the percentage of successful production deployment intervals that contain at least one incident, out of all evaluated successful production deployment intervals.
-Only successful production deployments that happen within the metric's 30-day computation window are evaluated.
+The 30-day window decides which successful production deployments are **in-window**.
 Deployments are processed as chronological pairs (`deployment` -> `nextDeployment`), and each pair defines an interval:
 `[deployment.createdAt, nextDeployment.createdAt)`.
 
 For each interval, if at least one incident has `createdAt` in that interval, the deployment is treated as failed.
 The result is: `(deploymentsWithIncidents / evaluatedDeployments) * 100`.
 
-The metric is **deployment-interval based**, not incident-window based: only incidents that fall between two successful production deployments are scored. An incident after the latest successful production deployment in the 30-day window is not counted in that run, even if it was created within the DORA 30-day window. It is attributed in a later DORA calculation to the interval closed by the next successful production deployment (the first deployment that follows).
+### Pre-window predecessor
 
-If fewer than two successful production deployments exist in the window, or there are no evaluable intervals (adjacent deployments share the same `createdAt`), calculation fails with an error.
+To attribute incidents that fall **before** the first in-window deployment, the metric also loads the latest successful production deployment **before** the window start (`windowFrom`) from the deployments collector.
+
+That predecessor is the start of the interval that **ends** at the first in-window deploy: `[predecessor.createdAt, firstInWindow.createdAt)`.
+When a predecessor exists, incidents are collected from `predecessor.createdAt` (not only from `windowFrom`), so incidents in that leading gap are counted.
+
+- Predecessor exists: one in-window successful production deployment is enough to form that interval.
+- No predecessor: behavior is unchanged. Two successful production deployments in the 30-day window are still required, and incident collection starts at `windowFrom`.
+
+The predecessor lookup is a second deployments-collector call with `from` at `1970-01-01T00:00:00.000Z` and `to` immediately before `windowFrom`, plus optional `fetchItemsLimit` (100). Default GitHub collectors honor that cap and keep the newest in-range rows. If those rows are all failed or non-production, the predecessor is omitted. Custom collectors that ignore the cap may return a larger history.
+
+The metric is **deployment-interval based**, not incident-window based: only incidents that fall between two successful production deployments are scored. An incident **after** the latest successful production deployment in the 30-day window is still not counted in that run, even if it was created within the DORA 30-day window. It is attributed in a later DORA calculation to the interval closed by the next successful production deployment (the first deployment that follows).
+
+If there is no measurable interval (no predecessor and fewer than two successful production deployments in the window, or adjacent deployments share the same `createdAt`), calculation fails with an error.
 
 ## Scope and limitation
 
@@ -90,6 +102,8 @@ Required input:
 - `from: string` (ISO datetime)
 - `to: string` (ISO datetime)
 
+The metric may call this collector twice: once for the 30-day window, and once for the pre-window predecessor range. Extra input fields (for example `fetchItemsLimit`) are allowed; they do not replace `from` / `to`.
+
 Required output:
 
 - `deployments: Array<{ id: string; commitSha: string; environment?: string; createdAt: string; result: 'success' | 'failure' | '' }>`
@@ -128,6 +142,8 @@ Required input:
 
 - `from: string` (ISO datetime)
 - `to: string` (ISO datetime)
+
+When a pre-window predecessor exists, `from` is that deployment's `createdAt` (which can be earlier than the 30-day window start). Otherwise `from` is the window start.
 
 Required output:
 
--- a/workspaces/scorecard/plugins/scorecard-backend-module-dora/src/constants.ts
+++ b/workspaces/scorecard/plugins/scorecard-backend-module-dora/src/constants.ts
@@ -20,3 +20,20 @@ export const DORA_DEFAULT_DEPLOYMENT_PULL_REQUESTS_COLLECTOR_ID =
 export const DORA_DEFAULT_INCIDENTS_COLLECTOR_ID = 'jira:incidents';
 export const DORA_TIME_WINDOW_DAYS = 30;
 export const DORA_DEFAULT_PRODUCTION_ENVIRONMENTS = ['production'];
+
+/**
+ * Max deployments to request for the pre-window predecessor collect.
+ *
+ * We take the latest successful production deploy in that capped batch. That can
+ * miss a real predecessor when the newest pre-window rows are all failed or
+ * non-prod.
+ *
+ * Paging until a successful prod deploy is found would close that gap, but it
+ * is a bad default: a repo with a long staging/failed burst (or a collector
+ * that ignores the cap) would walk unbounded GitHub/Jira history per metric
+ * run — rate limits, latency, and timeout risk on every catalog entity.
+ */
+export const DORA_PRE_WINDOW_DEPLOYMENT_FETCH_CAP = 100;
+
+/** Inclusive collector `from` for unbounded predecessor lookback. */
+export const DORA_PREDECESSOR_COLLECT_FROM = '1970-01-01T00:00:00.000Z';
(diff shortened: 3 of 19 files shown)


dzemanov:
Hi @PatAKnight, thank you for taking this on!
I haven't gone closer through the code yet, but I have a quick question: would it make sense to base this PR on the https://github.com/redhat-developer/rhdh-plugins/pull/4319 DORA database PR and load the last deployment before `windowFrom` from the dora database? It would simplify the code a lot and we would avoid `DORA_PRE_WINDOW_DEPLOYMENT_FETCH_CAP` with `mergeSuccessfulProductionDeploymentsWithPredecessor`.

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/metricProviders/DoraChangeFailureRateProvider.ts
```diff
@@ -153,7 +147,7 @@ export class DoraChangeFailureRateProvider implements MetricProvider<'number'> {
       entity,
       input: {
         ...this.config.incidentsCollector.input,
-        from: from.toISOString(),
+        from: getChangeFailureRateIncidentFrom(from, predecessor),
```
gustavolira, line 150: Widening the incident `from` to `predecessor.createdAt` makes this query unbounded backwards, and the Jira side isn't set up for that. `buildIncidentJql` emits no `ORDER BY`, and `sendPaginatedRequest` stops at `DEFAULT_PAGINATED_FETCH_ITEMS_LIMIT` (1000) with only a warn. For an entity whose predecessor is several months back in a project with frequent incidents, the query can exceed 1000 and the issues that get dropped are whatever Jira happens to page last — possibly the in-window ones. CFR would then report 0%, which lands on the elite threshold, for a service that actually had failing deploys. Could we cap the incident lookback, or add `ORDER BY created DESC` to the JQL so truncation at least drops the oldest?

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/constants.ts
```diff
@@ -20,3 +20,20 @@ export const DORA_DEFAULT_DEPLOYMENT_PULL_REQUESTS_COLLECTOR_ID =
...
+ *
+ * Paging until a successful prod deploy is found would close that gap, but it
+ * is a bad default: a repo with a long staging/failed burst (or a collector
+ * that ignores the cap) would walk unbounded GitHub/Jira history per metric
+ * run — rate limits, latency, and timeout risk on every catalog entity.
+ */
+export const DORA_PRE_WINDOW_DEPLOYMENT_FETCH_CAP = 100;
+
+/** Inclusive collector `from` for unbounded predecessor lookback. */
+export const DORA_PREDECESSOR_COLLECT_FROM = '1970-01-01T00:00:00.000Z';
```
gustavolira, line 39: The comment above says the cap is what keeps us from walking unbounded history per metric run, but with `from` at the epoch it doesn't quite get there. In `GithubClient.getDeployments`, rows newer than `to` are skipped without incrementing `deployments.length`, and `reachedOlderThanWindow` only flips when `deployedAt < fromTimestamp` — which can never happen for epoch. So the predecessor collect pages through every deployment created inside the 30-day window before it reaches a single usable row, and none of those count toward the 100. A repo with 500 deployments in the window pays 5+ extra GraphQL pages per metric, per entity, per run, and both CFR and lead time do it. Would a bounded lookback (something like `windowFrom - 90d`) work instead? The cap bounds rows returned, not requests made.

workspaces/scorecard/plugins/scorecard-backend-module-dora/docs/metrics/change-failure-rate.md
```diff
@@ -8,16 +8,28 @@
...
+### Pre-window predecessor
 
-If fewer than two successful production deployments exist in the window, or there are no evaluable intervals (adjacent deployments share the same `createdAt`), calculation fails with an error.
+To attribute incidents that fall **before** the first in-window deployment, the metric also loads the latest successful production deployment **before** the window start (`windowFrom`) from the deployments collector.
+
+That predecessor is the start of the interval that **ends** at the first in-window deploy: `[predecessor.createdAt, firstInWindow.createdAt)`.
+When a predecessor exists, incidents are collected from `predecessor.createdAt` (not only from `windowFrom`), so incidents in that leading gap are counted.
```
gustavolira, line 23: Since the lookback starts at the epoch, this leading interval has no upper bound on its length. For a repo that deploys quarterly, `[predecessor, firstInWindow)` is ~90 days, and one incident anywhere in there marks it failed — with a single in-window deploy that's the only interval, so CFR reads 100% for the whole month that deploy stays in-window. There's also a shift for entities that already worked: with N in-window deploys the denominator goes from N-1 to N, because the interval CFR scores is keyed on its start, which is now the pre-window predecessor. Neither is called out here, and the metric `description` in `getMetrics()` still tells users this is measured "over the past 30 days". Can we say in Scope and limitation how far back the gap can reach?

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/metricProviders/utils/preWindowDeploymentUtils.ts
```diff
@@ -0,0 +1,221 @@
...
+export function mergeSuccessfulProductionDeploymentsWithPredecessor(options: {
+  inWindowDeployments: Deployment[];
+  preWindowDeployments: Deployment[];
+  windowFrom: Date;
+  productionEnvironments: string[];
+}): DeploymentsWithPreWindowBoundary {
+  const { windowFrom, productionEnvironments } = options;
+  const windowFromTimestamp = windowFrom.getTime();
+
+  const inWindow = options.inWindowDeployments.filter(deployment =>
```
gustavolira, line 83: The predecessor side clips on time (`createdAtTimestamp >= windowFromTimestamp` -> skip, in `findLatestSuccessfulProductionPredecessor`), but this side filters only on success/environment — nothing checks that in-window rows are actually at or after `windowFrom`. `[predecessor, ...inWindow]` is then used without a re-sort, and the ascending `superRefine` in `deploymentsCollectorOutputSchema` only runs per-collect, so it never sees the merged array. `github:deployments` is safe because `getDeployments` re-checks the range client-side, but `getWorkflowRuns` trusts GitHub's `created` filter and pushes back whatever it gets, and custom collectors are explicitly supported. If one row lands before the predecessor, CFR drops that interval from both numerator and denominator (`nextDeploymentCreatedAt <= deploymentCreatedAt` -> continue) and lead time ends up comparing base/head in the wrong direction. A `.filter(d => new Date(d.createdAt).getTime() >= windowFromTimestamp)` here would make the two sides symmetric.

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/metricProviders/utils/preWindowDeploymentUtils.ts
```diff
@@ -0,0 +1,221 @@
...
+      entity: options.entity,
+      input: {
+        ...options.collectorInput,
+        from: predecessorRange.from,
+        to: predecessorRange.to,
+        fetchItemsLimit: predecessorRange.fetchItemsLimit,
+      },
+    });
+    preWindowDeployments = preWindowCollected.deployments;
+  } catch (error) {
```
gustavolira, line 204: This catch is wider than the predecessor lookup. `DefaultScorecardCollectorsService.collect` funnels auth failures, rate limits, GitHub 5xx and input-schema rejections into the same `Error`, so they all land here as one warn. Two cases stand out: an entity with exactly one in-window deploy hits a transient rate limit and the operator sees "need at least 2 successful production deployments... found 1", which points at their deployment data rather than at the API; and a custom deployments collector with a `.strict()` input schema rejects the extra `fetchItemsLimit` key on every run, so the feature is silently off forever with only this warn. Should the message say the predecessor lookup failed, so the two are distinguishable in the logs?

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/metricProviders/DoraChangeFailureRateProvider.test.ts
```diff
@@ -448,5 +450,130 @@ describe('DoraChangeFailureRateProvider', () => {
         expect.stringContaining('non-increasing createdAt'),
       );
     });
+
+    it('should count an incident between a predecessor and the first in-window deployment', async () => {
```
gustavolira, line 454: This asserts the incidents `from`, which is good, but nothing anywhere asserts what the predecessor deployments call actually sends. The mock collector only branches on `input.from === DORA_PREDECESSOR_COLLECT_FROM`, so dropping `fetchItemsLimit: predecessorRange.fetchItemsLimit` from the input in `preWindowDeploymentUtils` would leave the whole suite green while production starts asking for 1000 deployments since 1970 on every DORA run. Could we add a `toHaveBeenCalledWith` on the second deployments call covering `to` and `fetchItemsLimit`? The catch/fallback path in `collectSuccessfulProductionDeploymentsWithPreWindowBoundary` has no test either — `mockResolvedValueOnce(...).mockRejectedValueOnce(new Error('rate limited'))` would cover it.

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/metricProviders/DoraChangeFailureRateProvider.ts
```diff
@@ -110,34 +110,28 @@ export class DoraChangeFailureRateProvider implements MetricProvider<'number'> {
...
-    const successfulProductionDeployments =
-      deploymentsCollected.deployments.filter(deployment =>
-        isSuccessfulProductionDeployment(
-          deployment,
-          this.config.productionEnvironments,
-        ),
-      );
-
-    if (successfulProductionDeployments.length < 2) {
+    if (inWindow.length === 0 || successfulProductionDeployments.length < 2) {
```
gustavolira, line 129: nit: `inWindow.length === 0` can't be the sole trigger here — `collectSuccessfulProductionDeploymentsWithPreWindowBoundary` returns early with `deployments: []` whenever `inWindow` is empty, so `deployments.length < 2` already covers that case. Same disjunct in the lead time provider. Since the guard is duplicated at both call sites, would it be better living in the util?

workspaces/scorecard/.changeset/dora-first-in-window-deploys.md
```diff
@@ -0,0 +1,6 @@
+---
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-dora': minor
+'@red-hat-developer-hub/backstage-plugin-scorecard-backend-module-github': patch
```
gustavolira, line 3: nit: should the github module be `minor` rather than `patch`? It gains a new public collector input field that users can set from `app-config` and that's documented in the README. Recent changesets in this workspace use `minor` for that kind of addition (`threshold-annotation-controls.md`, `metric-collectors-endpoint.md`). `report.api.md` doesn't move here since the collectors aren't exported from `src/index.ts`, so the changeset is the only place the new field surfaces.

workspaces/scorecard/plugins/scorecard-backend-module-github/src/collectors/schemas/deploymentSchemas.ts
```diff
@@ -31,3 +32,10 @@ export type DeploymentResult = z.infer<typeof deploymentResultSchema>;
 export const deploymentsSchema = z.object({
   deployments: z.array(deploymentSchema),
 });
+
+export const fetchItemsLimitSchema = z
+  .number()
+  .int()
+  .positive()
+  .max(DEFAULT_DEPLOYMENT_FETCH_ITEMS_LIMIT)
```
gustavolira, line 40: nit: tying the max to `DEFAULT_DEPLOYMENT_FETCH_ITEMS_LIMIT` couples two things that move for different reasons — lowering the default later would turn previously valid `fetchItemsLimit` config into a hard input-validation error that fails the whole metric, and the README documents 1–1000 as if it were stable. A separate `MAX_DEPLOYMENT_FETCH_ITEMS_LIMIT` would keep them independent.

PatAKnight:
> Hi @PatAKnight, thank you for taking this on! I haven't gone closer through the code yet, but I have a quick question: would it make sense to base this PR on the #4319 DORA database PR and load the last deployment before `windowFrom` from the dora database? It would simplify the code a lot and we would avoid `DORA_PRE_WINDOW_DEPLOYMENT_FETCH_CAP` with `mergeSuccessfulProductionDeploymentsWithPredecessor`.

@dzemanov, yep, can do!

PatAKnight:
There are a lot of merge conflicts, so I am going to close this in favor of https://github.com/redhat-developer/rhdh-plugins/pull/4732

PR #4484, opened 2026-08-27
test(bulk-import): cover the NFS wiring, not just the translations module

gustavolira:
`nfsExports.test.ts` asserts only that the translations module exists and is exported as default. That module is a separate `FrontendModule` with `pluginId: 'app'`, so both assertions hold against a bulk-import plugin whose **own** `extensions` array is empty — the page and the API disappear from the app with no error, no console warning, exit 0.

Same shape as #4470 for adoption-insights, which merged earlier today.

## What's added

Three tests reading what the app actually resolves off the extensions:

- the page's `routePath` and `title`
- that the page **and** the API extension are registered on the plugin
- that both route refs survive, including the `tasks` subRouteRef

## Mutation-verified

Each mutation applied to `src/index.tsx`, suite re-run, source restored:

| Mutation | Result |
|---|---|
| page dropped from `extensions` | 2 failed |
| api dropped from `extensions` | 1 failed |
| `path: '/bulk-import'` changed | 1 failed |
| `title: 'Bulk import'` changed | 1 failed |
| `routeRef: rootRouteRef` removed | 1 failed |
| `tasks: importHistoryRouteRef` removed | 1 failed |

Six for six, and the counts separate the tests: dropping the page fails two, dropping the api fails only the registration one, dropping the subroute fails only the routes one.

## One finding worth carrying to the next plugin

The adoption-insights test asserts `toBe(rootRouteRef)` on the route ref and passes. **The same assertion fails here.**

`createFrontendPlugin` re-wraps route refs — probed it, and the object the plugin holds carries neither `id` nor `params`, so it is not the one `routes.ts` exported:

    imported  : {"id":"bulk-import","params":[],"$$type":"@backstage/RouteRef","version":"v1"}
    on plugin : {"$$type":"@backstage/RouteRef","version":"v1"}

Identity happens to survive for a plugin with a single route and does not for one that also declares a subRouteRef. That is a property of the wrapping, not of the plugin being correct, so this test asserts presence instead — which the mutation table above confirms still catches removal.

Flagging it because [RHIDP-16455](https://redhat.atlassian.net/browse/RHIDP-16455) has ten more plugins to go and `toBe` would look like the established pattern.

## Verification

From `workspaces/bulk-import`: `yarn tsc`, `yarn lint` and `yarn prettier --check` all clean; the suite runs 5 tests.

`@backstage/frontend-test-utils` is added at `^0.6.1`, the range four other workspaces use at this manifest version.

Part of [RHIDP-16455](https://redhat.atlassian.net/browse/RHIDP-16455).

### DIFF
--- a/workspaces/bulk-import/.changeset/olive-donuts-shake.md
+++ b/workspaces/bulk-import/.changeset/olive-donuts-shake.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-bulk-import': patch
+---
+
+Cover the new frontend system wiring with createExtensionTester: the page's path and title, that the page and the API extension are registered on the plugin, and that both route refs survive. The existing NFS test only covers the translations module, which is a separate FrontendModule — so it stays green against a plugin whose own extensions array is empty.
--- a/workspaces/bulk-import/plugins/bulk-import/package.json
+++ b/workspaces/bulk-import/plugins/bulk-import/package.json
@@ -89,6 +89,7 @@
     "@backstage/core-app-api": "^1.20.4",
     "@backstage/dev-utils": "^1.1.26",
     "@backstage/frontend-defaults": "^0.5.5",
+    "@backstage/frontend-test-utils": "^0.6.3",
     "@backstage/test-utils": "^1.7.21",
     "@backstage/ui": "^0.17.1",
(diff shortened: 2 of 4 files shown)


PatAKnight (approved):
LGTM

PR #4488, opened 2026-08-30
feat(x2a): implement cascading invalidation

yray-pixel:
When a phase completes successfully, any finished jobs for downstream phases are automatically marked as stale.


eloycoto:
@mareklibra can you check this PR about the migration, LGTM, but I would like to make sure? and also the MCP server, I do not think that we need anything here, but it's good to check.

workspaces/x2a/plugins/x2a-backend/migrations/2026081900_expand_jobs_constraints.ts
```diff
@@ -0,0 +1,141 @@
...
+  statuses: string[],
+): void {
+  table.uuid('id').primary();
+  table.text('log');
+  table.timestamp('started_at').notNullable();
+  table.timestamp('finished_at');
+  table.string('status').notNullable().defaultTo('pending').checkIn(statuses);
+  table.string('phase').notNullable().defaultTo('init').checkIn(PHASES);
+  table
+    .uuid('project_id')
```
mareklibra, line 56: SQLite recreate copies jobs with INSERT … SELECT *, but createJobsTable column order no longer matches the live table (project_id/module_id were moved before error_details in 806b605). That maps error_details → project_id (NOT NULL). Success rows with NULL error_details will fail the migration; other rows will be silently corrupted. Restore the adversarial-migration column order, or use a named-column INSERT. Please add a test that inserts a job, runs this up(), and asserts project_id is unchanged.

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
```diff
@@ -314,6 +328,44 @@ export function registerCollectArtifactsRoutes(
...
+  for (const downstreamPhase of downstreamPhases) {
+    const jobs = await x2aDatabase.listJobs({
+      projectId,
+      moduleId,
+      phase: downstreamPhase,
+      lastJobOnly: true,
+    });
+
+    const lastJob = jobs[0];
+    if (lastJob && JobStatus.from(lastJob.status).isSuccess()) {
```
mareklibra, line 358: After this invalidation, getNextPhase / getLastPhaseReached (not updated in this PR) still treat a stale migrate job as “the last phase reached”. For a module waiting on publish, re-running analyze then makes bulk “run next” call publish using stale migrate artifacts. Skip stale the same way cancelled is skipped so the next phase is the first stale/unfinished one. Please add getNextPhase tests for migrate stale → migrate, and migrate+publish stale → migrate.

workspaces/x2a/plugins/x2a-node/src/moduleStatus.ts
```diff
@@ -40,19 +40,18 @@ export function calculateModuleStatus({
...
   }
-  if (analyze) {
-    return { status: analyze.status, errorDetails: analyze.errorDetails };
+
+  if (JobStatus.from(latestPhaseJob.status).isCancelled()) {
+    return { status: 'cancelled', errorDetails: undefined };
   }
 
-  return { status: 'pending', errorDetails: undefined };
+  const status: ModuleStatus = JobStatus.from(latestPhaseJob.status).isStale()
```
mareklibra: Mapping stale → success hides the new state at module/project level: ModuleStatusCell still shows the Published chip whenever module.publish exists, and calculateProjectStatus counts the module in neither finished nor waiting. module.statuses.stale translations added in this PR are unused. Please either add stale to ModuleStatus and return it, or compute aggregate status from the earliest stale downstream phase. moduleStatus.test.ts needs cases for this.

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
```diff
@@ -355,6 +407,22 @@ async function processJobCompletion(
...
+  if (jobStatus.isSuccess()) {
+    try {
+      await invalidateDownstreamPhases(
+        phase,
+        projectId,
+        job.moduleId,
+        x2aDatabase,
+        logger,
+      );
+    } catch (err) {
```
mareklibra: Invalidation runs after updateJob has already committed success and errors are only logged. If markJobsAsStale fails, downstream jobs stay success and the UI never shows stale. Please run both updates in one transaction or fail the collectArtifacts callback if invalidation fails so the operator can retry.

workspaces/x2a/plugins/x2a-node/src/moduleStatus.ts
mareklibra: moduleStatus.ts now maps stale → success with no new tests, and getNextPhase / ModuleStatusCell / projectStatus / PhasesCard adversarial canRun were not updated for the new status. Please add moduleStatus and getNextPhase cases, plus a collectArtifacts test that an errored downstream job is not marked stale (the 806b605 isSuccess filter).

workspaces/x2a/.changeset/large-cloths-bow.md
```diff
@@ -0,0 +1,8 @@
+---
+'@red-hat-developer-hub/backstage-plugin-x2a-backend': patch
+'@red-hat-developer-hub/backstage-plugin-x2a-common': patch
+'@red-hat-developer-hub/backstage-plugin-x2a-node': patch
+'@red-hat-developer-hub/backstage-plugin-x2a': patch
+---
+
+invaldating stale jobs - jobs that had a former phase ran again
```
mareklibra: Please use a changelog-ready sentence, e.g. “Mark downstream phase jobs as stale when an upstream phase is re-run successfully.”

workspaces/x2a/plugins/x2a-backend/migrations/2026081900_expand_jobs_constraints.ts
```diff
@@ -0,0 +1,141 @@
...
+  const client = knex.client.config.client;
+
+  if (client === 'better-sqlite3') {
+    await recreateJobsTableSqlite(knex, ORIGINAL_STATUSES);
+  } else {
+    await knex.raw(
+      `ALTER TABLE jobs DROP CONSTRAINT IF EXISTS jobs_status_check`,
+    );
+    await knex.raw(
+      `ALTER TABLE jobs ADD CONSTRAINT jobs_status_check CHECK (status IN ('pending', 'running', 'success', 'error', 'cancelled'))`,
```
mareklibra, line 152: down() re-applies a CHECK without 'stale'. On Postgres this fails if any row is already stale; on SQLite INSERT into the restored table will fail the same way. Please convert stale rows first, or document that down is unsupported after this status has been used.

workspaces/x2a/plugins/x2a/src/components/PhaseDetails.tsx
```diff
@@ -77,11 +77,15 @@ const PhaseRunAction = ({
 
   const previousRunSucceeded =
     !!phase?.status && JobStatus.from(phase.status).isSuccess();
+  const isStale = !!phase?.status && JobStatus.from(phase.status).isStale();
```
mareklibra, line 78: previousRunSucceeded is isSuccess() only, so a stale phase gets the first-run button label (“Run migrate”) plus staleInstructions. Consider treating stale as a previous run (isSuccess() || isStale()) so the CTA stays “Re-run …”.

mareklibra (changes requested):

workspaces/x2a/plugins/x2a/src/components/PhaseDetails.tsx
yray-pixel: Fixed

workspaces/x2a/plugins/x2a-backend/migrations/2026081900_expand_jobs_constraints.ts
yray-pixel: Added a conversion before altering the tables

workspaces/x2a/.changeset/large-cloths-bow.md
yray-pixel: Rewrote the changeset

workspaces/x2a/plugins/x2a-node/src/moduleStatus.ts
yray-pixel: Fixed. calculateModuleStatus now returns 'stale' directly.
Added tests  verifying errored downstream jobs are not marked stale.

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
yray-pixel: Removed the try/catch to let it fail.

workspaces/x2a/plugins/x2a-node/src/moduleStatus.ts
yray-pixel: added stale to moduleStatus

workspaces/x2a/plugins/x2a-backend/src/router/collectArtifacts.ts
yray-pixel: Fixed by extending the ignoreCancelled skip in getLastPhaseReached to also skip stale.
Also added the tests

workspaces/x2a/plugins/x2a-backend/migrations/2026081900_expand_jobs_constraints.ts
yray-pixel: Change to named-column INSERT.
Also added a test at ./plugins/x2abackend/src/services/X2ADatabaseService/2026081900_expand_jobs_constraints.test.ts
not sure if this is the correct place for it

mareklibra (approved):

PR #4504, opened 2026-08-31
fix(extensions): show restart alert after plugin toggle in NFS

ciiay:
## Description

After enabling or disabling a plugin from the Actions menu side drawer on the Extensions catalog page, NFS now stays on the catalog tab so the **Backend restart required** alert still appears.

`handleToggle` previously navigated to `/extensions/installed-plugins`, which is not a registered tab (`/extensions/catalog` and `/extensions/installed-packages` are). In NFS that unregistered path can unmount the extensions page, so the restart alert never appears. OFS still showed catalog content via TabbedLayout fallback, which is why it looked fine there.

Pending-restart state also lived only in React `useState`, so a remount cleared it. It is now kept on `globalThis` so it survives remounts (cleared on a full page reload after a real backend restart).

## Fixed
- [RHDHBUGS-3701](https://redhat.atlassian.net/browse/RHDHBUGS-3701) — In nfs after enabling a plugin, the expected restart alert doesn't appear.in extensions catalog page

## UI before changes
attached file in [RHDHBUGS-3701](https://redhat.atlassian.net/browse/RHDHBUGS-3701)
## UI after changes
After fix

## Test Plan
- [ ] Open the Extensions catalog in NFS
- [ ] Open a plugin Actions drawer and disable or enable the plugin
- [ ] Confirm the app lands on `/extensions/catalog`
- [ ] Confirm the **Backend restart required** alert is visible
- [ ] Repeat the same flow in OFS/legacy and confirm the alert still appears

## Checklist

### DIFF
--- a/workspaces/extensions/.changeset/quiet-harbor-rest.md
+++ b/workspaces/extensions/.changeset/quiet-harbor-rest.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-extensions': patch
+---
+
+Navigate to the catalog tab after enable/disable so the NFS restart alert still appears.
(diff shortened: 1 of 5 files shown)


debsmita1 (approved):
Tested the PR on rhdh-local

Image

/lgtm

PR #4573, opened 2026-09-03
feat(scorecard): add entity-page sparkline charts for time-series metrics

Eswaraiahsapram:
## What
Adds sparkline (area chart) visualization support for entity-page scorecard metrics whose `defaultVisualization` is `sparkline`. This is the foundation PR — shared chart components and utilities are included here and will be reused by the homepage sparkline PR that follows.

### New components
- **`SparklineChart`** — Recharts-based area chart with gradient fill, error-dot markers, hover tooltip, and threshold legend
- **`SparklineTooltip`** / **`SparklineLegend`** — supporting chart sub-components
- **`EntitySparklineCard`** — entity-page card that fetches time-series data and renders a sparkline with a "View data sources" dialog for collector metadata
- **`EntityMetricCard`** — routing component that renders `EntitySparklineCard` or the existing `Scorecard` card based on `defaultVisualization`

### New API methods
- `getMetricTimeSeries` — `GET /metrics/catalog/:kind/:namespace/:name/time-series`
- `getMetricCollectors` — `GET /metrics/:metricId/collectors`

### New hooks
- `useMetricTimeSeries` — `useQuery`-based hook for 30-day entity metric time series
- `useMetricCollectors` — `useQuery`-based hook for collector metadata (fetched only when the data-sources dialog is open)

### New utilities
- `timeSeriesChartData` — maps API points to chart-ready data with interpolation for error gaps
- `sparklineLegend` — builds threshold legend items with color + line-style pairing
- `sparklineChartModel` — shared view-model factory used by both entity and homepage cards
- `metricVisualization` — `isSparklineVisualization()` helper
- `timeSeriesRange` — computes the default 30-day ISO-8601 range

### Refactors
- **`DataSourcesDialog`** now accepts generic `SourceRow[]` instead of building rows internally
- Extracted `collectorSourceRows.ts` (for sparkline metrics) and `metricSourceRows.ts` (for existing donut metrics) as separate row builders
- All collector labels (`GitHub`, `Jira`, empty value `--`, unavailable status `N/A`) are now translated via i18n keys instead of hardcoded strings

### Translations
- Added 6 new `dataSourcesDialog.*` keys to `ref.ts` and all locale files (de, es, fr, it, ja)

### Screen Recording
https://github.com/user-attachments/assets/6a939609-9ab1-485f-b351-41f4e0edb01a

## How to test
1. Configure a catalog entity with DORA metric providers (or any metric with `defaultVisualization: sparkline`)
2. Navigate to the entity's Scorecard tab
3. Verify the sparkline chart renders with a 30-day trend line
4. Click the menu → "View data sources" and verify collectors are listed
5. Error days should show red dots on the chart with tooltip messages


workspaces/scorecard/plugins/scorecard/src/api/index.ts
```diff
@@ -364,4 +369,185 @@ export class ScorecardApiClient implements ScorecardApi {
...
+
+      const data = await response.json();
+
+      if (
+        !data ||
+        Array.isArray(data) ||
+        typeof data !== 'object' ||
+        typeof data.id !== 'string' ||
+        typeof data.metricId !== 'string' ||
+        !Array.isArray(data.points)
```
dzemanov: Should we validate all returned data? E.g. also points structure?

workspaces/scorecard/plugins/scorecard/src/api/index.ts
```diff
@@ -364,4 +369,185 @@ export class ScorecardApiClient implements ScorecardApi {
...
+
+      const data = await response.json();
+
+      if (
+        !data ||
+        Array.isArray(data) ||
+        typeof data !== 'object' ||
+        typeof data.metricId !== 'string' ||
+        typeof data.entityRef !== 'string' ||
+        !Array.isArray(data.points)
```
dzemanov: Should we validate all returned data? E.g. also points structure?

workspaces/scorecard/plugins/scorecard/src/hooks/useMetricTimeSeries.tsx
```diff
@@ -0,0 +1,87 @@
...
+ */
+export const useMetricTimeSeries = (
+  metricId: string,
+): UseResponseData<MetricTimeSeriesResponse> => {
+  const { entity } = useEntity();
+  const scorecardApi = useApi(scorecardApiRef);
+  const { t } = useTranslation();
+
+  const hasEntity = Boolean(
+    entity?.kind && entity?.metadata?.namespace && entity?.metadata?.name,
```
dzemanov: Minor: namespace is not needed for `stringifyEntityRef`

dzemanov (changes requested):
Thank you for the PR!

Tested for data with both successes and errors (both at the start / end or in the middle):

https://github.com/user-attachments/assets/043278ca-df0b-415e-aeba-0c5a056fd77e

I am not sure if I have misconfigured something, but I see only the resulting threshold in legend.
I see we are not providing thresholds in `time-series` response, we might need to update that. Meanwhile you can skip fixing it or you can take if from metric result.

When I have only 1 data point:

Image
What could help:
The frontend should display data points for all 30 days, even if data is missing, indicating absence with "No data recorded for this date."

Only 2 data points:

Image

Translation:

Image

Collector description might need to also be provided:

Image

For development, I tried running:
`yarn start` in the plugins folder, but I did not see new sparkline cards:

Image
This is not blocking, but worth fixing / adding different cases for future to avoid seeding database for some of the testing.


**Group cards work nice:**

https://github.com/user-attachments/assets/7e60ed0d-e37a-4131-aff1-454481dde53c

There is `.` missing in median lead time for changes description, I will update it.

@ShiranHi, is View datasources description alright in this state?

This is description for group card:

Image

For individual dora cards, 'View datasources' shows:

Image

Image

For individual "View Datasources" DORA cards, should we also include both the metric description and the collector description? It would make the metrics easier to understand, but it would mean repeating the same information on each row.
Incorporating the metric name and description directly into the collector description is not possible due to the collector's use across multiple metrics.

Current descriptions:
`github:deployments`: Collects GitHub deployments.
`github:deploymentWorkflowRuns`: Collects deployments from GitHub Actions.
`github:deploymentPullRequests`: Collects pull requests linked to deployments.
`jira:incidents` Collects Jira incidents.

ciiay (changes requested):
Thanks for the PR — the shared chart stack, `DataSourcesDialog` refactor, and test coverage are in good shape. I agree with @dzemanov's manual testing feedback and am also requesting changes before merge.

## Blocking

1. **Sparse / missing days in the 30-day window** — The chart only plots API-returned points. With 1–2 data points the sparkline is misleading (flat line, no meaningful x-axis). Please fill the requested 30-day range on the frontend and mark missing days in the tooltip (e.g. "No data recorded for this date") rather than reusing `errors.metricDataUnavailable`.

2. **Threshold legend** — `EntitySparklineCard` currently passes only the matched rule (`legendRules: matchedRule ? [matchedRule] : undefined`), so users see a single band. Since the time-series response does not include thresholds yet, please derive the full legend from `metric.result.thresholdResult.definition.rules` while keeping line color/style tied to the current evaluation.

3. **Single-point UX** — When there is only one tick, `SparklineChart` hides x-axis labels (`xTicks.length <= 1`). Please show at least one date label or an explicit insufficient-history state.

4. **i18n** — `dataSourcesDialog.collectorStatusTooltip` hardcodes "DORA" in all locales but applies to any collector-backed metric. Please generalize (interpolation or neutral wording). A native-speaker pass on the new German strings would also help.

## Should address (non-blocking but recommended)

- **SonarCloud** — Quality gate fails on ~3.9% duplication in new code; likely duplicated mock time-series code in `dev/mocks.ts` and `dev/legacy.tsx` — extract a shared fixture.
- **`useMetricCollectors`** — Align error handling with `useMetricTimeSeries` (translated fetch errors in `queryFn`).
- **API response validation** — Consider validating `points[]` entry shape, not just top-level object checks (see inline threads).
- **`getAggregationTimeSeries`** — Fine to land here for the homepage follow-up PR, but worth a brief note in the PR description since it has no caller yet.
- **Data sources dialog content** — Open question from @dzemanov on whether individual DORA cards should include metric description alongside collector description; needs product sign-off.

Happy to re-review once the sparse-data and legend items are addressed.

workspaces/scorecard/plugins/scorecard/src/api/index.ts
ciiay: Agree — top-level shape checks are not enough once we rely on sparklines for daily data. At minimum validate each point has a string `timestamp` and that `value` is number | null (plus optional `error` for metric series). Same pattern would help on `getMetricCollectors` entries, which already has `isCollectorMetadata`.

workspaces/scorecard/plugins/scorecard/src/api/types.ts
```diff
@@ -86,4 +102,32 @@ export interface ScorecardApi {
    * @throws Error if the request fails or returns invalid data
    */
   getAggregationMetadata(aggregationId: string): Promise<AggregationMetadata>;
+
```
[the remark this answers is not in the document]
ciiay: Looks addressed in the latest commits — both `dev/mocks.ts` and `dev/legacy.tsx` now stub the three new API methods. 👍

workspaces/scorecard/plugins/scorecard/src/api/index.ts
ciiay: +1 here as well. Invalid/malformed points currently become confusing chart states instead of a clear API error. Worth sharing a small validator helper between aggregation and metric time-series paths.

workspaces/scorecard/plugins/scorecard/src/api/index.ts
```diff
@@ -364,4 +369,185 @@ export class ScorecardApiClient implements ScorecardApi {
...
+      `${baseUrl}/metrics/catalog/${entity.kind}/${entity.metadata.namespace}/${entity.metadata.name}/time-series`,
+    );
+    url.searchParams.set('metricId', metricId);
+    url.searchParams.set('from', from);
+    url.searchParams.set('to', to);
+
+    try {
+      const response = await this.fetchApi.fetch(url.toString());
+
+      if (!response.ok) {
```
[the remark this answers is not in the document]
ciiay: Minor, but agree for consistency — switching `getMetricCollectors` to `new URL()` like the other client methods would match the existing pattern.

workspaces/scorecard/plugins/scorecard/src/hooks/useMetricTimeSeries.tsx
ciiay: Good catch — `hasEntity` already guarantees namespace/name, so the extra namespace check in the guard is redundant. Fine to simplify when touching this file.

workspaces/scorecard/plugins/scorecard/src/translations/ref.ts
```diff
@@ -216,6 +216,12 @@ export const scorecardMessages = {
     unknownPlugin: 'Unknown',
     statusTooltip:
       'Value {{value}} matches threshold {{status}} {{expression}}',
+    collectorStatusTooltip:
+      'This collector provides input data only. The DORA metric value is calculated from collectors and shown on the scorecard card.',
+    collectorEmptyValue: '--',
+    collectorUnavailableStatus: 'N/A',
```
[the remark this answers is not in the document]
ciiay: Agree — this will be confusing for non-DORA sparkline metrics. Please replace the hardcoded "DORA" copy with a generic message or interpolate the metric title/name.

workspaces/scorecard/plugins/scorecard/src/hooks/useMetricCollectors.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { UseResponseData } from './types';
+
+/**
+ * Fetches collector metadata for a metric. Pass `enabled` so the request runs
+ * only when the data-sources dialog is open and the metric has collector IDs.
+ */
+export const useMetricCollectors = (
+  metricId: string,
+  enabled: boolean,
+): UseResponseData<CollectorMetadata[]> => {
```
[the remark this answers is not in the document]
ciiay: +1 — please mirror the `useMetricTimeSeries` try/catch + translated `errors.fetchError` wrapper so collector fetch failures surface consistently in the dialog.

workspaces/scorecard/plugins/scorecard/src/utils/timeSeriesChartData.ts
```diff
@@ -0,0 +1,172 @@
...
+    error:
+      point.status === 'error'
+        ? formatAggregatedTimeSeriesErrors(point.errors) ?? fallbackErrorLabel
+        : undefined,
+  }));
+
+/**
+ * Maps API time-series points into chart rows. Error / null values keep their
+ * x position and get an interpolated Y so the sparkline stays continuous.
+ */
```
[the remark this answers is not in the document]
ciiay: Agree — `toMetricSparklinePoints` already treats `value === null` as an error day when no message is present; `toAggregationSparklinePoints` should apply the same fallback for parity.

Eswaraiahsapram:
Thanks a lot @dzemanov , for reviewing this 🙏 

> I am not sure if I have misconfigured something, but I see only the resulting threshold in legend. I see we are not providing thresholds in `time-series` response, we might need to update that. Meanwhile you can skip fixing it or you can take if from metric result.

@dzemanov , this is intentional for the **entity page**. As per [prototype](https://scorecards-1152db.pages.redhat.com/catalog/default/component/billing-api/scorecards), the legend shows only the **matched / current** threshold (e.g. Medium), and the line color/style follow that evaluation.

For **homepage** aggregation cards we will show all threshold bands in the legend. That is in the homepage follow-up PR.


> 
> When I have only 1 data point:
> 
> Image What could help: The frontend should display data points for all 30 days, even if data is missing, indicating absence with "No data recorded for this date."

Thanks @dzemanov, we can check with UX before changing it.

@ShiranHi, Right now, with a single sample we plot that point as-is (dot, no filled 30-day line). That matches the prototype

Filling all 30 days and showing “No data recorded for this date.” on the gaps would make the trend easier to read, but it is also a different look from the prototype (lots of empty/missing days vs one point).

@ShiranHi, which should we follow for the entity sparkline?
1. Keep the prototype: only plot days the API returns (1–2 points stay as dots).
2. Always show the 30-day window and mark missing days in the tooltip as “No data recorded for this date.”

ShiranHi:
> Thanks @dzemanov, we can check with UX before changing it.
> 
> @ShiranHi, Right now, with a single sample we plot that point as-is (dot, no filled 30-day line). That matches the prototype
> 
> Filling all 30 days and showing “No data recorded for this date.” on the gaps would make the trend easier to read, but it is also a different look from the prototype (lots of empty/missing days vs one point).
> 
> @ShiranHi, which should we follow for the entity sparkline?
> 
> 1. Keep the prototype: only plot days the API returns (1–2 points stay as dots).
> 2. Always show the 30-day window and mark missing days in the tooltip as “No data recorded for this date.”

There are two topics here:
1. **Threshold legend:** The legend should show all threshold ranges in all use cases. The current behavior, where only the current threshold is shown, is not intentional. I’ll update it so the full legend is displayed.
2. **Missing data**: The prototype uses two states. When there are enough data points, the chart shows the 30-day trend, as in the Deployment Frequency card. When there is only one data point, it intentionally shows a single dot with its date and value, as in the Change Failure Rate card. I propose keeping this behavior for the entity page, since we don’t have enough information to draw a trend line for missing dates.

ciiay:
Thanks for the follow-up commits — re-checked the latest head (`2dec824`).

**Looks good**
- Full threshold legend on entity cards (aligns with @ShiranHi’s UX direction)
- API time-series point validation
- Dev mock/fixture refactor (Sonar gate is green now)
- Tooltip kept inside the card

**Question before merge**
- **`collectorStatusTooltip` still hardcodes "DORA"** in `ref.ts` and all locales, but the tooltip applies to any collector-backed metric. Shouldn't we generalize that copy (or interpolate the metric name) before merge?

**Minor (non-blocking?)**
- Single-point charts still hide the x-axis label when there is only one tick — OK to leave as-is per prototype, or worth showing at least one date on the dot?
- `useMetricCollectors` still lacks the translated error wrapper used in `useMetricTimeSeries`
- Data sources dialog content (metric vs collector description) — still waiting on product sign-off?

Happy to re-review once the i18n tooltip is addressed.

dzemanov:
Hi @Eswaraiahsapram, @ciiay
it would be nice to merge this PR to have some time to work on https://github.com/redhat-developer/rhdh-plugins/pull/4596.

If there are some problems with fixing tooltip translation, we can remove `collectorStatusTooltip` translation that hardcodes "DORA" from this PR and instead create a bug for it, together with bug for translating check description of collectors in  'View datasources'.

dzemanov:
Is should be possible to pass param of metric id to general tooltip description.
Then, tooltip description would change from:
```
This collector provides input data only. The DORA metric value is calculated from collectors and shown on the scorecard card.
```
To:
```
This collector provides input data only. The dora.deploymentFrequency check value is calculated from collectors and shown on the scorecard card.
```

We could also pass `dora` (plugin id) to the tooltip but I think it makes sense to use the concrete metric id in it instead, as this can be then reused by other composite metrics.
Maybe we also want to avoid speaking about `metric`, as in frontend, this is called `check`.

cc @ShiranHi

workspaces/scorecard/plugins/scorecard/src/components/Scorecard/EntitySparklineCard.tsx
```diff
@@ -0,0 +1,246 @@
...
+      lastSynced: metric.result?.timestamp
+        ? getLastUpdatedLabel(metric.result.timestamp, locale)
+        : MISSING_EVALUATION_LABEL,
+      unknownPlugin: t('dataSourcesDialog.unknownPlugin'),
+      emptyValue: t('dataSourcesDialog.collectorEmptyValue'),
+      unavailableStatus: t('dataSourcesDialog.collectorUnavailableStatus'),
+      pluginLabels: {
+        github: t('dataSourcesDialog.pluginGithub'),
+        jira: t('dataSourcesDialog.pluginJira'),
+      },
```
dzemanov: For future fix PR, pluginLabels should not be hardcoded to only Jira and github and any other plugins we add, as we don't need translations for plugin IDs and return for all that are not translated `Unknown`.
We could directly get it from collector id: `pluginId:collecorName`, just uppercase the first letter.
We can have translation for github though, as it is `GitHub` instead of `Github`

dzemanov: Or better, github should be only hardcoded instead of using translation, as `GitHub` is the same in every language.

workspaces/scorecard/plugins/scorecard/src/translations/de.ts
```diff
@@ -30,6 +30,12 @@ const scorecardTranslationDe = createTranslationMessages({
     'dataSourcesDialog.unknownPlugin': 'Unbekannt',
     'dataSourcesDialog.statusTooltip':
       'Wert {{value}} entspricht Schwellenwert {{status}} {{expression}}',
+    'dataSourcesDialog.collectorStatusTooltip':
+      'Dieser Collector liefert nur Eingabedaten. Der {{metric}}-Check-Wert wird aus Collectors berechnet und auf der Scorecard-Karte angezeigt.',
+    'dataSourcesDialog.collectorEmptyValue': '--',
+    'dataSourcesDialog.collectorUnavailableStatus': 'k. A.',
+    'dataSourcesDialog.pluginGithub': 'GitHub',
+    'dataSourcesDialog.pluginJira': 'Jira',
```
dzemanov: @Eswaraiahsapram I see you have also fixed `pluginLabels`, thank you! Would you mind then removing `dataSourcesDialog.pluginJira` from translations if is not used?

dzemanov (approved):
Confirmed that the functionality works as expected:

Single points:

https://github.com/user-attachments/assets/227ff911-27fb-4f38-949a-2d1ba8a5aaa9

Various data with errors:

https://github.com/user-attachments/assets/8272ddfa-5ac8-4734-82d0-b21312d4f6fe

Translation:

https://github.com/user-attachments/assets/86c5b789-c7c6-4d9c-af29-f7f5bdd2ebfa

Translation of collectors description is to be worked within a different ticket.
Adding check description together with collector description to `View datasources` will be also done within a different ticket, as it was confirmed on today's meeting.

It would be also nice to update all mocked collector ids as they were updated to the narrowed down version, to do within a different PR:
- `github:deployments` -> `github:doraDeployments`
- `github:deploymentWorkflowRuns` -> `github:doraDeploymentWorkflowRuns`
- `github:deploymentPullRequests` -> `github:doraDeploymentPullRequests`
- `jira:incidents` -> `jira:doraIncidents`

ciiay (approved):
Hi @Eswaraiahsapram , the changes look good to me, thanks for the update. I also agree with @dzemanov 's point about updating the collector IDs in the mocked data. If you'd prefer to handle that in a separate story, we can merge this PR now. Please open a Jira story to track the work and add the link here for reference 🤝 

/lgtm

Eswaraiahsapram:
> Hi @Eswaraiahsapram , the changes look good to me, thanks for the update. I also agree with @dzemanov 's point about updating the collector IDs in the mocked data. If you'd prefer to handle that in a separate story, we can merge this PR now. Please open a Jira story to track the work and add the link here for reference 🤝
> 
> /lgtm

Thanks @ciiay , I can take care of updating the collector IDs in the Homepage sparkline PR https://github.com/redhat-developer/rhdh-plugins/pull/4596.

Merging now

PR #4596, opened 2026-09-07
feat(scorecard): add homepage sparkline charts for time-series metrics

Eswaraiahsapram:
### Fix - https://redhat.atlassian.net/browse/RHIDP-16834

Follows [#4573](https://github.com/redhat-developer/rhdh-plugins/pull/4573) (merged). Reuses the shared sparkline chart, legend, and time-series utilities from that PR on the Scorecard homepage.

## What
Adds sparkline visualization on Scorecard homepage cards when aggregation metadata `visualization` is `sparkline` (for example DORA). Each card shows a 30-day trend and a threshold legend. Legend and line color come from the time-series threshold rules.

## Screenshots

https://github.com/user-attachments/assets/f3e57394-77df-4e0e-90b5-f6a72fa2c1f5

### Homepage rendering
- **`ScorecardHomepageCard`** loads `GET /aggregations/:aggregationId/metadata` first. When `metadata.visualization` is `sparkline`, it fetches the aggregation time series instead of the snapshot and renders **`AggregatedSparklineCard`**. Other aggregations keep the existing pie, donut, and scalar cards.
- **`AggregatedSparklineCard`** — homepage sparkline card using the shared `SparklineChart` / `toSparklineChartModel` path from #4573, including entity subheader (success count / total) and last-synced info.

### New API hook
- **`useAggregationTimeSeries`** — `useQuery` hook for the default 30-day aggregation time series (`GET /aggregations/:aggregationId/time-series`).

### Homepage widgets
Registers NFS and legacy homepage widgets for:
- Average deployment frequency (`avgDeploymentFrequency`)
- Average change failure rate (`avgChangeFailureRate`)
- Average median lead time for changes (`avgMedianLeadTimeForChanges`)

### Docs
- README documents the metadata → time-series → `AggregatedSparklineCard` path for sparkline aggregations.

## How to test
1. Enable DORA (or any scalar aggregation whose metadata `visualization` is `sparkline`) and add a homepage Scorecard card with `aggregationId` such as `avgDeploymentFrequency`.
2. Open the homepage and confirm the card shows a 30-day sparkline with a threshold legend.
3. Confirm line color matches the latest successful point's threshold rule.
4. Confirm non-sparkline homepage cards (GitHub open PRs, Jira, filecheck, weighted health) are unchanged.
5. With no time-series points, confirm the existing empty state is shown instead of a chart.


workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/AggregatedSparklineCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { t } = useTranslation();
+
+  const lastPoint = series.points[series.points.length - 1];
+  const lastSuccessful = [...series.points]
+    .reverse()
+    .find(point => point.status === 'success' && point.value !== null);
+  const matchingThresholdKey =
+    lastSuccessful?.value !== undefined && lastSuccessful?.value !== null
+      ? getMatchingThresholdKey(lastSuccessful.value, series.thresholds)
+      : undefined;
```
dzemanov: There is no need to resolve matching threshold in frontend, you won't get different answer to the backend one.

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/AggregatedSparklineCard.tsx
```diff
@@ -0,0 +1,145 @@
...
+    evaluation: matchingThresholdKey ?? null,
+    thresholdRules: series.thresholds?.rules,
+  });
+  let chartColor: string;
+  if (series.aggregationChartDisplayColor) {
+    chartColor = resolveStatusColor(theme, series.aggregationChartDisplayColor);
+  } else if (matchingThresholdKey) {
+    chartColor = resolveStatusColor(theme, statusConfig.color);
+  } else {
+    chartColor = theme.palette.grey[500];
```
dzemanov: We should make this constant, like SCORECARD_UNKNOWN_CHART_COLOR.
Or reuse SCORECARD_ERROR_STATE_COLOR.
https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard/src/utils/constants.ts#L20

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
```diff
@@ -314,3 +314,69 @@ export const aggregatedCardWithAvgOpenPrsWidget = HomePageWidgetBlueprint.make({
...
+            <ScorecardHomepageCardWithProvider aggregationId="avgChangeFailureRate" />
+          ),
+      ),
+    },
+  });
+
+/**
+ * NFS widget: AggregatedCardWithAvgMedianLeadTimeForChanges (scalar sparkline).
+ */
+export const aggregatedCardWithAvgMedianLeadTimeForChangesWidget =
```
dzemanov: Would you mind adding here also the last DORA metric: `aggregatedCardWithAvgMedianTimeToRestoreWidget` ?

workspaces/scorecard/packages/app-legacy/src/App.tsx
```diff
@@ -309,6 +309,96 @@ const mountPoints: HomePageCardMountPoint[] = [
...
+      },
+      props: {
+        aggregationId: 'avgChangeFailureRate',
+      },
+    },
+  },
+  {
+    Component: ScorecardHomepageCard as ComponentType,
+    config: {
+      id: 'scorecard-avgMedianLeadTimeForChanges',
```
dzemanov: Would you mind adding here also the last DORA metrics, `avgMedianTimeToRestore`?

workspaces/scorecard/plugins/scorecard/README.md
```diff
@@ -304,6 +313,9 @@ To align with the legacy EntityPage (Scorecard on component pages and default en
    - `AggregatedCardWithTotalOpenBugs` (title: **Scorecard: Total open bugs**)
    - `AggregatedCardWithEntitiesWithOpenPrs` (title: **Scorecard: Entities with open PRs**)
    - `AggregatedCardWithAvgOpenPrs` (title: **Scorecard: Average open PRs**)
+   - `AggregatedCardWithAvgDeploymentFrequency` (title: **Scorecard: Average deployment frequency**)
```
dzemanov: When you will add the last DORA metric for `MedianTimeToRestore`, please update also docs, thank you!

dzemanov (changes requested):
Thank you for the PR.
Looks good, just a couple of quick comments:

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
```diff
@@ -314,3 +314,69 @@ export const aggregatedCardWithAvgOpenPrsWidget = HomePageWidgetBlueprint.make({
...
+      name: 'AggregatedCardWithAvgChangeFailureRate',
+      title: 'Scorecard: Average change failure rate',
+      layout: defaultCardLayout,
+      componentProps: {
+        Renderer: BorderlessHomeWidgetRenderer,
+      },
+      components: lazyScorecardWidget(
+        ScorecardHomepageCardWithProvider => () =>
+          (
+            <ScorecardHomepageCardWithProvider aggregationId="avgChangeFailureRate" />
```
dzemanov: When testing with default configuration, I get:

https://github.com/user-attachments/assets/024bdb19-1532-48b1-927c-624105e81139

All avg dora widgets except for deployment frequency are always loading.

This is because default `app-config` configuration is missing all avg dora configurations except for `avgDeploymentFrequency`: https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/app-config.yaml#L373

I am not familiar much with NFS blueprints in RHDH. If we are providing only these blueprint widgets, this means that in NFS for example for filecheck, users are unable to create aggregation cards except for checking for codeowners and license file because that is the only configuration we provide [here](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx#L150) and [here](https://github.com/redhat-developer/rhdh-plugins/blob/main/workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx#L172) in blueprints?
This also means if they forget to add that key to their `app-config` configuration, that they are always in load state?

It would be good to fix that if metadata API returns not found, widgets are not in always loading state.

Going back to dora, maybe it makes sense to provide at least 1 dora metric with Deprecated metric id (at least for legacy - so we can easier test it is working) - so for example `dora.changeFailureRate`. We can keep `avgDeploymentFrequency` as we already have it in app-config. But maybe we really want to provide all 4 dora metrics also in blueprints with deprecated metric id, as otherwise users will need to have them configured in their `app-config`?

For aggregationId, I am thinking we might want to make the ids general? So instead of `avgChangeFailureRate` we have 'changeFailureRateKpi' similiar to `openIssuesKpi`? Just because you can specify different `type: sum` etc in your scorecard configuration (`avg` is default), what do you think? Using custom `aggregationId` instead of deprecated metric id means you will need to add all those configurations to default app config.

dzemanov (changes requested):

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
Eswaraiahsapram: Thanks @dzemanov. 

1. Default configurations: I'll add the missing KPI config to the default app-config.

2. NFS blueprints: The available homepage cards depend on the blueprints we provide. Users cannot create new widget with custom aggregation IDs unless we/they provide a blueprint widget. Custom KPI IDs also require matching app-config entries

3. I'll add at least one DORA widget using deprecated metricId to test

4. Aggregation ID naming: I agree that avgChangeFailureRate is not a good name because the aggregation type can be changed. I'll use names like changeFailureRateKpi and deploymentFrequencyKpi, following the existing openIssuesKpi pattern.

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
```diff
@@ -314,3 +314,91 @@ export const aggregatedCardWithAvgOpenPrsWidget = HomePageWidgetBlueprint.make({
...
+
+/**
+ * NFS widget: AggregatedCardWithAvgDeploymentFrequency (scalar sparkline).
+ */
+export const aggregatedCardWithAvgDeploymentFrequencyWidget =
+  HomePageWidgetBlueprint.make({
+    name: 'scorecard-avg-deployment-frequency',
+    params: {
+      name: 'AggregatedCardWithAvgDeploymentFrequency',
+      title: 'Scorecard: Average deployment frequency',
```
dzemanov: should we use as default titles the ones that are in design? So `DORA - Deployment Frequency`?

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
```diff
@@ -314,3 +314,113 @@ export const aggregatedCardWithAvgOpenPrsWidget = HomePageWidgetBlueprint.make({
...
+      name: 'AggregatedCardWithAvgDeploymentFrequency',
+      title: 'Scorecard: DORA - Deployment Frequency',
+      layout: defaultCardLayout,
+      componentProps: {
+        Renderer: BorderlessHomeWidgetRenderer,
+      },
+      components: lazyScorecardWidget(
+        ScorecardHomepageCardWithProvider => () =>
+          (
+            <ScorecardHomepageCardWithProvider aggregationId="avgDeploymentFrequency" />
```
dzemanov: Yeah, pardon, would you mind updating also `avgDeploymentFrequency` to `doraDeploymentFrequencyKpi` - to match all other dora metrics.

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
```diff
@@ -314,3 +314,113 @@ export const aggregatedCardWithAvgOpenPrsWidget = HomePageWidgetBlueprint.make({
...
+        Renderer: BorderlessHomeWidgetRenderer,
+      },
+      components: lazyScorecardWidget(
+        ScorecardHomepageCardWithProvider => () =>
+          (
+            <ScorecardHomepageCardWithProvider aggregationId="changeFailureRateKpi" />
+          ),
+      ),
+    },
+  });
```
dzemanov: Since we renamed these to be `changeFailureRateKpi`, we should also update naming of all widgets data accordingly. So:

```suggestion
export const aggregatedCardWithDoraChangeFailureRateWidget =
  HomePageWidgetBlueprint.make({
    name: 'scorecard-dora-change-failure-rate',
    params: {
      name: 'AggregatedCardWithDoraChangeFailureRate',
      title: 'Scorecard: DORA - Change Failure Rate',
      layout: defaultCardLayout,
      componentProps: {
        Renderer: BorderlessHomeWidgetRenderer,
      },
      components: lazyScorecardWidget(
        ScorecardHomepageCardWithProvider => () =>
          (
            <ScorecardHomepageCardWithProvider aggregationId="doraChangeFailureRateKpi" />
          ),
      ),
    },
  });
```

All other dora widgets should also update all their naming.

I added `dora` prefix as I think it would be safer to have it to prevent conflicts. E.g we have `openIssuesKpi` for jira, but there is also `openIssues` in sonarqube...

My apologies for focusing so much on the export naming conventions. As this is a public API, and you said users will not be able to define their own custom kpi IDs, it is important to get it right.

workspaces/scorecard/app-config.yaml
```diff
@@ -310,6 +365,86 @@ sonarqube:
...
+    medianTimeToRestoreKpi:
+      title: DORA - Median Time to Restore
+      type: average
+      description: Tracks the median time to restore service after an incident over the past 30 days. Elite performers restore service in under one hour.
+      metricId: dora.medianTimeToRestore
+      options:
+        thresholds:
+          rules:
+            - key: elite
+              expression: '<24'
```
dzemanov:
```suggestion
              expression: '<1'
```

workspaces/scorecard/app-config.yaml
```diff
@@ -310,6 +365,86 @@ sonarqube:
...
+              color: success.main
+              icon: scorecardSuccessStatusIcon
+            - key: medium
+              expression: '24-168'
```
dzemanov:
```suggestion
              expression: '1-24'
```

workspaces/scorecard/app-config.yaml
```diff
@@ -310,6 +365,86 @@ sonarqube:
...
+              color: warning.main
+              icon: scorecardWarningStatusIcon
+            - key: low
+              expression: '>168'
```
dzemanov:
```suggestion
              expression: '>24'
```

workspaces/scorecard/packages/app-legacy/src/App.tsx
```diff
@@ -309,6 +340,126 @@ const mountPoints: HomePageCardMountPoint[] = [
...
+      props: {
+        aggregationId: 'avgDeploymentFrequency',
+      },
+    },
+  },
+  {
+    Component: ScorecardHomepageCard as ComponentType,
+    config: {
+      id: 'scorecard-changeFailureRateKpi',
+      title: 'Scorecard: Average change failure rate',
```
dzemanov: We can leave it as it is - it is just testing for this workspace. However, it might be confusing if testing for old frontend system has different widgets then NFS. (different titles)

workspaces/scorecard/plugins/scorecard/src/components/AggregatedMetricCards/AggregatedSparklineCard.tsx
```diff
@@ -0,0 +1,137 @@
...
+  const locale = useLanguage();
+  const { t } = useTranslation();
+
+  const lastPoint = series.points[series.points.length - 1];
+  const thresholdRules = series.thresholds?.rules;
+  const chartColorToken = series.aggregationChartDisplayColor;
+  const matchingThresholdKey = chartColorToken
+    ? thresholdRules?.find(
+        rule =>
+          getThresholdRuleColor(thresholdRules, rule.key) === chartColorToken,
```
dzemanov, line 62: Yeah, this reverse engineering of matched threshold by color to get stroke is not great. But backend doesn't provide it so thank you for getting around this. We can fix that in future if needed.

workspaces/scorecard/app-config.yaml
```diff
@@ -310,6 +365,86 @@ sonarqube:
...
+          rules:
+            - key: elite
+              expression: '>=7'
+              color: success.main
+              icon: scorecardSuccessStatusIcon
+            - key: medium
+              expression: '1-7'
+              color: warning.main
+              icon: scorecardWarningStatusIcon
+            - key: error
```
dzemanov:
```suggestion
            - key: low
```

dzemanov (changes requested):
Retested,

default configuration works as expected.

No configuration of aggregation kpi is no longer loading infinitely:

Image

It doesn't look nice, but is in line with how other errors are shown, thanks for the update.

I don't know if this is as a result of fixing loading, but first time when I add scorecard, I get auth error. When I reload, everything works.

https://github.com/user-attachments/assets/37f33bfd-1cc8-46a9-a947-b7f63a28d5d2

No data works fine:

Image

Single points:

Image
Drill down with single points:

Image

2 points:

Image

Different values:

https://github.com/user-attachments/assets/80cf9b30-d8b2-457c-97b1-66977380177d

Translation:

https://github.com/user-attachments/assets/857596a5-4c41-4e37-81f4-2ef2eefdccbb

Just wanted to mention, title and description translations are currently unused for
dora aggregation KPIs in NFS blueprints (and also all other KPIs that don't use as aggregation id metric id), as the `app-config` always overrides
them (this is expected behavior as user configured those). For a future release, we could explore removing the need for the `app-config` configuration by adding blueprints that use `aggregationId=metricId` and using as aggregation thresholds by default metric-defined thresholds instead of DEFAULT_NUMBER_THRESHOLDS.

Would be good to truncate the values for graphs:

Image



@Eswaraiahsapram, adding `current` value is to be done within a different ticket?

Design:

Image

View datasources settings are missing as well for home page, can be added in a different PR as well.

dzemanov (commented):
Permissions work fine:

https://github.com/user-attachments/assets/b95c3b34-33c2-4668-b791-3d2b388c9851

workspaces/scorecard/app-config.yaml
Eswaraiahsapram: Updated

workspaces/scorecard/packages/app-legacy/src/App.tsx
Eswaraiahsapram: Updated

workspaces/scorecard/app-config.yaml
Eswaraiahsapram: updated

workspaces/scorecard/app-config.yaml
Eswaraiahsapram: updated

workspaces/scorecard/app-config.yaml
Eswaraiahsapram: updated

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
Eswaraiahsapram: That makes sense. I’ve updated the naming accordingly.

workspaces/scorecard/plugins/scorecard/src/extensions/homePageCards.tsx
Eswaraiahsapram: updated

Eswaraiahsapram:
Thanks a lot @dzemanov for the feedback.

> It doesn't look nice, but is in line with how other errors are shown, thanks for the update.

Thanks. We have a jira bug for similar issue - https://redhat.atlassian.net/browse/RHDHBUGS-3263

> I don't know if this is as a result of fixing loading, but first time when I add scorecard, I get auth error. When I reload, everything works.

I haven't been able to reproduce this locally. Could you please share the steps? If it still reproduces, I’m happy to file a follow-up bug.

https://github.com/user-attachments/assets/9321ca7a-036d-4e36-a36f-e3625cde74a0

> Would be good to truncate the values for graphs

Fixed - graph values are truncated to 4 decimal places.

> @Eswaraiahsapram, adding current value is to be done within a different ticket?

Thanks for flagging this. I’ve added the `current` value in this PR on the drill-down page. Homepage cards still show only the chart and legend.

> View datasources settings are missing as well for home page, can be added in a different PR as well.

Agreed, we'll pick this up in a follow-up PR.

dzemanov (approved):
Thank you for the updates.

Retested. Truncating works correctly. Current value works nicely.

One point:

https://github.com/user-attachments/assets/1a3cc67f-d828-4456-a7f6-bc874d7c1cc2

Multiple values:

https://github.com/user-attachments/assets/967ee3fe-65ef-4e03-81f7-468636265f24

2 points / all points errors - current is not shown

https://github.com/user-attachments/assets/155927a0-0e58-43b6-a001-e770e0c2c607

Current is translated:

Image

dzemanov:
> I haven't been able to reproduce this locally. Could you please share the steps? If it still reproduces, I’m happy to file a follow-up bug.

@Eswaraiahsapram I haven't been able to reproduce it again, no idea how this was happening. If I encounter it again I will file a ticket with details how to reproduce.

PR #4656, opened 2026-09-09
Upgrade Backstage version to '1.54.6'

asmasarw:
## Summary

- Bump DCM workspace to **Backstage v1.54.6** and update related app, backend, and plugin dependencies.
- Fix **Data Center** sidebar submenu alignment so the icon and label are vertically centered in the active bordered item.
- Fix **Agents** toolbar UI:
  - Correct the **Health status** dropdown label when **All** is selected (outlined `Select` notch/label behavior).
  - Add spacing between the **Register** button and the health filter dropdown.
- Update catalog entity pages for Backstage API changes by removing deprecated `variant="gridItem"` props from entity cards.
- Add a changeset for the DCM plugin version bump and dedupe `workspaces/dcm/yarn.lock`.

## Test plan

- [x] Start the DCM app and verify **Data Center** sidebar item is vertically centered when active.
- [x] On **Agents** tab, confirm **Health status** label renders correctly with **All** selected.
- [x] Confirm spacing between **Register** and the health filter dropdown.
- [x] Run `yarn tsc:full` in `workspaces/dcm`.
- [x] Verify lockfile check passes (`yarn dedupe` in `workspaces/dcm`).


workspaces/dcm/package.json
```diff
@@ -56,7 +56,7 @@
     "knip": "^5.27.4",
     "node-gyp": "^9.0.0",
     "prettier": "^2.3.2",
-    "typescript": "~5.3.0"
+    "typescript": "~5.5.0"
```
mareklibra: Other 1.54.6 workspace bumps this dep to ~5.8.0 alongside @backstage/cli ^0.36.5.

workspaces/dcm/packages/app/src/components/Root/Root.tsx
```diff
@@ -97,21 +103,33 @@ const useSidebarItemStyles = makeStyles(theme => {
...
+      },
     },
     inactiveItem: {
       backgroundColor: 'transparent',
     },
     // Scoped so sidebar item styles win over upstream without !important
     administrationGroup: {
       '& $submenuItem': {
-        '& .MuiSvgIcon-root': { fontSize: '20px' },
+        '& .MuiSvgIcon-root': { fontSize: '14px' },
```
mareklibra, line 119: is that intentional?

workspaces/dcm/packages/app/src/components/Root/Root.tsx
```diff
@@ -97,21 +103,33 @@ const useSidebarItemStyles = makeStyles(theme => {
         dark ? 'rgba(255,255,255,0.2)' : 'rgba(0,0,0,0.12)'
       }`,
       backgroundColor: dark ? 'rgba(255,255,255,0.12)' : 'rgba(0,0,0,0.06)',
+      borderLeft: `1px solid ${
+        dark ? 'rgba(255,255,255,0.2)' : 'rgba(0,0,0,0.12)'
+      }`,
+      '& [data-testid="login-button"]': {
```
mareklibra: Depending on data-testid for a non-test purpose smells...
Can you find another way?

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
```diff
@@ -268,8 +270,24 @@ export function AgentsTabContent() {
...
-        inputProps={{ 'data-testid': 'health-filter' }}
+        label={healthFilterLabel}
+        input={
+          <OutlinedInput
+            notched
+            label={healthFilterLabel}
+            inputProps={{ 'data-testid': 'health-filter' }}
+          />
+        }
+        renderValue={(value: unknown) => {
```
mareklibra, line 281: The outlined Select + renderValue change is the actual Health-status “All” fix, but tests still only fireEvent.change on the hidden input. Could you add an assertion that the empty value shows the translated “All” label (and maybe one selected option)? That is also the uncovered patch line on Codecov.

workspaces/dcm/package.json
```diff
@@ -69,8 +69,8 @@
     "@types/react": "^18",
     "@types/react-dom": "^18",
     "fsevents": "~2.3.2",
-    "@backstage/backend-app-api": "1.4.1",
-    "@backstage/backend-plugin-api": "1.6.2",
+    "@backstage/backend-app-api": "1.7.3",
```
mareklibra: This is better than nothing but can you double-check that explicit resolution is still needed? 
Other plugins dropped this explicit setting, maybe even DCM can

mareklibra (changes requested):

mareklibra:
Pasted a few comments.

When writing this, I am not sure about completeness of the steps needed for the upgrade towards RHDH 2.0.0, let's discuss them offline first.

workspaces/dcm/packages/app/src/components/Root/Root.tsx
asmasarw: Yes, after the upgrade, the Icon was bigger than before

workspaces/dcm/package.json
asmasarw: Correct, Other Workspaces uses 5.8.0 -- adopted here as well.

workspaces/dcm/packages/app/src/components/Root/Root.tsx
asmasarw: Replaced with normal class name (This was my mistake - forget to double check the class names added)

workspaces/dcm/plugins/dcm/src/pages/agents/AgentsTabContent.tsx
asmasarw: Test Class added

workspaces/dcm/package.json
asmasarw: Dropped from Resolution.

PR #4582, opened 2026-09-04
RHIDP-16658: Consume backend-provided color for scalar aggregation KPI cards

imykhno:
## Hey, I just made a Pull Request!

In this PR, we added the `result.aggregationChartDisplayColor` attribute to the scalar aggregation response. The color will be calculated on the backend, which now returns only the correct value.

Dropping `result.thresholds` was not implemented in this PR, as we decided that the UI may still need threshold information for certain charts

#### This PR is for:
* [RHIDP-16658](https://redhat.atlassian.net/browse/RHIDP-16658)

#### :heavy_check_mark: Checklist

- [x] A changeset describing the change and affected packages. ([more info](https://github.com/redhat-developer/rhdh-plugins/blob/main/CONTRIBUTING.md#creating-changesets))
- [ ] Added or Updated documentation
- [x] Tests for new functionality and regression tests for bug fixes
- [ ] Screenshots attached (for UI changes)

PR #4657, opened 2026-09-09
RHIDP-16510: Consume backend-provided color for scalar aggregation KPI cards

ciiay:
## Summary

- `ScalarStatCard` now uses `result.aggregationChartDisplayColor` from the backend instead of re-evaluating threshold expressions on the frontend
- Grey fallback is preserved when the backend returns `null` (empty or all-failed aggregations with `total === 0`)
- Min/max scalar drill-down pages now default-sort the entities table by Value:
  - **Min** → ascending (lowest values first)
  - **Max** → descending (highest values first)
- Follow-up frontend work for RHIDP-16658 / #4582 and the agreement from #4395 review

### UI / behavior changes

- **Scalar KPI cards** (sum, average, min, max): tile background color now comes from backend `aggregationChartDisplayColor` instead of frontend threshold re-evaluation
- **Empty scalar KPIs** (`total === 0`): tile renders neutral grey when backend returns `null` for display color
- **Min/max drill-down**: entities table opens with Value sorted so the entity driving the aggregate appears first; users can still change sort via column headers

#### This PR is for:

* [RHIDP-16510](https://redhat.atlassian.net/browse/RHIDP-16510)

## Test plan

- [x] `yarn test --testPathPatterns=ScorecardHomepageCard.test` passes
- [x] `yarn test --testPathPatterns=EntitiesTable.test` passes
- [x] `yarn test --testPathPatterns=getDefaultEntitiesTableSort.test` passes
- [x] `yarn tsc` passes in scorecard workspace
- [ ] Manual smoke test with `yarn start` — scalar KPI cards load with expected colors; min/max drill-down entities table opens with correct default Value sort


imykhno (approved):
Tested locally, looks good to me! Thank you for the PR and for fixing the `Min` and `Max` aggregation types on the homepage cards.

/lgtm

PR #4667, opened 2026-09-09
feat(ai-integrations): add 'mcp-registry-server-mapping' and 'mcp-registry-provider' openspec files

michael-valdron:

Contributes openspecs for MCP Registry Provider plugin implementation.

https://redhat.atlassian.net/browse/RHIDP-15658


michael-valdron:
/fs-fix Apply the following fix from [previous review comment](https://github.com/redhat-developer/rhdh-plugins/pull/4527#issuecomment-5502359609): `Change schema: rhdh-spec-driven to schema: spec-driven in both .openspec.yaml files to match the established convention`

michael-valdron:
/fs-fix Apply the following fix from [previous review comment](https://github.com/redhat-developer/rhdh-plugins/pull/4527#issuecomment-5502359609): `Change schema: rhdh-spec-driven to schema: spec-driven in both .openspec.yaml files to match the established convention`

michael-valdron:
/fs-fix
- Add `status: draft` to the openspecs: `Six of seven existing siblings include status: draft. Both new files omit it.`
- Apply: `Add H1 headings to match the established pattern (e.g., # Proposal: MCP Registry Provider).`

michael-valdron:
/fs-fix Restructure each decision body under `workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/design.md` and `workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md` to use the established label pattern.

michael-valdron:
/fs-fix Clarify that the hash suffix is always applied when sanitization changes any character or combined length exceeds 63 chars, preemptively preventing collisions since different canonical inputs produce different hashes.

michael-valdron:
/fs-fix Update spec, design D10, and proposal to state the annotation value MUST use url: format.

johnmcollier (commented):
Overall, this aligns largely with what I would expect from an entity provider ingesting from an MCP registry. One question that stands out to me:

How do we want to handle collisions between multiple MCP registries? What if multiple MCP registries emit the same catalog entity identity? We might want to consider some kind of namespacing per-mcp registry?

michael-valdron:
> Overall, this aligns largely with what I would expect from an entity provider ingesting from an MCP registry. One question that stands out to me:
> 
> How do we want to handle collisions between multiple MCP registries? What if multiple MCP registries emit the same catalog entity identity? We might want to consider some kind of namespacing per-mcp registry?

@johnmcollier Good point, supporting multiple MCP Registries would make sense. Though not listed as out of scope, the current design from the feature only considers one MCP Registry at a time:
> `baseUrl`: #The URL of the MCP Registry.

michael-valdron:
> > Overall, this aligns largely with what I would expect from an entity provider ingesting from an MCP registry. One question that stands out to me:
> > How do we want to handle collisions between multiple MCP registries? What if multiple MCP registries emit the same catalog entity identity? We might want to consider some kind of namespacing per-mcp registry?
> 
> @johnmcollier Good point, supporting multiple MCP Registries would make sense. Though not listed as out of scope, the current design from the feature only considers one MCP Registry at a time:
> 
> > `baseUrl`: #The URL of the MCP Registry.

That being said I could revise a bit at least to have the consideration that multiple registries may be supported down the road to avoid future catalog breakages.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
```diff
@@ -0,0 +1,155 @@
...
+#### Scenario: Two versions of the same server produce distinct entities
+
+- **WHEN** two `server.json` documents share the canonical name `io.github.user/weather` but declare `version` `1.0.0` and `2.0.0`
+- **THEN** the two produced entities have distinct `metadata.name` values (each incorporating its version), each carries the same `modelcontextprotocol.io/name: io.github.user/weather`, and each carries its own `modelcontextprotocol.io/version` (`1.0.0` and `2.0.0` respectively)
+
+#### Scenario: Reverse-DNS name and version are sanitized into metadata.name
+
+- **WHEN** `server.json` `name` is `io.github.user/weather` and `version` is `1.0.2`
+- **THEN** `metadata.name` is the sanitized `<name>__<version>` form (e.g. `io.github.user-weather__1.0.2`), the original `io.github.user/weather` is preserved verbatim in `modelcontextprotocol.io/name`, and `1.0.2` is recorded in `modelcontextprotocol.io/version`
+
```
gabemontero, line 84: The spec requires a hash suffix only when a collision is observed, but this transform is pure over one document and cannot know whether another document sanitizes to the same name. Could we make the rule deterministic per input (for example, hash whenever sanitization changes the identity or the value is truncated), or explicitly move collision resolution into the provider?

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/design.md
```diff
@@ -0,0 +1,144 @@
...
+- **`metadata.name` collisions across registries** (same name+version from two registries) → Out of scope here (no dedup); documented so the future ingestion change can namespace or dedup. Within a single source the `<name>__<version>` + hash-suffix rule guarantees uniqueness.
+- **Draft schema drift** → D7 fail-open projection; the mapping table is versioned against the draft and revisited when the schema changes.
+- **Lossy flattening of deep `packages[]` config** → Accepted; runtime package details are preserved as scalar-leaf annotations for discoverability, not interpreted. Round-trip fidelity is guaranteed only for scalar leaves.
+- **Secret leakage into searchable annotations** (remote `headers`/`variables`, `environmentVariables` carrying `default`/`value`) → D9 prunes the `default`/`value` leaves of any `isSecret: true` input from projection. This is a deliberate carve-out from scalar round-trip fidelity — secret leaves are intentionally unrecoverable from the entity. Non-secret metadata on the same input still projects, so discoverability is preserved.
```
gabemontero: D9 still projects the `choices` leaf for `isSecret: true` inputs. If choices enumerate valid secret values, those values become searchable catalog annotations. Please either redact secret-associated choices too or document and test why they are safe.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
```diff
@@ -0,0 +1,155 @@
...
+The [`server.json` `repository` object](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/draft/server.schema.json) supplies a repository-root `url` (browse and git clone), a hosting-service `source` identifier, and an optional `subfolder` that is a clean relative path from the repository root. It does not supply a git branch or ref. The mapping therefore MUST NOT invent a branch name such as `main`. The mapping SHALL compute the combined repository URL as follows:
+
+1. **Normalize the base URL.** Let `base` be `repository.url` with any trailing `/` removed and with a trailing `.git` suffix removed when present.
```
gabemontero: The algorithm strips trailing `/` and `.git`, then emits only the combined URL, but the annotation-projection contract promises recovery of every non-secret scalar. That means the original `repository.url` is not recoverable for those inputs. Please preserve it in a dedicated annotation or narrow the round-trip guarantee with an explicit exception and scenario.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
```diff
@@ -0,0 +1,155 @@
...
+- **THEN** the mapping produces a deterministic, unique `metadata.name` by truncating and appending a stable hash suffix derived from the canonical name and version
+
+### Requirement: Map descriptive metadata to native Backstage fields
+
+The mapping SHALL map `server.json` descriptive attributes to native Backstage `metadata` fields: `title` → `metadata.title`; `description` → `metadata.description`; `websiteUrl` → a `metadata.links` entry whose `url` is the `websiteUrl` and whose `title` is `Website`; and `repository.url` combined with `repository.subfolder` when present (per the repository URL combination algorithm below) → a `metadata.links` entry whose `url` is that combined URL and whose `title` is `Source Code`. The mapping SHALL ALSO emit the same combined URL as a `backstage.io/source-location` annotation whose value MUST use the `url:` format (`url:<combined-url>`), so the repository is captured both as the canonical Backstage source-location annotation (for source-aware tooling) and as a human-visible source link. The mapping SHALL set `metadata.tags` to include the upstream mcp-server convention tags (`mcp`, `ai`).
```
gabemontero: Because `websiteUrl` and repository URLs are registry-controlled, could the spec define safe handling for disallowed schemes such as `javascript:`/`data:` and clarify the policy for internal-network URLs before copying them into links and `backstage.io/source-location`? Please add adversarial scenarios.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Alternative considered:** Trust a single page (prototype behavior) — rejected; silently truncates ingestion for any registry larger than one page.
+
+**Rationale:** Unlike the reference prototype (first page only), full cursor pagination ensures complete ingestion for registries of any size.
+
+### D5: Delegate wholly to `mcp-registry-server-mapping`; supply `defaultOwner` as the caller override
+
+**Choice:** For each server the provider calls the mapping transform, passing the instance's `defaultOwner` as the caller-override owner default (and any future caller defaults like `lifecycle`). The provider adds only provider-level concerns on top of the transform's output: the managed-by-location annotation / `locationKey` for catalog attribution and pruning. It never re-derives names, annotations, or `spec.remotes`.
```
gabemontero: This says “managed-by-location annotation / `locationKey`,” which conflates two different Backstage mechanisms. Could the design state precisely whether the implementation adds `backstage.io/managed-by-location` or relies on the provider mutation's `locationKey`, then align the task and scenario with that choice?

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Rationale:** Following the proposal's stated `v1` establishes the forward-looking default; operators override `apiVersion` to match their registry's actual version.
+
+## Risks / Trade-offs
+
+- **apiVersion default (`v1`) does not match the live registry (`/v0`, `/v0.1`)** → Operators must set `apiVersion` to their registry's actual version; the default is documented and overridable, and startup/first-sync errors name the endpoint that was requested. Revisit the default if the registry standardizes on a version.
+- **Non-terminating or repeating cursor from a buggy registry** → D4 loop safeguard (max pages/total + repeated-cursor detection) trips and fails the run (D6) rather than looping forever or committing a partial catalog.
+- **Partial-page fetch failure mid-pagination** → D6 fails the whole run with no mutation, preserving prior catalog state; no partial full mutation is ever committed.
+- **`metadata.name` collisions across registries** (same name+version from two registries) → Out of scope (carried from the mapping's non-goals); per-instance `locationKey` keeps each provider's entities attributable, but true dedup/merge is deferred. Documented for a future change.
```
gabemontero: This is adding onto the prior discussion @michael-valdron and @johnmcollier had, where the last comment to date was at https://github.com/redhat-developer/rhdh-plugins/pull/4667#issuecomment-5640550664

Deferring same `(kind, namespace, name)` output from multiple registries as “dedup/merge” does not resolve catalog identity collisions, nor do I see at least some detail on how the existence of the situation is easily identifiable ; 

Some clear warning level logging seems a meets min given

Otherwise, could we define a super simple, implied, one provider always wins or takes precedence (maybe the order in the config), or maybe names are source-qualified?

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md
```diff
@@ -0,0 +1,129 @@
...
+- **THEN** the current sync's full mutation omits that server's entity, and the catalog removes it
+
+#### Scenario: Updated server reflects latest state
+
+- **WHEN** a server's `server.json` changes between syncs (e.g. a new description)
+- **THEN** the produced entity in the current full mutation reflects the latest `server.json`
+
+### Requirement: Resilient, agent-native error handling
+
+A single server entry that cannot be mapped (e.g. it omits a `server.json`-required field and the mapping rejects it) SHALL be logged with an actionable message identifying the entry and SHALL be skipped, without aborting the sync or discarding the other entries. A registry transport or protocol error (unreachable host, non-success HTTP status, or unparseable response body) SHALL fail the current sync run: the provider SHALL NOT commit a partial full mutation, SHALL log the error, and SHALL retry on the next scheduled tick, leaving the prior catalog state intact.
```
gabemontero: If an entity was present in a prior full mutation but its entry becomes unmappable, this rule skips it and the next full mutation omits it. Is pruning that prior entity intentional? Please specify and test whether it is pruned or retained.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
```diff
@@ -0,0 +1,132 @@
...
+
+**Alternative considered:** Require `schedule` and fail if absent — rejected; a sensible default keeps first-run setup simple, consistent with the mapping's "never fail for a supplyable default" stance.
+
+**Rationale:** A missing schedule should not block ingestion. A sensible default keeps first-run setup simple.
+
+### D4: Full cursor pagination is mandatory
+
+**Choice:** The provider loops: request `<baseUrl>/<apiVersion>/servers`, accumulate `servers[]`, read `metadata.nextCursor`, and re-request with `?cursor=<value>` until the cursor is absent/empty. Cursors are opaque and passed verbatim. A **loop safeguard** (max-pages / max-total bound, plus detecting a repeated cursor) prevents a misbehaving registry from spinning forever; hitting the bound fails the run (D6) rather than committing a partial catalog.
```
gabemontero: The safeguard is described as “max-pages / max-total” but no concrete bound or configuration source is defined, while the open question defers those values. Please specify deterministic defaults so implementations and tests cannot choose incompatible termination behavior.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/tasks.md
```diff
@@ -0,0 +1,51 @@
...
+
+## 1. Plugin Scaffolding & Packaging
+
+- [ ] 1.1 Create the backend `catalog-backend-module` plugin package (Backstage catalog-backend-module naming convention) with `package.json`, `tsconfig`, and lint config matching the workspace's plugin conventions
+- [ ] 1.2 Add the `createBackendModule` skeleton that registers against `catalogProcessingExtensionPoint`, depending on `coreServices` (`rootConfig`, `logger`, `scheduler`)
+- [ ] 1.3 Document installation in the plugin `README.md` (add via `backend.add(...)`, minimal app-config example)
+
+## 2. Configuration
+
+- [ ] 2.1 Author `config.d.ts` declaring `catalog.providers.mcpRegistry.<id>` with `baseUrl` (required), `apiVersion?` (default `v1`), `schedule?` (`SchedulerServiceTaskScheduleDefinitionConfig`), and `defaultOwner?`
```
gabemontero:
```suggestion
- [ ] 2.1 Author `config.d.ts` declaring `catalog.providers.mcpRegistry.<id>` with `baseUrl` (required), `apiVersion?` (default `v1`), `schedule?` (`SchedulerServiceTaskScheduleDefinitionConfig`), and `defaultOwner?`; require `@visibility backend` annotations for backend-only fields such as `baseUrl`
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/proposal.md
```diff
@@ -0,0 +1,77 @@
...
+- `mcp-registry-annotation-projection`: The fallback rule that projects every `server.json` attribute without a native home into `modelcontextprotocol.io/<dot.path>` annotations — dot-path encoding of nested objects and array indices, catalog-valid key sanitization/truncation, no-overwrite of natively-mapped fields, and scalar round-trip fidelity.
+
+### Modified Capabilities
+
+_(none — no long-lived specs exist under `openspec/specs/` yet; this change introduces new capabilities only and consumes the upstream Backstage `mcp-server` `API` entity contract defined in [`backstage/backstage`](https://github.com/backstage/backstage).)_
+
+## Non-goals
+
+- **Ingestion / runtime.** No registry HTTP client, polling schedule, catalog entity provider/processor, or entity lifecycle management. This change defines the pure `server.json` → entity transform only; the component that fetches entries and applies the mapping is a separate future change.
```
gabemontero: The branch now contains the sibling `mcp-registry-provider` change that consumes this contract, so “a separate future change” is stale. Could this explicitly reference the sibling change instead?

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
+# Audit: MCP Registry Provider
+
+## Audit Report: mcp-registry-provider
+
+**Last audited:** 2026-08-27T18:25:00Z
```
gabemontero: This committed audit report is stale relative to the final branch: its `Last audited` timestamp predates the later OpenSpec revisions in this PR, and it reports no findings. Please regenerate this audit after the final artifact changes and either record the remaining findings or document their resolution so the report is not mistaken for evidence that the final artifacts were audited.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
+# Audit: MCP Registry Server Mapping
+
+## Audit Report: mcp-registry-server-mapping
+
+**Last audited:** 2026-08-25T00:03:43Z
```
gabemontero: This audit report is also stale relative to the final branch: the timestamp predates the later repository-URL revisions, while the report says the artifacts are clean. Please rerun the audit after the final changes and record any findings that remain, or document why each later review concern was resolved.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+| H — Security lint          | 0        | 0       | 0          |
+| **Total**                  | **0**    | **0**   | **0**      |
+
+### CRITICAL
+
+- None
+
+### WARNING
+
+- None
```
gabemontero:
```suggestion
- **[C]** [`openspec/changes/mcp-registry-provider/proposal.md:13`](proposal.md#L13) — The proposal chooses a `v1` default while documenting that the cited reference registry serves `v0`/`v0.1`, but does not define how the default is validated or made operationally safe. Resolve the intended compatibility policy and add a test for the default endpoint (or explicitly label the default as a forward-looking contract).
- **[D]** [`openspec/changes/mcp-registry-provider/tasks.md:17`](tasks.md#L17) — The config task does not require `@visibility backend` on `baseUrl` and other backend-only provider settings. Require those annotations in `config.d.ts`, per `AGENTS.md`, so registry endpoints are not exposed to the frontend config surface.
- **[E]** [`openspec/changes/mcp-registry-provider/design.md:118`](design.md#L118) — The design defers same `(kind, namespace, name)` output from multiple registries as “dedup/merge,” but `locationKey` does not remove catalog identity collisions. Define whether one provider wins, the mapping name is source-qualified, or configuration rejects this case; add a scenario covering two registries publishing the same server version.
- **[G]** [`openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md:119`](specs/mcp-registry-provider/spec.md#L119) — A mapping failure is skipped and the next full mutation contains only successful entries, but behavior when an already-ingested entity becomes unmappable is unspecified. State and test whether that prior entity is pruned or retained on the next sync.
- **[G]** [`openspec/changes/mcp-registry-provider/design.md:91`](design.md#L91) — “managed-by-location annotation / `locationKey`” conflates an entity annotation with the provider mutation’s `locationKey`. Define the exact Backstage mechanism and whether the implementation adds `backstage.io/managed-by-location` or relies solely on provider `locationKey`; update the task and scenario consistently.
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+
+### SUGGESTION
+
+- None
```
gabemontero:
```suggestion
- **[G]** [`openspec/changes/mcp-registry-provider/design.md:83`](design.md#L83) — “max-pages / max-total bound” has no concrete value or configuration source, while the open question defers those bounds. Specify deterministic defaults (and their precedence if configurable) so implementations and tests cannot choose incompatible safeguards.
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+
+| Category                   | CRITICAL | WARNING | SUGGESTION |
+| -------------------------- | -------- | ------- | ---------- |
+| A — Entity propagation     | 0        | 0       | 0          |
+| B — Enum / vocabulary      | 0        | 0       | 0          |
+| C — Semantic contradiction | 0        | 0       | 0          |
+| D — Codebase & convention  | 0        | 0       | 0          |
+| E — Namespace & ownership  | 0        | 0       | 0          |
+| F — Template / copy-paste  | 0        | 0       | 0          |
+| G — Extended coherence     | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| G — Extended coherence     | 0        | 2       | 1          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+
+**Last audited:** 2026-08-27T18:25:00Z
+
+### Summary
+
+| Category                   | CRITICAL | WARNING | SUGGESTION |
+| -------------------------- | -------- | ------- | ---------- |
+| A — Entity propagation     | 0        | 0       | 0          |
+| B — Enum / vocabulary      | 0        | 0       | 0          |
+| C — Semantic contradiction | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| C — Semantic contradiction | 0        | 01      | 0          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+| D — Codebase & convention  | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| D — Codebase & convention  | 0        | 1      | 0          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+| E — Namespace & ownership  | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| E — Namespace & ownership  | 0        | 1       | 0          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/audit.md
```diff
@@ -0,0 +1,31 @@
...
+| F — Template / copy-paste  | 0        | 0       | 0          |
+| G — Extended coherence     | 0        | 0       | 0          |
+| H — Security lint          | 0        | 0       | 0          |
+| **Total**                  | **0**    | **0**   | **0**      |
```
gabemontero:
```suggestion
| **Total**                  | **0**    | **6**   | **1**      |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+
+_Re-audit following the `repository.url`/`websiteUrl` link-mapping revision (repository.url now dual-emits `backstage.io/source-location` + a `metadata.links` "Source Code" entry; `websiteUrl` link titled "Website"). Fix-and-reaudit loop ran 2 passes; all findings from both passes were fixed and confirmed. No CRITICAL findings at any point in this run._
+
+### Summary
+
+| Category                               | CRITICAL | WARNING | SUGGESTION |
+| -------------------------------------- | -------- | ------- | ---------- |
+| A (Entity propagation)                 | 0        | 0       | 0          |
+| B (Enum / vocabulary)                  | 0        | 0       | 0          |
+| C (Semantic contradiction)             | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| C (Semantic contradiction)             | 0        | 2       | 0          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+| D (Codebase & convention grounding)    | 0        | 0       | 0          |
+| E (Namespace & cross-change ownership) | 0        | 0       | 0          |
+| F (Template / copy-paste residue)      | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| F (Template / copy-paste residue)      | 0        | 0       | 1         |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+| G (Extended coherence)                 | 0        | 0       | 0          |
+| H (Security lint)                      | 0        | 0       | 0          |
```
gabemontero:
```suggestion
| H (Security lint)                      | 0        | 2      | 0          |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+| **Total**                              | **0**    | **0**   | **0**      |
```
gabemontero:
```suggestion
| **Total**                              | **0**    | **4**   | **1**      |
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+
+### CRITICAL
+
+- None
+
+### WARNING
+
+- None
```
gabemontero:
```suggestion
- **[C]** [`openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md:59`](specs/mcp-registry-server-mapping/spec.md#L59) — The spec requires distinct sanitized-name collisions to receive a hash suffix, but the transform is pure over one document and cannot know whether another document collides. Make the rule deterministic per input (for example, hash whenever sanitization changes the identity or the value is truncated), or explicitly move collision resolution to the provider and specify that contract.
- **[C]** [`openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md:71`](specs/mcp-registry-server-mapping/spec.md#L71) — The repository algorithm strips trailing `/` and `.git`, then emits only the combined URL while the annotation-projection spec promises recovery of every non-secret scalar. The original `repository.url` is therefore not recoverable for such inputs. Preserve the original URL in a dedicated annotation or narrow the round-trip guarantee with an explicit exception and scenario.
- **[H]** [`openspec/changes/mcp-registry-server-mapping/design.md:133`](design.md#L133) — D9 continues projecting `choices` for `isSecret: true` inputs. If choices contain allowed secret values, those values become searchable catalog annotations. Redact secret-associated choices too, or document and test why they are safe.
- **[H]** [`openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md:67`](specs/mcp-registry-server-mapping/spec.md#L67) — Registry-controlled `websiteUrl` and repository URLs are copied into links and `backstage.io/source-location` without a scheme/host safety rule. Specify safe handling for disallowed schemes such as `javascript:`/`data:` and decide how internal-network URLs should be treated; add adversarial scenarios.
```

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/audit.md
```diff
@@ -0,0 +1,47 @@
...
+
+### SUGGESTION
+
+- None
```
gabemontero:
```suggestion
- **[F]** [`openspec/changes/mcp-registry-server-mapping/proposal.md:29`](specs/mcp-registry-server-mapping/proposal.md#L29) — The proposal still calls ingestion “a separate future change,” although this branch contains the sibling `mcp-registry-provider` change that consumes this contract. Replace that wording with an explicit sibling reference to remove stale final-artifact context.
```

gabemontero (commented):
First @michael-valdron @johnmcollier - glad to see the new openspec audit skill has already been leveraged :-) 

For my review, codex and I did try to factor in the reviews from fullsend and @johnmcollier  ... generally speaking, we either
- added to existing review comments in the case of fullsend
- or tried to reference prior review discussion form @johnmcollier and @michael-valdron 

that said, fwiw, codex and the openspec-audit skill think they found items not previously raised 

let's see @michael-valdron what you and your AI say about those findinds

michael-valdron:
@gabemontero I pulled your `audit.md` changes and re-running it on my end (with recent revision from @johnmcollier's feedback).

After this is done, going to also review your other feedback and apply the fixes where applicable then iterate again over those changes.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/design.md
michael-valdron: @gabemontero As I'm understanding the text `isSecret: true` condition will redact the `value` / `default` fields with the secret values. If `headers`/`variables`, `environmentVariables` under the `server.json` are not secret, `isSecret: false`, then should they still be hidden? Maybe I'm not fully understanding the feedback here.

Assuming the `server.json` is coming from a private MCP registry as ALL values in those fields will be seen be anyone with access to `/v<version>/servers` endpoint.

gabemontero: does adding the clarifications you just noted @michael-valdron into the line here or around it in design.md satisfy the `document ... why they are safe` option from mine and codex's suggestions ?

If so, let's do that and resolve this comment

michael-valdron: @gabemontero I did a `grill-me` on the comments and here's the clarification and recommendation:

> **Mostly, with two important mismatches.**
>
> What D9 actually does (design and the projection spec agree):
> 
> - `isSecret: true` → **omit** that input’s `default` and `value` from catalog annotations. It does not mask them as `***`. They are gone from the entity.
> - `isSecret: false`, or `isSecret` omitted → `default`/`value` **are projected**. They are not hidden.
> - The containers `headers`, `variables`, and `environmentVariables` are not hidden as a group. Only those two leaves on inputs flagged secret are dropped.
> - Siblings still project even when `isSecret: true`: `name`, `description`, `format`, `isRequired`, `isSecret` itself, and `choices`.
> 
> So if you expected “non-secret headers/env vars should also be hidden,” **that is not what `design.md` says.** The “Non-secret input value is retained” scenario is explicit.
> 
> Your private-registry point is also real, and D9 does **not** follow it. Anyone who can `GET /v1/servers` already has the full `server.json`. D9 is not about hiding data from registry clients. It is about not copying publisher-flagged credentials into **Backstage catalog annotations**, which are plaintext, searchable, and usually visible to a wider set of people than the registry API (catalog UI, entity YAML, search, agents). D11’s “trust the configured registry” is about host/scheme policy, not about treating the catalog as the same trust boundary as `/servers`.
> 
> **Is the catalog a different trust boundary from the private registry API, so D9 stays (omit `isSecret: true` `default`/`value` only), or is the catalog the same audience as `/servers`, so D9 is redundant and those leaves should project too?**
> 
> **Recommendation:** Keep D9 as written. `isSecret` is the publisher’s signal; `isSecret: false` values should stay in the catalog (that is useful config, not a secret). Do not hide all headers/env defaults. The private registry already stores the secret; the mapping’s job is not to re-publish it into a second, more searchable store.

Based on the output, looks like most is good to keep as-is however..

> D9 still projects the `choices` leaf

I clarified with [json-schema.app](https://json-schema.app/view/%23?url=https%3A%2F%2Fstatic.modelcontextprotocol.io%2Fschemas%2F2025-09-29%2Fserver.schema.json) what `choices` is which I noticed is the central focus of your and `audit.md` feedback and this should be also redacted from the mapping if `isSecret: true` as it contains possible values for `value` field (already redacted).

image

gabemontero: sure let's go with this and move on @michael-valdron  ... sufficient due diligence for me ... we always have the code review to sort through any gaps missed during ours and AI's attempts to cover this, where if need be we patch the openspec's in the fullsend coding PR

gabemontero: marking resolved

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
michael-valdron: @gabemontero Does the changes in https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/c818f7c79828550d0d579bbfc1a48942032eeea4 work? It defines a `prefix` / `baseName` for now with single MCP Registry connected but considers doing a `<registryPrefix>__<name>__<version>` down the road.

Otherwise, your suggestion also works (only keep entries that come first / parse MCP servers from first to last MCP registry when multiple registries feature is implemented).

cc @johnmcollier

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md
michael-valdron: @gabemontero I'm not sure what is being asked here, do you mean what is the action on prior added entities from a `server.json` if the provider reaches one that is not a valid schema?

This is what I got from `grill-me` with cursor:

> **Fact from the artifacts:** skip plus a successful full mutation already **prunes**.
> 
> The skip rule only keeps the **run** going. The mutation is still `{ type: 'full', entities: successfullyMapped }`. Backstage replaces this provider’s bucket with that list. A server that mapped last tick and fails mapping this tick is not in the list, so the catalog **removes** it.
> 
> That is **not** written as a scenario. “Removed server is pruned” only covers **absent from the registry**. “One malformed server does not abort” does not say whether a **prior** entity for that same server stays or goes. D6’s “don’t prune healthy entities” is about **registry fetch failure** (no mutation). It does not protect a last-good entity when mapping fails.
> 
> **Decision:** when a previously ingested server is still in the registry but this tick’s `server.json` is unmappable, should the next full mutation **prune** that catalog entity, or **retain** the last-good one?
> 
> **Recommendation: prune, and specify it.** Full mutation means the catalog is “what we can map from this tick,” not “last-good union skip.” Retention needs extra last-good state, including when mapping cannot even produce `metadata.name`. Stale catalog copy of a now-invalid document is worse than a logged skip plus deletion. Add a WHEN/THEN (prior success, this tick mapping reject → entity omitted → catalog removes it) and a test next to 4.5/5.3.
> 
> If you instead want **retain**, that is a different contract: skip means “do not change this ref,” not “omit from the full list.”
> 
> Which should it be: **prune** or **retain**?

gabemontero: the review comment centers around the scenario where 

- the server entry's data was originally of a valid format, and hence imported into the catalog
- subsequent updates from the external source for that entity now have a payload which is now an invalid format

what do we do in that case ... leave the now stale original version, or prune it since its source of truth data is now invalid

the spec handles the case where the format was invalid form the start 

does that clarification help you @michael-valdron ?

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
michael-valdron: @gabemontero I think https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/cb5e52f4cdd704d4aee89573a2950ed5a68588a9 resolves the not so clear separation in the `design.md`.

Let me know if the recent changes still doesn't address this and what I'm missing, otherwise this thread should be resolved.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/proposal.md
michael-valdron: Resolved with https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/3749c28c9fb8a7d7d261cc7a52d07c9d62acc7be.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
michael-valdron: Addressed in https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/9ffcc6983acde09bc30e1c830ae30f871c1302dd.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
michael-valdron: Addressed in https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/0a3f885ac44562da8b3f3ecfe00f3f3f23c1737b.

workspaces/ai-integrations/openspec/changes/mcp-registry-server-mapping/specs/mcp-registry-server-mapping/spec.md
michael-valdron: Addressed in https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/e049ec254d95cdb5004a7cace6552003b724c222.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
michael-valdron: Added these safeguards in the openspecs and removed from deferred open questions: https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/86aee9d708225dfa78f974d4aa361487e0c55667

johnmcollier (approved):

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md
michael-valdron: @gabemontero This clears it up now, thanks!

In this scenario we'll want to keep the original entries from the valid format and fail the invalid format with no changes (**retain**). I'll apply changes to address this.

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
gabemontero: I'm fine with it @michael-valdron 

marking as resolved

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md
gabemontero: sounds good @michael-valdron 

I would add one small bit ... include an annotation/tag/label that indicates later revisions were rejected because of errors

we don't want users to have to mine logs to understand why later versions did not show up in the catalog

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/design.md
gabemontero: works for me .... marking resolved

workspaces/ai-integrations/openspec/changes/mcp-registry-provider/specs/mcp-registry-provider/spec.md
michael-valdron: @gabemontero Good point, I'll add a revision to include this along with https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/30e1ebf6161efd9172dba66810fb9e71132c4431 that addresses the issue itself.

michael-valdron: @gabemontero I've added https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/5570b25b64873a6bd11ff990692d748552532756 to add the `redhat.com/rhdh-mcp-registry-sync-status` to store the status if the entities are synced `ok` or last-good due to a failure `degraded`.

Also added https://github.com/redhat-developer/rhdh-plugins/pull/4667/commits/37a56dd0c1ab27d42b5814f54a5f57916715b435 to tighted the specification of the last-good strategy.

michael-valdron:
@gabemontero @johnmcollier 

Looks like I addressed @gabemontero's remaining feedback and got a clear status from `audit` and Fullsend.

gabemontero (approved):

PR #4688, opened 2026-09-10
feat(adoption-insights): add if predicate gating

Eswaraiahsapram:

**Resolves:**
https://redhat.atlassian.net/browse/RHIDP-15543

**Solution description:**
- Register the NFS Adoption Insights page with an **if predicate** (`permissions: { $contains: 'adoption-insights.events.read#read' }`) so the page and sidebar entry are only shown when the user is allowed `adoption-insights.events.read`.
- Keep the existing `usePermission` check on `AdoptionInsightsPage` for the legacy (OFS) flow, which still shows the permission-required empty state on direct navigation. The OFS sidebar item is not gated.
- Add Playwright permission e2e tests (allow/deny for sidebar and page access) with multi-locale support, plus a unit test that the NFS page is registered with the predicate.
- Update the frontend and backend READMEs to document NFS permission-based visibility.

## Screenshot:
image

### DIFF
--- a/workspaces/adoption-insights/.changeset/adoption-insights-events-read-if-predicate.md
+++ b/workspaces/adoption-insights/.changeset/adoption-insights-events-read-if-predicate.md
@@ -1,0 +1,5 @@
+---
+'@red-hat-developer-hub/backstage-plugin-adoption-insights': patch
+---
+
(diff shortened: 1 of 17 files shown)


jrichter1 (approved):

debsmita1 (commented):
Reviewed and tested the PR, works fine locally

PR #4701, opened 2026-09-10
feat(dcm): forward OIDC tokens and support auth-disabled deployments (FLPATH-4480)

chadcrum:
FLPATH-4480\n\nhttps://redhat.atlassian.net/browse/FLPATH-4480\n\nForward signed-in users' OIDC access tokens to authenticated DCM control planes while preserving RHDH authentication. Support DCM auth-disabled deployments and guest-only development without an IdP.

### DIFF
--- a/workspaces/dcm/README.md
+++ b/workspaces/dcm/README.md
@@ -30,3 +30,22 @@ This runs the full app and backend concurrently (frontend at http://localhost:30
 - **yarn start:dev** – Run both plugins in standalone mode (no full app/backend).
 
 Configuration is in `app-config.yaml`. Example catalog data is in `examples/`.
+
+## Authentication
+
+Set `dcm.auth.enabled` to match the DCM control plane's `auth.enabled` setting.
+
+- When enabled, the DCM UI uses the signed-in user's OIDC access token. Configure
+  an OIDC provider in RHDH `auth.providers`, including its metadata URL, client
+  ID, and client secret. The proxy requires normal RHDH/Backstage credentials
+  and the user token, then forwards that token to DCM as an upstream Bearer
+  token.
+- When disabled, the UI does not require an OIDC provider or send an upstream
+  Bearer token. The proxy still requires normal RHDH/Backstage credentials, so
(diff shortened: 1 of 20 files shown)


asmasarw (changes requested):
Hi @chadcrum ,
Overall looks fine - but there are some items need to be addressed.

1. Missing changeset
The PR touches three packages (`dcm`, `dcm-backend`, `dcm-common`) and changes public API (`dcm-common/report.api.md `now exports `DcmOidcTokenProvider` and a new `DcmBaseClient` constructor option). 

Please Run `yarn changeset` from workspaces/dcm and add entries for the affected packages.

2. License header truncated in `plugins/dcm-backend/src/routes/proxy.test.ts`
The full Apache 2.0 header was reduced to a 4-line stub.
Please Restore the full standard header block.

3. Side effect during render + module-level mutable singleton
`DcmAuthConfigurator` calls `setDcmAccessTokenProvider(...)` directly in the render body (`Router.tsx`), mutating a module-global in `DcmAuth.ts`. 

Under React 18 StrictMode/concurrent rendering this runs on every render and re-entrantly mutates shared global state.

Wrap the call in `useEffect (deps [authEnabled, oidcAuthApi])`, or better, wire the token provider through a proper Backstage API factory in plugin.ts rather than a module-level singleton, so it aligns with how the other DCM clients are constructed.

chadcrum:
@asmasarw addressed the review feedback:

- Added `workspaces/dcm/.changeset/quiet-auths-forward.md` covering the frontend, backend, and common DCM packages.
- Restored the full Apache 2.0 header in `plugins/dcm-backend/src/routes/proxy.test.ts`.
- Replaced the render-time/module-global auth provider approach with Backstage API factories. The default factory reads `dcm.auth.enabled` without resolving `internal.auth.oidc`; enabled hosts can register the DCM OIDC adapter at app scope.

The changes are in `72750d6234`. Targeted tests, type checking, API reports, builds, and formatting checks pass. The worktree is clean.

@asmasarw, could you take another look when you get a chance?

asmasarw (approved):
LGTM

PR #4713, opened 2026-09-11
fix(theme): prevent entity page content clipping

rohitkrai03:
## Summary

- Fix the RHDH entity-page main-content layout so longer pages can grow beyond the viewport.
- Scope the flex layout override to BUI entity pages instead of relying on the previous broad main selector.
- Add regression coverage for the updated layout values.
- Include a patch changeset for the published theme package.

## Validation

- Focused theme package tests: 5 suites / 26 tests passed
- yarn prettier:check passed
- Changeset status passed

The workspace-wide theme TypeScript check still reports unrelated existing errors in plugins/bui-test/src/components/OtherExample.tsx.

### DIFF
--- a/workspaces/theme/.changeset/entity-page-layout.md
+++ b/workspaces/theme/.changeset/entity-page-layout.md
@@ -1,0 +1,6 @@
+---
+'@red-hat-developer-hub/backstage-plugin-theme': patch
+---
+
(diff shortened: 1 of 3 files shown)


christoph-jerolimov (approved):
Approve, let us test it together with more plugins.

PR #4716, opened 2026-09-11
fix(auth-app): handle no auth provider configured error

JessicaJHee:
### Changes

- Replaces the broken Guest/GitHub fallback with a descriptive ErrorPanel when auth.providers is empty or misconfigured.
- Removed `DEFAULT_PROVIDER = 'github'`, auth providers need to explicitly installed dynamically
- Removed default Guest login when `auth.environment` is development

Fixes [RHIDP-16623](https://redhat.atlassian.net/browse/RHIDP-16623)

Screenshot 2026-09-11 at 11 24 31 AM


workspaces/app-defaults/plugins/app-auth/src/components/SignInPage.tsx
```diff
@@ -233,8 +229,20 @@ export function SignInPage(props: SignInPageProps): React.JSX.Element {
     );
 
   if (providerConfigs.length === 0) {
```
ciiay, line 228: **Blocker — guest-only `auth.providers` breaks default app-defaults dev**

When `signInPage` is omitted, keys under `auth.providers` are used but `guest` is filtered out. With only `guest: {}` configured (see `workspaces/app-defaults/app-config.yaml`), `providerNames` is empty and this branch shows the new ErrorPanel.

That replaces today’s dev behavior (implicit GitHub fallback + auto-prepended Guest). Please confirm intent and either:
- restore a supported guest-dev path (e.g. inject Guest when `auth.providers.guest` exists and `auth.environment === 'development'`), or
- update the default app-config/README/e2e so local dev is not guest-only without extra config.

Also consider a distinct message when the only configured provider is `guest` — the current copy points at `auth.providers` even though guest is already set.

ciiay, line 228: **Blocker — `signInPage: guest` no longer resolves**

`guest` is not in `createProviders`, so `signInPage: 'guest'` yields an empty `providerConfigs` and hits this ErrorPanel. Previously, empty configs fell back to GitHub and dev prepended `'guest'` to the provider list.

If guest sign-in remains supported, resolve `'guest'` explicitly (Backstage accepts `providers={['guest', ...]}`) instead of only mapping through `createProviders`.

ciiay (changes requested):
Thanks for the PR 👏 

The no-fake-GitHub direction looks good. Only added some comments regarding guest / local-dev regressions below.

workspaces/app-defaults/plugins/app-auth/src/components/SignInPage.tsx
JessicaJHee: Added

JessicaJHee: Addressed

JessicaJHee:
@ciiay Thank you so much for the review and catching that! While addressing your feedback I realized that the guest provider config doesn't have frontend visibility. I made the contribution upstream (it's merged but will need the patch PR [here](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/3854) before we can pull that in). 

I tested with rhdh-local with [this image](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/3854#issuecomment-5720392461) and it's working as expected now 👍

ciiay (approved):
Thanks for the updates 💯 

/lgtm

PR #4721, opened 2026-09-14
refactor(app-defaults): limit main entry point exports to feature loader

christoph-jerolimov:
## Description

Narrow the public API of `@red-hat-developer-hub/backstage-plugin-app-defaults` by removing named re-exports from the package root entry point. Consumers should import the default feature loader from the package root, or use dedicated subpath exports (for example `./app-defaults-translations-module`) for other modules.

The API report is updated to reflect the reduced surface area.

## Checklist


christoph-jerolimov:
Close this in favor of #4724

PR #4724, opened 2026-09-14
refactor(app-defaults): expose plugin overrides as subpath exports

christoph-jerolimov:
## Summary
Follow-up to #4449. Instead of re-exporting the empty-state plugin overrides from the package entry point, each is now exposed as its own default-export npm subpath, matching the `./app-defaults-translations-module` convention:

- `./catalog-plugin-override`
- `./catalog-graph-plugin-override`
- `./scaffolder-plugin-override`
- `./api-docs-plugin-override`
- `./docs-plugin-override`

Each subpath default-exports its override via a small `*Export.ts` wrapper so it can be loaded individually (e.g. for dynamic / module-federation setups). The `createFrontendFeatureLoader` default export continues to bundle all overrides for the batteries-included path, so consumers using the feature loader are unaffected.

The overrides are no longer named exports of the package root.

## Changes
- Added five `*PluginOverrideExport.ts` wrapper files (default exports)
- Added the five subpath entries to `package.json` `exports` and `typesVersions`
- Removed the override named exports from `src/index.ts` (feature loader still imports them internally)
- Regenerated API reports (one per new subpath)
- Added a changeset (minor)

## Test plan
- [x] `yarn tsc` clean
- [x] ESLint clean
- [x] 14 unit tests pass
- [x] `yarn backstage-cli repo build --all` exits 0 (dts rollup succeeds)


davidfestal (approved):

christoph-jerolimov:
Discussed on slack and approved by David, merging it.

PR #4730, opened 2026-09-14
test(intelligent-assistant): cover the NFS wiring, not just that the plugin exists

gustavolira:
Closes the rhdh-plugins side of RHIDP-16455. `intelligent-assistant` was the last workspace in this repo with an overlay e2e suite and no `createExtensionTester` coverage.

## What was there before

`src/plugin.test.ts`, in full:

```ts
describe('lightspeed', () => {
  it('should export plugin', () => {
    expect(lightspeedPlugin).toBeDefined();
```

That passes against a plugin that contributes no UI at all. An NFS plugin which fails to contribute produces a clean boot with an empty page — no error, no console warning, exit 0. At the e2e layer that arrives as `heading "X" not found` after a ~20 minute deploy, in a different repository from the code that broke.

## What the new assertions pin

This plugin has more surface than most: two APIs, a page, an app drawer, and three `createFrontendModule` exports.

| Assertion | What breaks without it |
| --- | --- |
| page serves `LIGHTSPEED_PATH` and has a route ref | the sidebar entry and the legacy `/lightspeed` redirect land nowhere |
| both APIs, the page and the drawer are in `extensions[]` | `createExtensionTester` builds an extension in isolation, so dropping one leaves the path assertion green. The APIs carry no extension data, so this lookup is all that stands between them and a page whose every request fails |
| both route refs resolve | `lightspeedConversation` is a subRouteRef; losing it breaks deep links to a conversation without touching the page |
| the three modules declare `pluginId: 'app'` | they attach to `app`, not to this plugin. A module naming a host the app does not have is silently orphaned rather than an error |

## Dependency

`@backstage/frontend-test-utils` added as a devDependency — the plugin carried only `@backstage/test-utils`, which RHIDP-16455 flags as a recurring gap.

## Verification

7 tests, passing on a branch rebased onto current main.

Mutation verification is still running at the time of opening — each `yarn test` in this workspace takes a couple of minutes. I will report the results here and amend if any assertion turns out to be toothless. Opening now so review can start in parallel.


gashcrumb (approved):

karthikjeeyar (approved):
/lgtm

PR #4732, opened 2026-09-14
feat(scorecard): score first in-window DORA deployments for lead time and CFR

PatAKnight:

Median lead time and change failure rate only paired successful production deployments inside the 30-day window, so the first in-window deploy was never fully scored.
This change loads the latest persisted successful production deployment immediately before `windowFrom` and uses it as:

- **Lead time:** `baseCommitSha` for PRs into the first in-window deploy
- **CFR:** start of the interval ending at the first in-window deploy, so incidents in that gap are counted

Collector sync stays on the 30-day window. If no prior deploy is still in the DB, behavior is unchanged. Incident sync is not expanded; incidents before `windowFrom` count only if already retained.

Also included the `knip-reports` and removed some of the unused dependencies across the plugins. I can drop them in the event that we a cleaner, more focused PR.


workspaces/scorecard/plugins/scorecard-node/knip-report.md
```diff
@@ -0,0 +1,9 @@
+# Knip report
+
+## Unlisted dependencies (2)
+
+| Name              | Location                                             | Severity |
+| :---------------- | :--------------------------------------------------- | :------- |
+| @backstage/config | src/utils/thresholds/getThresholdsFromConfig.test.ts | error    |
```
dzemanov, line 7: Nit: Would you mind adding `@backstage/config` dev dependency to `scorecard-node`?

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/database/DatabaseDoraDeployments.ts
```diff
@@ -96,6 +107,32 @@ export class DatabaseDoraDeployments implements DoraDeploymentsStore {
     return rows.map(fromDoraDeploymentRow);
   }
 
+  async readCandidatesBefore(
+    catalogEntityRef: string,
+    collectorId: string,
+    collectorInputHash: string,
+    before: Date,
+    limit: number,
```
dzemanov, line 115: To simplify, let's load the single required deployment directly instead of returning candidates with limit of 50. We can filter in the database by `environment`.

Function `isProductionEnvironment` is going away anyways in this PR: https://github.com/redhat-developer/rhdh-plugins/pull/4767

dzemanov (changes requested):
Thank you @PatAKnight for this fix, I only have 2 small comments.

dzemanov (approved):
Tested that this works fine, thank you!

workspaces/scorecard/plugins/scorecard-backend-module-dora/src/service/DoraDataService.ts
```diff
@@ -62,6 +75,29 @@ export class DefaultDoraDataService implements DoraDataService {
...
 
+  async readLatestProductionDeploymentBefore(
+    catalogEntityRef: string,
+    options: CollectorCallOptions & {
+      before: Date;
+      productionEnvironments: string[];
+    },
+  ): Promise<DbDoraDeployment | undefined> {
+    const candidates = await this.deploymentsDb.readCandidatesBefore(
+      catalogEntityRef,
```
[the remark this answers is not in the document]
dzemanov: Outdated, we directly read latest before

PR #4755, opened 2026-09-14
feat(app-defaults): unify empty-state import button with route and permission checks

christoph-jerolimov:
Follow-up to #4449 / #4724.

## Summary
Consolidates the empty-state action button for the catalog, catalog graph, scaffolder, API docs, and TechDocs pages into a single, gated "import" button owned by `EmptyCatalogGate`.

- **Moved the button into `EmptyCatalogGate`.** Custom pages now pass an `importButtonTitle` string instead of building their own `ButtonLink` / passing an `action` element.
- **Route-aware.** The button is only shown when the `page:catalog-import` extension is installed, resolved via `useRouteRef(catalogImportPlugin.routes.importPage)`. The same route also provides the button's `href` (no more hardcoded `/catalog-import`).
- **Permission-aware.** The button is only shown when the user is allowed to create catalog entities (`catalog.entity.create`, via `usePermission`).
- The route and permission hooks live in an internal `CatalogEmptyState` child component, so they only run when the empty state is actually rendered (not during loading, on error, or when entities exist).

## Changes
- `src/catalog/EmptyCatalogGate.tsx` — extract `CatalogEmptyState`, gate the button on route availability + permission
- Custom pages (`catalog`, `catalogGraph`, `scaffolder`, `apiDocs`, `docs`) pass `importButtonTitle`
- New deps: `@backstage/plugin-catalog-import`, `@backstage/plugin-catalog-common`, `@backstage/plugin-permission-react`
- Tests for button shown / hidden (route unavailable, permission denied)
- Changeset (patch)

## Test plan
- [x] `yarn tsc` clean
- [x] ESLint clean
- [x] 16 unit tests pass
- [x] `yarn backstage-cli repo build --all` exits 0


karthikjeeyar (approved):
Verified, changes looks good to me!

/lgtm

PR #4757, opened 2026-09-15
feat(app-defaults): extensible sidebar with blueprints, a default layout, and route-guarded items

christoph-jerolimov:
## Description

Makes the RHDH sidebar extensible for the new frontend system. Plugins contribute entries through blueprints from `app-react`, and `app-defaults` renders them in priority order together with a default layout (logo, search, notifications, administration, settings), so an app that installs app-defaults gets a complete sidebar without hand-written nav code.

### Blueprints (`@red-hat-developer-hub/backstage-plugin-app-react`)

| Blueprint | Kind | Purpose |
| --- | --- | --- |
| `SidebarItemBlueprint` | `sidebar-item` | A sidebar entry with `title`, `icon` (system icon key or component), `to` or `onClick`, `priority`, an optional `group`, and an optional `requiresRoute` (only render when the app has a nav route matching `to`). |
| `SidebarItemGroupBlueprint` | `sidebar-item-group` | A group (`id`, `title`, `icon`, optional `to`, `priority`) collecting the items that reference its `id`. `variant: 'inline'` (default) renders them in a collapsible list below the entry, `variant: 'flyout'` in a hover submenu. |
| `SidebarElementBlueprint` | `sidebar-element` | A custom React component (`component`, `priority`, optional `to`) for entries that need hooks or context, such as the search modal or the notifications item. |
| `SidebarSpacerBlueprint` | `sidebar-spacer` | A fixed gap, or with `grow: true` a flexible space that pushes lower entries to the bottom. |
| `SidebarDividerBlueprint` | `sidebar-divider` | A horizontal divider. |

All of them attach to `nav-content:app/sidebar`. `title`, `icon`, `to`, `priority`, `group`, `requiresRoute`, `variant`, and `grow` can be overridden per extension from `app-config.yaml`, and any extension can be disabled there.

### Rendering (`@red-hat-developer-hub/backstage-plugin-app-defaults`)

`appSidebarExtension` (`nav-content:app/sidebar`, also exposed as `appSidebarModule` and the `./app-sidebar-module` subpath, and part of `appDefaultsModule`) collects the contributions and renders them with these rules:

- Top-level entries (items, groups, elements) sort by `priority`, higher first, with title or extension id as tiebreak. Items inside a group sort the same way.
- Positive-priority entries stay pinned above a scroll wrapper and negative-priority entries below it; the default (priority 0) menu items scroll independently inside the wrapper.
- Inline groups toggle on click and start expanded when the current route matches one of their items. Groups without visible items and without a `to` are not rendered.
- Nav items that Backstage auto-discovers from page extensions are merged in at priority 0. An item or group with the same `to` replaces the auto-discovered entry, so a plugin can retitle, regroup, or reprioritize its own page.
- An element with a `to` hides every item and auto-discovered page with the same path, so the search modal replaces the plain search page entry.
- Items flagged `requiresRoute` are dropped unless an auto-discovered nav item links to the same path, so an entry that points at an optional plugin's page disappears when that plugin is absent.
- Every custom element is wrapped in an `ErrorBoundary`, so a failing element (or a missing plugin) never takes down the rest of the sidebar. Search and notifications are additionally guarded by API (and, for notifications, route) availability.

### Default layout

Shipped with `appSidebarModule` / `appDefaultsModule`, top to bottom:

| Extension ID | Priority | What it is |
| --- | --- | --- |
| `sidebar-element:app/logo` | 2000 | `CompanyLogo`: full logo while the sidebar is open, icon logo while collapsed, from `app.branding.fullLogo` / `iconLogo` / `fullLogoWidth` (light or dark variant by the app bar background scheme), with the RHDH logos as fallback. Aligned with the global header's `useBrandingFullLogo`. |
| `sidebar-spacer:app/logo` | 1500 | Gap below the logo. |
| `sidebar-element:app/search` | 1000 | `SidebarSearchModal`, claims `/search`. Disabled by default; enable via app-config. |
| `sidebar-divider:app/search` | 900 | Divider below the search. |
| *(auto-discovered pages, plugin items and groups)* | 0 | |
| `sidebar-spacer:app/bottom` | -30 | Growing spacer that pushes the rest to the bottom. |
| `sidebar-divider:app/bottom` | -35 | |
| `sidebar-element:app/notifications` | -40 | `NotificationsSidebarItem`, claims `/notifications`. Disabled by default; enable via app-config. |
| `sidebar-divider:app/settings` | -90 | |
| `sidebar-item-group:app/admin` | -95 | Administration group (`id: admin`). Ships the route-guarded RBAC item and collects any further items with `group: 'admin'`; stays hidden while it has no visible items. |
| `sidebar-item:app/rbac` | 10 (in group) | RBAC item linking to `/rbac`, guarded with `requiresRoute` so it only shows when the RBAC plugin's page is installed. |
| `sidebar-item-group:app/settings` | -100 | Settings group (`id: settings`) linking to `/settings`; plugins can add items with `group: 'settings'`. |

Each default is an ordinary extension:

```yaml
app:
  extensions:
    - sidebar-element:app/search: true
    - sidebar-element:app/notifications: true
    - sidebar-item-group:app/settings:
        config:
          variant: flyout
    - sidebar-item:my-plugin/users:
        config:
          group: admin
          priority: 10
```

### Learning Paths sidebar item

The bundled Learning Paths module (`pluginId: app`) contributes a localized sidebar element that renders a Backstage `SidebarItem` with the title from the app-defaults translations (`menuItem.learningPaths`, translated for de/es/fr/it/ja) and links via the page route ref, replacing the plain auto-discovered nav item.

### Demo app

`packages/app` drops its hand-written nav module and registers a `sidebar-demo` module on top of the defaults: an inline Documentation group that takes over the TechDocs and API docs pages, a Help action item, an App Visualizer item in the Settings group, a Plugins item in the Administration group, a fixed top spacer above the logo, and a deliberately crashing element that demonstrates the per-element error boundary (disabled in app-config).

image

I've decided to keep the Company Logo like this for now, later this need to be disabled by default because its part of the header already...

## Checklist

https://claude.ai/code/session_016MQxi5MDUWcL4rcGwBefNE


workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
```diff
@@ -0,0 +1,240 @@
...
+      text={item.title}
+      onClick={() => item.onClick?.()}
+    />
+  );
+}
+
+function SidebarModelSubmenuItem({ item }: { item: SidebarModelItem }) {
+  const icon = useSidebarIcon(item.icon);
+  return <SidebarSubmenuItem title={item.title} to={item.to} icon={icon} />;
+}
```
divyanshiGupta, line 107: looks like this renderer silently drops item.onClick for grouped action items. Can we handle this —for example, render action items with an action-capable sidebar row or fall them back to the inline renderer?

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
```diff
@@ -0,0 +1,240 @@
...
+            );
+          case 'element': {
+            const Element = entry.element.component;
+            return <Element key={entry.element.id} />;
+          }
+          default:
+            return null;
+        }
+      })}
+    </Sidebar>
```
divyanshiGupta, line 266: The new sidebar renders every entry directly under Sidebar, without the previous SidebarScrollWrapper around the main navigation. With a sufficiently large plugin menu, (havent tried it but assuming looking at the code) the whole drawer scrolls, moving the company logo and the notifications/settings block out of view. Can you check once?

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
```diff
@@ -0,0 +1,151 @@
...
+  params: { priority: -35 },
+});
+
+/** Notifications item. Extension ID: `sidebar-element:app/notifications`. */
+export const sidebarNotificationsElement = SidebarElementBlueprint.make({
+  name: 'notifications',
+  params: {
+    component: NotificationsSidebarItem,
+    to: '/notifications',
+    priority: -40,
```
divyanshiGupta: These default extensions always mount NotificationsSidebarItem and SidebarSearchModal, even when the consuming app has not installed or registered the corresponding frontend features. NotificationsSidebarItem calls useApi(notificationsApiRef) during render, which can take down the sidebar; search can fail when opened. If this is correct, please add required checks, or explicitly document and enforce the required frontend modules.

divyanshiGupta (changes requested):

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
karthikjeeyar: I can reproduce this `No implementation available for apiRef{plugin.notifications.service}` when the rhdh doesn't have the notifications installed.

I think if we early return `null` when the `notificationsApiRef` is not available in the apiHolder then it should solve this issue and maybe we can also get rid of the configuration? or no?

```
- sidebar-element:app/notifications: false
```

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
christoph-jerolimov: Thanks, very good point, added `SidebarScrollWrapper`

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
christoph-jerolimov: Thats true, it would be great if we could add that sidebar element in the overlay repo of the notifications plugin. Maybe we can simplify that in the future. For now I think its okay if customers can opt-out from the extension, or? I will check if I can return null when the api isn't there.

christoph-jerolimov: Btw, we can disable both by default since search and notification are part of the header in RHDH.

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
christoph-jerolimov: SidebarSubmenuItem doesn't support onClick currently. I would keep it for the moment as it is.

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
christoph-jerolimov: Implemented, also if I'm not 100% happy.

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
christoph-jerolimov: Both are disabled by default and guarded now with checks if the searchApi / notificationApi+notifications route is available. I added also an ErrorBoundary around every custom nav item element (custom omponent).

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
```diff
@@ -0,0 +1,268 @@
...
+      case 'group':
+        return (
+          <SidebarModelGroupEntry key={entry.group.id} group={entry.group} />
+        );
+      case 'element': {
+        const Element = entry.element.component;
+        return (
+          <ErrorBoundary key={entry.element.id}>
+            <Element />
+          </ErrorBoundary>
```
karthikjeeyar, line 230: Default backstage `ErrorBoundary` renders an ErrorPanel, which is too disruptive for a single sidebar item. 

Image

---

How about creating a custom SidebarElementErrorBoundary that just renders `null` and logs the error in the browser console?

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
```diff
@@ -0,0 +1,193 @@
...
+ * `sidebar-item-group:app/admin`.
+ */
+export const sidebarAdminGroup = SidebarItemGroupBlueprint.make({
+  name: 'admin',
+  params: {
+    id: 'admin',
+    title: 'Administration',
+    icon: AdminIcon,
+    priority: -95,
+  },
```
karthikjeeyar, line 143:
```suggestion
    priority: -95,
  },
    if: {
    permisisons: {
      $contains: catalogEntityCreatePermission.name,
    }
  }
```

Can we add the `if` predicate here for administration tab guarded by `catalogEntityCreatePermission`?

workspaces/app-defaults/plugins/app-react/src/sidebar/extensions/SidebarItemBlueprint.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+export const SidebarItemBlueprint = createExtensionBlueprint({
+  kind: 'sidebar-item',
+  attachTo: { id: 'nav-content:app/sidebar', input: 'items' },
+  output: [sidebarItemDataRef],
+  dataRefs: {
+    item: sidebarItemDataRef,
+  },
+  configSchema: {
+    title: z.string().optional(),
```
karthikjeeyar, line 67:
```suggestion
    title: z.string().optional(),
    titleKey:  z.string().optional(),

```

Shall we add `titleKey` for translation support ? or Is there a separate ticket to handle translation support ?

workspaces/app-defaults/plugins/app-defaults/src/sidebar/AppSidebar.tsx
christoph-jerolimov: Like that idea, I thought its an edge case. For me the ErrorBoundary was at least collapsed by default.

workspaces/app-defaults/plugins/app-react/src/sidebar/extensions/SidebarItemBlueprint.ts
christoph-jerolimov: I wanted at it as a follow-up. Currently looking into translation for the entity tabs. At the moment I translate there label directly and I would do something similar here I guess. Will do this as a follow-up.

workspaces/app-defaults/plugins/app-defaults/src/sidebar/defaultSidebarExtensions.ts
christoph-jerolimov: I don't think the admin tab has a relation ship to create entity tab

karthikjeeyar (approved):
/approve

PR #4758, opened 2026-09-15
feat(app-defaults): legacy RHDH entity page tabs on NFS

rohitratannagar:
## Summary

Ports the legacy RHDH **EntityPage** experience into **`@red-hat-developer-hub/backstage-plugin-app-defaults`** for the **new frontend system (NFS / app-next)**. When the app-defaults feature loader is installed, catalog entities get RHDH-default tabs and Overview layout without adopter `app.extensions` YAML.

### What this implements

- **Overview (`/`)** — Overrides `entity-content:catalog/overview` with legacy `OverviewTabContent` in the RHDH 12-column **container-query** `Grid` (from [rhdh EntityPage](https://github.com/redhat-developer/rhdh/tree/main/packages/app/src/components/catalog/EntityPage)). Uses `params.loader` so the custom loader replaces the stock NFS card grid.
- **Api (`/api`)** — Overrides `entity-content:api-docs/apis` for service **components**.
- **Definition (`/definition`)** — Overrides `entity-content:api-docs/definition` for **API** entities.
- **Dependencies (`/dependencies`)** — `entity-content:catalog/rhdh-dependencies` for **components**.
- **Diagram (`/system`)** — `entity-content:catalog/rhdh-system-diagram` for **systems**.
- **Hygiene** — Disables NFS overview cards that belong on other tabs; disables catalog-graph **Relations** on overview where legacy did not (e.g. groups).
- **Deps** — `@backstage/catalog-model`, `@backstage/plugin-org`; `FrontendFeatureLoader` on default export for dynamic loading.

Empty-state gates (catalog, API docs, TechDocs, scaffolder, catalog-graph) are unchanged.

### What we gain

- Legacy **tab set and card placement** for the five built-in RHDH tabs without per-adopter config.
- NFS-native plugin overrides + dynamic app-defaults bundle for app-next / RHDH.

### What we lose / trade-offs

- **Per-card Overview toggles** via `entity-card:*: false` in `app.extensions` (Overview is a fixed loader + `OverviewTabContent`, not the stock card assembly). Whole-tab changes via `entity-content:*` still work.
- **Not full OFS EntityPage** — no `entity.page.*` mount-point tab matrix or legacy `entityTabs` ordering; plugin tabs use NFS `entity-content:*`.
- **No mount-point `config.layout` YAML** — layout is in code.
- **Follow-up** — `EntityContentLayoutBlueprint` + stock cards could restore configurability with legacy placement.

## Test plan

- [ ] `yarn build` in `workspaces/app-defaults/plugins/app-defaults`
- [ ] Group: Overview — About left; profile / ownership / members right; no Relations on overview
- [ ] Service component: Dependencies + Api tabs
- [ ] API: Definition tab; System: Diagram tab
- [ ] Rebuild dynamic plugin and verify on rhdh-local with app-defaults enabled

 #### JIRA: https://redhat.atlassian.net/browse/RHIDP-16564


workspaces/app-defaults/plugins/app-defaults/src/catalog/entityPageExtensions.ts
```diff
@@ -0,0 +1,80 @@
...
+ * limitations under the License.
+ */
+
+import { isKind } from '@backstage/plugin-catalog';
+import catalogPlugin from '@backstage/plugin-catalog/alpha';
+import { EntityContentBlueprint } from '@backstage/plugin-catalog-react/alpha';
+
+import { legacyTabLoader } from './entityPage/legacyTabLoader';
+
+const overviewOverride = catalogPlugin
```
rohitkrai03, line 23: Could we avoid replacing the entire Overview loader here? The legacy cards are useful, but this bypasses NFS's `inputs.cards` and `EntityContentLayoutBlueprint` path. That means cards contributed by other plugins or adopters can no longer be composed here, and the card and layout logic all moves into `OverviewTabContent`.

My suggestion would be to register the cards with `EntityCardBlueprint` and let a custom layout consume `EntityContentLayoutProps.cards`.

workspaces/app-defaults/plugins/app-defaults/src/catalog/entityPage/OverviewTabContent.tsx
```diff
@@ -0,0 +1,282 @@
...
+              <EntityLinksCard />
+            </Grid>
+          </EntitySwitch.Case>
+        </EntitySwitch>
+        <Grid item sx={{ gridColumn: '1 / -1' }}>
+          <EntityAboutCard />
+        </Grid>
+      </Grid>
+    </Grid>
+    <EntitySwitch>
```
rohitkrai03, line 95: One edge case to consider: this `EntitySwitch` has no `component` or `plugin` branch. With the stock relation, dependency, and API cards disabled, those entities can end up showing only About/Links, leaving the main area empty.

Could we add a generic fallback, or keep these cards in the NFS card composition so the layout can render whatever is available?

rohitkrai03 (changes requested):
Thanks for bringing the legacy Overview cards and tabs into the new frontend system. I agree that we need to solve two separate gaps here:

1. Restore the legacy cards that are not currently available through NFS.
2. Avoid the empty column created by the default Overview layout when one group of cards is missing.

The PR does address the first point by adding the old cards and tabs. My concern is that it addresses the second by replacing the whole Overview with a copied legacy implementation and disabling the corresponding NFS cards. That may reproduce the legacy screen, but it also makes this Overview responsible for card discovery and composition. Cards contributed by other plugins or adopters can no longer be composed through the normal NFS path.

I think `EntityContentLayoutBlueprint` gives us a more balanced approach. We could:

- keep the new API, Dependencies, Definition, and Diagram tabs as `EntityContentBlueprint`s;
- expose the missing legacy cards individually through `EntityCardBlueprint`; and
- add an `EntityContentLayoutBlueprint` targeting the existing catalog Overview.

The layout blueprint would receive `EntityContentLayoutProps.cards`, rather than owning the card list. It could group the cards by their `info`/`content` type and render those groups in the desired responsive grid: use two columns when both groups contain cards, and let the non-empty group span the full width when the other group is empty. Any ungrouped cards could be rendered in a fallback section.

This keeps NFS responsible for discovering and composing cards while the RHDH-specific layout only controls placement. It should give us the legacy experience, fix the blank-space problem, and still allow cards from RHDH extensions or adopters to appear without another hard-coded entity-page implementation. The Backstage demo uses a similar pattern: https://github.com/backstage/demo/blob/master/packages/app/src/components/catalog/EntityOverviewLayout.tsx

rohitratannagar:
closing this  in favour of https://github.com/redhat-developer/rhdh-plugins/pull/4832

PR #4767, opened 2026-09-15
fix(scorecard): filter deployments in by environment in db

djanickova:

Fixes [RHDHBUGS-3722](https://redhat.atlassian.net/browse/RHDHBUGS-3722)

DORA metric providers no longer load all stored deployments for an entity/window and filter production environments in memory. 

`readDeployments` now accepts optional `productionEnvironments` and applies that filter in SQL when reading `dora_deployments` (null/empty environment still counts as production; named environments match case-insensitively). Deployment Frequency, Median Lead Time for Changes, and Change Failure Rate pass `scorecard.plugins.dora.productionEnvironments` through. Omitting the option keeps an unfiltered read for non-metric callers/tests. Config behavior is unchanged.


dzemanov (approved):
Thank you @djanickova!
Tested with different environments, including empty and null, and works fine.

PR #4771, opened 2026-09-15
fix(theme): preserve legacy Page grid with BUI content

rohitkrai03:
## Summary

Follow-up to #4713. Its BUI descendant selector could also match a legacy Backstage Page and replace the Page's grid layout with flex.

- preserve `display: grid` for `main[data-backstage-core-page]`
- retain the flexible, long-content layout for classless NFS/BUI page wrappers
- add unit and browser regression coverage for both cases

## Testing

- `yarn tsc`
- `yarn workspace @red-hat-developer-hub/backstage-plugin-theme test --watchAll=false` (95 tests)
- focused NFS Playwright layout tests
- theme package build
- Prettier check


rohitkrai03:
Closing because https://github.com/redhat-developer/rhdh-plugins/pull/4891 made it obsolete.

PR #4797, opened 2026-09-16
feat(x2a): add default prompt template for new adversarial agents

yray-pixel:
Adds a default prompt template to the "Add Agent" form so users have a starting prompt when creating new  adversarial agents


workspaces/x2a/plugins/x2a/src/components/AdversarialAgentsPage/AgentDialog.tsx
```diff
@@ -44,6 +44,18 @@ interface AgentDialogProps {
   agent?: AdversarialAgent;
 }
 
+const DEFAULT_PROMPT = `You are a hostile reviewer. Your job is to find [describe the specific concern, e.g. unsafe destructive operations] in the migrated Ansible content.
```
eloycoto, line 47: I'm more on favor to have a curated list of Adversarial agents, the idea of this is what we recommend you to have, and validate before you're getting into x2ansible. And people can take frrom this: 

Image

PR #4804, opened 2026-09-16
chore(global-header): remove legacy OFS and Scalprum code

jonkoops:
Removes the global-header legacy dev shell, OFS plugin source, and Scalprum dependencies. E2E and workspace scripts use the NFS app only.

[RHIDP-15353](https://redhat.atlassian.net/browse/RHIDP-15353)


workspaces/global-header/plugins/global-header/app-config.dynamic.yaml
```diff
@@ -11,136 +11,3 @@ dynamicPlugins:
       #         config:
```
jonkoops, line 11: I left this NFS config commented out deliberately; please let me know if this needs to be enabled now that the legacy config has been removed.

jonkoops:
Not sure if a changeset is needed here; I can add one, which would likely be a major version bump (I presume).

christoph-jerolimov (approved):
Lgtm! A changeset would be nice. I approve it already so that you can merge it after that.

ciiay:
LGTM on the deletion + NFS-only e2e. Please add a **major** changeset for removing `/legacy`, and either update or clearly deprecate the remaining OFS mount-point docs (`configuration.md`, index “two modes”, component examples). NFS extensions in `app-config.dynamic.yaml` can stay commented.

jonkoops:
Added a changeset and updated the docs. I did not remove all the references to the OFS but marked them as removed in the documentation because I don't have time to update all the docs to be correctly ported to NFS, nor the knowledge of this specific plugin, so I will leave this to the plugin owners to take care of.

ciiay (approved):
With follow up PR #4906 opened, approving this PR 🤝 

/lgtm

workspaces/global-header/plugins/global-header/app-config.dynamic.yaml
ciiay: Keeping it commented looks right to me. NFS wiring goes through `app.extensions` (as the comment shows), not through `dynamicPlugins.frontend` mountPoints the way OFS did — so enabling this block as live config wouldn’t do what people expect. Fine as guidance only.

PR #4809, opened 2026-09-16
chore(homepage): remove legacy OFS and Scalprum code

jonkoops:
Removes legacy frontend support from the homepage workspace. Dev and e2e are NFS-only. Follows the same approach as #4804.

[RHIDP-15353](https://redhat.atlassian.net/browse/RHIDP-15353)


christoph-jerolimov (approved):
Lgtm. Thanks @jonkoops

PR #4832, opened 2026-09-16
Feat/rhidp 16564 entity overview nfs

rohitratannagar:
## Summary

Aligns **NFS catalog entity pages** in `@red-hat-developer-hub/backstage-plugin-app-defaults` with **legacy RHDH 1.10** when the app-defaults feature loader is used—**without** requiring `app.extensions` YAML in the host app.

**Jira:** [RHIDP-16564](https://issues.redhat.com/browse/RHIDP-16564)

> **“Stock”** = default **upstream Backstage** NFS entity pages (no app-defaults overrides). **RHDH defaults** = this PR’s behavior via `appDefaultsFeatureLoader`.

### Default UX by kind

| Kind | Overview | Other tabs |
|------|----------|------------|
| **Component** | About + Links **left**; no dependency/API list cards; no relations graph | **Dependencies**: graph **left**, list cards **right** |
| **API** | About left; relations graph **right**; no Definition card on Overview | Definition unchanged under Documentation → **Definition** |
| **System** | About left; relations graph **right** | **System Diagram** |
| **Group / User** | About left | No relations graph |

Stock Backstage tabs (Documentation group, Kubernetes, etc.) are unchanged.

### Zero-config enable

Load **`appDefaultsFeatureLoader`** from `@red-hat-developer-hub/backstage-plugin-app-defaults`.

Plugin overrides register:

- Overview layout: `entity-content-layout:catalog/rhdh`
- About/Links `type: info` in `catalog-plugin-override` (not app YAML)
- Tabs: `entity-content:catalog/rhdh-dependencies`, `entity-content:catalog/rhdh-system-diagram`
- Graphs: default `entity-card:catalog-graph/relations` disabled; `rhdh-overview-relations` (API/System) and `rhdh-dependencies-relations` (components)
- API docs: `entity-card:api-docs/definition` hidden on Overview

**Cards moved to Dependencies (components):** `depends-on-components`, `depends-on-resources`, `has-subcomponents`, `consumed-apis`, `provided-apis`.

More detail: plugin README — **Catalog entity pages (NFS)**.

### Extension IDs (quick reference)

| Purpose | Extension ID |
|--------|----------------|
| Overview layout | `entity-content-layout:catalog/rhdh` |
| Dependencies tab | `entity-content:catalog/rhdh-dependencies` |
| System Diagram tab | `entity-content:catalog/rhdh-system-diagram` |
| Overview graph (API, System) | `entity-card:catalog-graph/rhdh-overview-relations` |
| Dependencies graph (component) | `entity-card:catalog-graph/rhdh-dependencies-relations` |
| Default relations card (replaced) | `entity-card:catalog-graph/relations` |
| API definition on Overview (hidden) | `entity-card:api-docs/definition` |

---

### Default — full RHDH UX

Use `appDefaultsFeatureLoader` only; no entity-page YAML required.

### A — Disable Dependencies / System Diagram tabs

```yaml
app:
  extensions:
    - entity-content:catalog/rhdh-dependencies: false
    - entity-content:catalog/rhdh-system-diagram: false
```

### B — Default upstream overview columns (disable RHDH layout)

```yaml
app:
  extensions:
    - entity-content-layout:catalog/rhdh: false
```

### C — Default upstream single relations graph on Overview

```yaml
app:
  extensions:
    - entity-card:catalog-graph/rhdh-overview-relations: false
    - entity-card:catalog-graph/rhdh-dependencies-relations: false
    - entity-card:catalog-graph/relations: true
```

### E — Tune relations graph (height / direction)

```yaml
app:
  extensions:
    - entity-card:catalog-graph/rhdh-overview-relations:
        config:
          height: 500
    - entity-card:catalog-graph/rhdh-dependencies-relations:
        config:
          height: 480
          direction: TB
```

### Automated

- [x] `yarn tsc:full` in `workspaces/app-defaults`
- [x] `yarn workspace @red-hat-developer-hub/backstage-plugin-app-defaults test`
- [x] `yarn workspace @red-hat-developer-hub/backstage-plugin-app-defaults build`
- [x] `yarn build:api-reports:only` in `workspaces/app-defaults`
- [x] Prettier + pre-commit ESLint on changed plugin sources

### Manual (`workspaces/app-defaults`, `yarn start`)

| Check | Expected |
|-------|----------|
| Component → Overview | About/Links left; no list cards; no graph |
| Component → Dependencies | Graph left; list cards right |
| API → Overview | Graph right; no Definition card |
| API → Definition tab | Definition loads |
| System → Overview | Graph right |
| System → System Diagram | Diagram loads |
| Group/User → Overview | About left; no graph |


workspaces/app-defaults/plugins/app-defaults/src/catalog/buildEntityFilterFn.ts
```diff
@@ -0,0 +1,38 @@
...
+): (entity: Entity) => boolean {
+  if (filterFunction && filterExpression) {
+    return filterFunction;
+  }
+  if (filterFunction) {
+    return filterFunction;
+  }
+  if (filterExpression) {
+    return () => false;
+  }
```
karthikjeeyar: The NFS entity-card configuration supports both predicate functions and string filter expressions but `filterExpression` is never parsed and evaluated here. It always returns `false`, therefore card with filterExpression is always filtered out.

example filterExpression: 
```
  filter: "kind = component"
```

workspaces/app-defaults/plugins/app-defaults/src/catalog/entityDependenciesContentExtension.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { EntityDependenciesLayout } from './entityDependenciesLayout';
+
+/**
+ * Dependencies tab composed from NFS {@link EntityCardBlueprint} extensions.
+ *
+ * @internal
+ */
+export const entityDependenciesContentExtension =
+  EntityContentBlueprint.makeWithOverrides({
+    name: 'rhdh-dependencies',
```
karthikjeeyar: `rhdh-dependencies` does not communicate that extension is for a entity dependencies, and could lead to confusions. Consider renaming it to be more specific, eg: `rhdh-component-dependencies`  or something similar.

workspaces/app-defaults/plugins/app-defaults/src/catalog/buildEntityFilterFn.ts
karthikjeeyar: Add tests for function filters, string filters, and the function plus expression case.

workspaces/app-defaults/plugins/app-defaults/src/catalog/entityDependenciesLayout.tsx
```diff
@@ -0,0 +1,88 @@
...
+          key={card.element.key ?? index}
+          sx={{
+            ...gridItemSx,
+            gridColumn: graphColumn,
+            gridRow: { md: '1 / span 6', lg: '1 / span 6' },
+          }}
+        >
+          {card.element}
+        </Grid>
+      ))}
```
karthikjeeyar: Multiple info cards are currently assigned to the exact same grid position (gridColumn and gridRow set to 1 / span 6). Because of this, any extra info card added to the Dependencies tab, including consumer overrides ends up hiding behind the graph card and looks like it's missing entirely.

I have added `Links` card to this dependencies and it is completely hidden behind the relations graph.

Image

---

To reproduce this, you can add a component with links metadata and simply attach it to the `entity-content:catalog/rhdh-dependencies` extension as a info card.

app-config.yaml
```
app:
 extensions:
    - entity-card:catalog/links:
        attachTo:
          id: entity-content:catalog/rhdh-dependencies
          input: cards
        config:
          type: info
```

karthikjeeyar (commented):
Thanks @rohitratannagar!  I have verified the plugin locally. The changes in this PR looks good to me.

LGTM!

Tagging @rohitkrai03 for a final look as he had some comments in the older PR.

workspaces/app-defaults/plugins/app-defaults/src/catalog/entityOverviewLayout.tsx
```diff
@@ -0,0 +1,128 @@
...
+  );
+  const ungroupedCards = cards.filter(
+    card => card.type && card.type !== 'info' && card.type !== 'content',
+  );
+
+  const hasInfo = infoCards.length > 0;
+  const hasContent = contentCards.length > 0;
+  // Legacy RHDH keeps About/Links in the left band even when Overview has no
+  // right-column cards (e.g. components after dependency cards move off Overview).
+  const infoColumn = leftColumn;
```
rohitkrai03, line 66: One thing I think we still need to handle here is the case where only one group of cards is present. Right now `infoColumn` and `contentColumn` always use fixed spans, so if a component only has About/Links after its dependency cards are moved to the Dependencies tab, those cards still occupy only the left 4/12 and the rest of the page is empty. The same happens in reverse for content-only entities.

Could we make the non-empty group span the full width and only use the two-column spans when both groups have cards? That would address the blank-space issue this layout is intended to solve.

workspaces/app-defaults/plugins/app-defaults/src/catalog-graph/catalogGraphPluginOverride.tsx
```diff
@@ -16,16 +16,33 @@
...
 export const catalogGraphPluginOverride = catalogGraphPlugin.withOverrides({
   extensions: [
+    catalogGraphPlugin
+      .getExtension('entity-card:catalog-graph/relations')
+      .override({
+        factory(originalFactory) {
+          // Replaced by rhdh-overview-relations / rhdh-component-dependencies-relations.
+          return originalFactory({
+            params: {
+              filter: () => false,
```
rohitkrai03, line 48: I noticed the README says that setting `entity-card:catalog-graph/relations: true` restores the stock relations card. With this override, though, the extension's factory always supplies `filter: () => false`, so turning it back on would still render nothing.

Could we either leave a way for the original filter/factory to be restored, or adjust the documentation to say that the catalog-graph override itself has to be removed or disabled to get the stock card back?

rohitratannagar: Addressed — restoring the stock relations card now uses `config.useOriginalFactory` (plain `: true` is not enough because the override factory still applied `filter: () => false`).

```yaml
app:
  extensions:
    - entity-card:catalog-graph/relations:
        config:
          useOriginalFactory: true
    - entity-card:catalog-graph/rhdh-overview-relations: false
    - entity-card:catalog-graph/rhdh-component-dependencies-relations: false
```

Docs updated in the README under **Reverting to stock Backstage NFS entity UX**.

workspaces/app-defaults/plugins/app-defaults/src/api-docs/apiDocsPluginOverride.tsx
```diff
@@ -17,15 +17,29 @@
...
+ *
+ * Stock NFS also registers `entity-card:api-docs/definition` on Overview; RHDH
+ * shows API definition only on the Definition tab (`entity-content:api-docs/definition`).
  *
  * @public
  */
 export const apiDocsPluginOverride = apiDocsPlugin.withOverrides({
   extensions: [
+    ...entityDependenciesApiDocsCardAttachments,
+    apiDocsPlugin.getExtension('entity-card:api-docs/definition').override({
```
rohitkrai03, line 36: One more consistency issue with the relations override: `entity-card:api-docs/definition` is also hidden by replacing the original filter with `() => false`. If an adopter enables that extension again, this override still wins, so the card will not come back.

Could we expose the same `useOriginalFactory` option here, or document the supported configuration for restoring Definition on Overview? That would keep the “adopters can override individual extensions” behavior consistent.

workspaces/app-defaults/plugins/app-defaults/src/catalog/catalogPluginOverride.tsx
```diff
@@ -17,15 +17,24 @@
...
 
 /**
  * Override of the Backstage catalog plugin that adds an empty state
- * when no catalog entities are available.
+ * when no catalog entities are available, RHDH overview layout, NFS Dependencies tab,
+ * and legacy System Diagram entity tab.
  *
  * @public
  */
 export const catalogPluginOverride = catalogPlugin.withOverrides({
```
rohitkrai03, line 32: The new behavior now depends on several card attachments, layouts, and plugin overrides, but the tests currently only cover the filter parser. Could we add a small integration-level test for the default card placement and the `useOriginalFactory` path? This would help catch wiring regressions without testing the CSS layout itself.

workspaces/app-defaults/plugins/app-defaults/src/api-docs/apiDocsPluginOverride.tsx
rohitratannagar: Addressed in cde82a825 — entity-card:api-docs/definition now supports config.useOriginalFactory (same pattern as the relations override), and the README documents the restore YAML.

workspaces/app-defaults/plugins/app-defaults/src/catalog/catalogPluginOverride.tsx
rohitratannagar: Addressed in cde82a825 — added entityPageDefaults.integration.test.ts covering default Overview/Dependencies card placement and the useOriginalFactory paths for relations + API definition, plus entityOverviewLayout.test.tsx for the info/content column matrix.

rohitkrai03 (approved):
/approve

PR #4839, opened 2026-09-17
perf(intelligent-assistant): make GET /v1/models non-blocking

JslYoon:
## Summary

`GET /v1/models` probed each LLM's vision capability inline, awaiting up to `VISION_PROBE_TIMEOUT_MS` (10s) per model. On a cold cache this delayed the response by up to ~10s.

This change serves vision support from the capabilities cache and moves probing off the request path:

- **Warm at startup** — on router load, the model list is fetched and every un-cached LLM is probed in the background, so the first real request is already enriched.
- **Non-blocking on miss** — a request never awaits a probe. A cache hit returns the real value; a miss returns the conservative default (`supportsVision: false`) immediately and triggers a single background probe (deduped per model via an in-flight set).
- **Self-healing** — because only cache misses are probed, the cache re-probes models that are new to the LCS list or whose previous probe failed and was not cached, while leaving known models untouched.

The per-user `POST /v1/validate-model-vision` path is unchanged (it still awaits, since it needs the answer).

## Behavior change

On a cold cache, an LLM may report `supportsVision: false` for the brief window between the first request and the background probe completing; the next request reports the real value.

## Testing

- Updated the `supportsVision enrichment` suite for the non-blocking + warm-on-boot contract; added coverage for startup warm-up and lazy self-heal when the LCS list changes.
- `yarn backstage-cli package test` — 335/335 pass.
- `yarn tsc:full` — clean.
- Verified live against a running backend + Lightspeed service: `/v1/models` returns in ~50ms (was up to 10s), fully enriched.


maysunfaisal (approved):
generally lgtm

yangcao77:
I do not want to take the risk to merge PR might have regression concerns before FF. since it's a bugfix,and should not block FF, let's put this PR on hold for a while. 

some review comments from my Claude pr review skill, the critical ones do not look like easy to test or fix. PTAL
```
 Critical

  1. TTL expiry now returns a wrong answer instead of a slow correct one — router.ts:752-769 + attachment-validation.ts:29-33

  ModelCapabilitiesCache.get() deletes the entry and returns undefined once expired — there's no stale-serving path. So the
  first /v1/models after a positive entry's 24h TTL lapses reports supportsVision: false for a genuinely vision-capable model.
  On main this was invisible: the handler awaited the probe and returned true.

  The comment at router.ts:762 — "the next /v1/models call returns the real value" — is where this becomes user-visible,
  because the frontend doesn't make a next call:

  // plugins/intelligent-assistant/src/hooks/useAllModels.ts:36-37
  staleTime: 1000 * 60 * 5,
  refetchOnWindowFocus: false,

  Five minutes minimum, in practice until remount. Downstream, screen-context-utils.ts:453 silently drops the screenshot and
  :473 returns the textOnlyNoVision tooltip — the UI asserts to the user that their model lacks vision when the truth is "we
  haven't checked yet." Worth noting the frontend type already specifies the correct contract and the backend violates it:

  ▎ types.ts:77 — "set by the backend GET /v1/models handler; absent if capability could not be determined"

  supportsVision?: boolean is optional precisely so "undetermined" is representable.

  Fix: serve stale while revalidating (a read that ignores expiry), or omit the field on a miss and render a neutral
  "checking…" state.

  2. Warm-up is one-shot, untimed, and can fail completely silently — router.ts:259, :283-289

  Two compounding issues:

  - const upstream = await fetch(\${lcsBaseUrl}/v1/models`)has noAbortSignal, unlike probeModelVisionSupportat:180. An LCS 
  that accepts the TCP connection but never responds leaves this pending forever — the .catch()at:283` never fires, so not
  even the warning is logged. One-line fix, closes the most invisible hole in the diff.
  - It runs exactly once from createRouter() during registerInit, and never retries. LCS is a separate process at
  127.0.0.1:8080 that commonly isn't up yet at backend boot. The !upstream.ok branch at :260 has the same problem — a 503 from
  an initialising LCS permanently skips warm-up.

  So the changeset's "the cache is warmed at startup" only holds when LCS wins the startup race. Since warm-up only probes
  cache misses it's already idempotent — running it on a 10-15min interval fixes the boot race and resolves finding 1 by
  refreshing entries before they expire. Register the timer with coreServices.lifecycle.

  Important
  
  3. Both new debug messages state the opposite of what happened — router.ts:754, :765

  'vision probe success' fires on any cache hit, including cached === false — a model that probed as not vision-capable is
  logged as a success. 'vision probe failed' fires on a cache miss, where nothing failed and no probe ran. Real failures are
  logged under a different string entirely (Background vision probe failed for…, :242). Someone grepping the logs will chase
  an LCS outage that never happened. Suggest 'vision capability cache hit' / '…cache miss, probing in background', and { 
  identifier: cacheKey, supportsVision } instead of spreading ...model per model per request.

  4. A comment documenting a real footgun was dropped — router.ts:751

  main had, above const cacheKey = model.identifier:

  // `identifier` is already the `provider/model` key LCS and
  // /v1/validate-model-vision use — do not re-prefix with provider_id.

  Now gone. This matters because validation.ts:177 builds `${provider}/${model}` while the handler and the new warm-up key on
  model.identifier. If those ever diverge, chat 400s on every image upload while /v1/models reports supportsVision: true, and
  nothing logs the mismatch. Restore it.

  5. No shutdown cancellation — router.ts:236-289, plugin.ts

  Nothing registers with coreServices.lifecycle. Background probes (10s each) and the untimed warm-up outlive shutdown and
  then call logger.warn on a torn-down service. A shared AbortController in a shutdown hook covers this and the timer from
  finding 2.

  6. Every test now does unrelated network I/O into a module-global cache — router.test.ts

  startBackendServer() is called ~91 times and each now fires the warm-up. There's no default POST /v1/responses msw handler,
  so probes hit the catch-all. Running the suite confirms it: 94 passed, plus 16 stray Caught request to unknown path: 
  .../v1/responses lines. The count is 16 rather than ~180 because of a leak — the catch-all 404s, so the very first test (GET
  /health) caches false for both default models for the full 20-min negative TTL, and later warm-ups skip them. Only the
  vision describe's beforeEach masks this, and only inside that block. Passes today; any future test outside that block
  inherits false from whichever test ran first. Add a default POST /v1/responses handler to lcsHandlers.ts plus a top-level
  afterEach(() => ModelCapabilitiesCache.clear()).

  Medium
  
  - Non-ok probe caches false and logs nothing (:208 via :240). attachment-validation.ts:17-19 explicitly says LCS returns the
  same 5xx for a non-vision model and a transient error — so this is exactly the case that needs a log line, and there isn't
  one. An LCS schema change or a malformed TEST_VISION_JPEG would mark every model non-vision with zero log output.
  - Unbounded fan-out at startup (:270-277). One concurrent inference call per LLM, fired when LCS is coldest and most likely
  to shed load with 5xx — which then gets cached as false for 20 min. The warm-up can cause the false negatives it exists to
  prevent. Cap at 2-4.
  - No backoff (:237-251). The in-flight Set dedupes concurrent probes only; .finally() clears on settle. During an LCS outage
  every /v1/models re-arms a fresh 10s probe per model, each ending in a warn.
  - Changeset overstates self-healing. "one whose previous probe failed" is only true for network/timeout failures (nothing
  cached). A non-ok probe caches false for the full 20-min TTL and won't re-probe before then.
  - .catch().finally() discards the terminal promise (:240-250). If logger.warn throws, it's an unhandled rejection and the
  .finally() is skipped, wedging that key in the in-flight set for the process lifetime.

  Test nits

  - router.test.ts:402 — waitFor(() => probeCount === 1) on strict equality means a regression to 2 probes surfaces as a
  generic timeout, not a useful assertion failure. Use >= 1 and keep the existing expect(probeCount).toBe(1).
  - router.test.ts:412 — expect(llm.identifier).toBe('openai/gpt-4-turbo') is a leftover unrelated to the test's purpose.

  Suggested merge order

  1. AbortSignal.timeout on the warm-up fetch (one line, closes a silent hang)
  2. Fix the two inverted log messages + restore the dropped identifier comment (one-line each, and they clear the open
  fullsend review)
  3. Periodic re-warm + stale-while-revalidate — together these are what make the PR's headline claim actually hold
  4. Test-isolation cleanup, then the medium items


```

PR #4860, opened 2026-09-17
fix(intelligent-assistant): sources chip on stream end and unknown route handling

rohitratannagar:
### Notebook sources chip

Notebook chat did not show the RAG **sources chip** until the page was refreshed, even though the assistant response had finished and the server already had `referenced_documents`.

This change fixes three gaps in the live notebook path:

1. **SSE parsing** — Flush the decoder when the stream closes and parse a trailing `end` frame so `referenced_documents` are applied even when the final chunk has no `\n\n` delimiter.
2. **Persisted merge** — After the stream reaches `complete`, merge sources from refetched conversation messages when the in-memory transcript is still missing them.
3. **UI refresh** — Include the source count in the PatternFly `Message` React key so `SourcesChipModal` mounts when sources arrive after tokens stop.

**Root cause:** The sources chip depends on `referenced_documents` from the SSE `end` event and on `endContent` updating on the last bot message. Trailing SSE data could be dropped on `reader.done`, and the message component did not remount when only `sources` changed (content unchanged), so the chip appeared only after a full reload from the API.

### Unknown Intelligent Assistant URLs

Visiting an unrecognized path under `/intelligent-assistant` (for example `/intelligent-assistant/sdc`) previously rendered a blank screen.

The plugin router now catches unknown subpaths and:

- **Redirects** to `/intelligent-assistant` when `intelligent-assistant.chat` is allowed (uses `LIGHTSPEED_PATH`, not app root `/`).
- **Shows a 404** when chat access is denied.

## Test Plan

- [ ] Open a notebook with uploaded documents and send a prompt that returns RAG sources.
- [ ] Confirm the sources chip appears when the response completes, without refreshing the page.
- [ ] Open the chip popover and verify document titles/links.
- [ ] Send a follow-up message and confirm chips on earlier and latest bot replies.
- [ ] Switch Lightspeed display mode (overlay/docked/fullscreen) and confirm sources chips remain visible.
- [ ] Navigate to `/intelligent-assistant/<unknown>` with chat allowed — should land on `/intelligent-assistant` with the chat UI.
- [ ] Navigate to `/intelligent-assistant/<unknown>` without chat permission — should show Backstage 404.

## Checklist


its-mitesh-kumar:
The loader is going at wrong place just before the chat modal appear. Maybe we can create a minor bug to fix it.

Cc: @HusneShabbir 

https://github.com/user-attachments/assets/0c468203-df86-44b6-85ca-07ea9e3f4bc8

Also have we confirmed this loader with @ShiranHi / @tobiastal.

its-mitesh-kumar:
This is not introduced by this PR but we need a bug to track it. Delete Modal is getting hidden inside chat modal.
S_ 2026-09-21 at 1 23 24 PM
S_ 2026-09-21 at 1 28 57 PM

Lets have a bug for it.

Cc: @HusneShabbir

PR #4865, opened 2026-09-18
fix(x2a): trust extra CAs for converter Job git clone 

mareklibra:
fix(x2a): trust extra CAs for converter Job git clone

Fixes: FLPATH-4871

GitLab clone failed on a private CA because git runs in the Job, not the Hub.
Mount extra issuing CAs and keep skip-TLS as a last-resort hatch.


eloycoto:
shouldn't we add the CA into the secrets in the deployment directly?

https://github.com/x2ansible/x2ansible.github.io/blob/8be61f172ff0f8b75eac180a39310802c760652a/deploy/app.yaml#L304-L328

Having a way to add a custom CA, sems a bit insecure to me, and I do not think that scales well enough (What happens if multiple CA, etc..)

mareklibra:
Good question, I believe this has already been addressed:
- the caBundle is a public CA cert (not a secret)
- a PEM bundle already holds multiple CAs in one block and `useClusterTrustedCABundle` is precisely the "reuse deployment/cluster trust" path so operators don't paste PEM at all
- the `skipSSLVerification` is the only risky knob and it's auto-ignored whenever a CA is set.

Could you confirm this addresses the concern?

eloycoto:
Can we take a step back here and ask why we need this? and how to address this?

Could we think that maybe it's better to implement a think like `deployment[0]` in the config, and allow the agent to run with the patch? That allow the system to inject:

- Any configMap
- Any SSL cert
- Any custom secret that the convertor needs(Connect to local chef supermarket, puppet's forge)

So, instead of building a requirement, for something that was not requested by the user, and can be injected by the openshift[1], do something that can be compose on top of that? Something that it's more native in terms of Kubernetes and allow users to define what is needed for them.

[0]
```
$ --> kubectl explain backstage.spec.deployment
GROUP:      rhdh.redhat.com
KIND:       Backstage
VERSION:    v1alpha5

FIELD: deployment <Object>


DESCRIPTION:
    Valid fragment of Deployment to be merged with default/raw configuration.
    Set the Deployment's metadata and|or spec fields you want to override or
    add.
    Optional.

FIELDS:
  kind	<string>
  enum: Deployment, StatefulSet
    Kind of the deployment object. Can be Deployment or StatefulSet.

  patch	<Object>
    Valid fragment of Deployment to be merged with default/raw configuration.
    Set the Deployment's metadata and|or spec fields you want to override or
    add.
    Optional.


$ --> kubectl explain backstage.spec.deployment.patch
GROUP:      rhdh.redhat.com
KIND:       Backstage
VERSION:    v1alpha5

FIELD: patch <Object>


DESCRIPTION:
    Valid fragment of Deployment to be merged with default/raw configuration.
    Set the Deployment's metadata and|or spec fields you want to override or
    add.
    Optional.
```


[1] https://docs.redhat.com/en/documentation/openshift_container_platform/4.9/html/security_and_compliance/configuring-certificates

PR #4867, opened 2026-09-18
feat(app-defaults): migrate AutoLogout config to NFS

gashcrumb:
## Description

Implements [RHIDP-16432](https://redhat.atlassian.net/browse/RHIDP-16432): brings the AutoLogout feature from the legacy OFS app into the New Frontend System.

The `AutoLogout` component from `@backstage/core-components` is mounted as an `AppRootElementBlueprint` extension inside `appDefaultsModule`, which lives in the `app-defaults` dynamic plugin. The feature is **disabled by default** (`enabled: false`) and reads its full configuration from `auth.autologout.*` in `app-config.yaml` — matching the behaviour of the original OFS implementation ([RHIDP-9394](https://redhat.atlassian.net/browse/RHIDP-9394), PR [#3643](https://github.com/redhat-developer/rhdh/pull/3643)). Operators opt in by setting `auth.autologout.enabled: true`.

This PR also adds a `./app-defaults-module` sub-path export (`src/appDefaultsModuleExport.ts`) so that `appDefaultsModule` is correctly detected and loaded as a dynamic frontend feature by the plugin installer — previously it had no sub-path export and was not being registered at all.

## Fixed

- [RHIDP-16432](https://redhat.atlassian.net/browse/RHIDP-16432) — Migrate AutoLogout config to NFS

## Testing

Verified locally using `rhdh-cli plugin dev` against an `rhdh-local` instance configured with OIDC authentication (Keycloak). With `auth.autologout.enabled: true`, `idleTimeoutMinutes: 0.5`, and `promptBeforeIdleSeconds: 10` set in `app-config.local.yaml`, the "Are you still there?" dialog appeared after ~20 seconds of inactivity and the session was terminated after the prompt timeout.

The `app-visualizer` plugin was used to confirm that `app-root-element:app/auto-logout` is present and active in the extension tree after the plugin is loaded.

Screenshot From 2026-09-18 08-58-22

## Checklist


JessicaJHee (approved):
/lgtm

Thank you for this! Tested locally with rhdh-local and the GitHub auth provider and it's working as expected. I'll follow up with adding back the auto logout E2E tests in a separate PR in the overlays repo (based on these[ core tests](https://github.com/redhat-developer/rhdh/blob/3477b2eb06275498c351cb3c0d364dddb8c0bfdf/e2e-tests/playwright/e2e/auth-providers/oidc.spec.ts#L322))

PR #4877, opened 2026-09-19
fix(scorecard): Add a view data source menu on homepage sparkline card

Eswaraiahsapram:


dzemanov (approved):
I would like to point out that not all metrics with `visualization: sparkline` have collectors in `View datasources` - only metrics that are composite (like DORA). Non-composite metrics should show metric description. You can see it in the design:

https://github.com/user-attachments/assets/455fef27-28fe-4f1c-aefa-f884947f8289

Currently, there is no way to set visualization to a different type then the default type, and we have only DORA metric (that is composite). But, in future with CRUD, users can configure visualization type (so for example `github.openPRs` can be shown as 30 days trend). Furthermore, even now, users can contribute their own metrics and they can set `defaultVisualization` to sparkline, which would mean, `View datasources` is showing nothing for them (same for the entity page).

You can test it out by setting `defaultVisualization` for GithubOpenPrsProvider to `sparkline` and updating openPrsKpi from `statusGrouped` to `type: average`.

Design also shows that `View datasources` should be shown in all scorecards (homepage and entity page), not only sparkline scorecards. It can be fixed as a bug.
We have discussed this in [slack](https://redhat-internal.slack.com/archives/C09EF8982EQ/p1787140431624839?thread_ts=1787136628.063079&cid=C09EF8982EQ).
Adding metric description above collector description for View datasources entity and homepage is also to be worked upon as a separate bug. Translation of collectors description as well. I will create those tickets now to not forget.

I have tested that for composite metrics, VIew datasources works correctly:
(at the end, there is shown the bug for non-composite metric with sparkline.

https://github.com/user-attachments/assets/b33721cc-1328-41ea-8b56-935968babf6e

Translation + no View datasources dialog in drill down:

https://github.com/user-attachments/assets/ad78c7d3-b563-4a74-a6bb-63435b115e3e

dzemanov:
Created bugs:
- https://redhat.atlassian.net/browse/RHDHBUGS-3796
- https://redhat.atlassian.net/browse/RHDHBUGS-3797

